PR #51, opened 2023-02-12
Updated navbar design

MattIPv4:
Implements an updated navbar design mimicking the main Alveus site. Allows for each page to provide its own background for the navbar, which'll allow pages to do the header image slideshows etc. as is done on the main site.


pjeweb:
LGTM

Do we want to have a small indicator that its the community page in the header until we migrate the rest of the org website? Maybe putting the .gg in the title? Or a "Community Page" badge on the bottom right of "Alveus"?

PR #55, opened 2023-02-12
Implement homepage

MattIPv4:
Adds a homepage roughly matching what the main Alveus site has, with a Twitch player being added to the hero that links over to /live.


apps/website/src/pages/index.tsx
```diff
@@ -0,0 +1,435 @@
...
+      console.log("The video is playing", player);
+      setTwitchEmbedPlaying(true);
+    });
+  }, [twitchEmbedId, twitchEmbedLoaded])
+
+  return (
+    <>
+      <Head>
+        <title>Alveus.gg</title>
+        <meta name="robots" content="noindex" />
```
pjeweb, line 200: Should we set`noindex` on all pages (or in a robots.txt) and can we remove it from the pages then?

pjeweb (approved):
Great work!

apps/website/src/pages/index.tsx
MattIPv4: I've been setting it on the pages that I'm porting for now to avoid risking any SEO collision. I think it's fine to allow the "normal" pages on here (such as /live) to be indexed if found, as they don't collide with the main site.

pjeweb: Yeah makes sense. I was unsure about the giveaways etc. But they will be disabled soon anyways, depending on how long donators take to enter their addresses. I think Google wont find the site by itself until then and I would hold off registering it in the web search console.

PR #61, opened 2023-02-13
Footer: Include chanelrooh credits, link github

pjeweb:


apps/website/src/components/layout/footer/Footer.tsx
```diff
@@ -12,22 +16,41 @@ export const Footer: React.FC = () => {
           <ul className="flex flex-wrap items-center justify-between">
             <li className="basis-full p-2 md:basis-1/3">
               <p>
-                Built by the Alveus.gg team and community,
-                <br />
+                Original design by{" "}
```
MattIPv4, line 19:
```suggestion
                Original design by
                {" "}
```
Nit, so welcome to disagree, but I prefer putting these on a newline when I need them

apps/website/src/components/layout/footer/Footer.tsx
```diff
@@ -12,22 +16,41 @@ export const Footer: React.FC = () => {
...
+                <Link
+                  href="https://chanelrooh.com"
+                  target="_blank"
+                  rel="noreferrer"
+                  className={footerLinkClasses}
+                >
+                  chanelrooh.com
```
MattIPv4, line 26:
```suggestion
                  Chanelrooh
```
I think it'd be better to use their name?

apps/website/src/components/layout/footer/Footer.tsx
```diff
@@ -12,22 +16,41 @@ export const Footer: React.FC = () => {
...
+                </Link>
+                ,<br />
```
MattIPv4, line 28:
```suggestion
                {", "}
                <br />
```
Similar nitpick -- also adding a space, so the text copies correctly

apps/website/src/components/layout/footer/Footer.tsx
```diff
@@ -12,22 +16,41 @@ export const Footer: React.FC = () => {
...
+                built by the{" "}
```
MattIPv4, line 29:
```suggestion
                built by the
                {" "}
```
Similar nitpick

apps/website/src/components/layout/footer/Footer.tsx
```diff
@@ -12,22 +16,41 @@ export const Footer: React.FC = () => {
...
+                  target="_blank"
+                  className={footerLinkClasses}
+                >
+                  <IconGithub
+                    size={16}
+                    className="mr-1 inline-block fill-current"
+                  />
+                  Alveus.gg team
+                </Link>{" "}
+                and community, <br />
```
MattIPv4, line 42:
```suggestion
                and community
                {", "}
                <br />
```

apps/website/src/icons/IconGithub.tsx
```diff
@@ -0,0 +1,24 @@
...
+  <svg
+    xmlns="http://www.w3.org/2000/svg"
+    width={size}
+    height={size}
+    viewBox="0 0 98 96"
+    className={className}
+  >
+    <path
+      fillRule="evenodd"
+      clipRule="evenodd"
```
MattIPv4, line 18:
```suggestion
      fill="currentColor"
```
To be consistent with other icons, default to using the text color

apps/website/src/icons/IconGithub.tsx
```diff
@@ -0,0 +1,24 @@
+import React from "react";
+
+type IconProps = {
+  size?: number;
+  className?: string;
+};
+
+const IconGithub: React.FC<IconProps> = ({ size = 24, className }) => (
```
MattIPv4, line 8:
```suggestion
const IconGitHub: React.FC<IconProps> = ({ size = 24, className }) => (
```
Super nitpick, GitHub's brand name has a capital `H` (would also apply to the file name)

apps/website/src/components/layout/footer/Footer.tsx
pjeweb: Yeah i do not like what prettier does here at all with the newlines, but I have it set to run on save. Not sure if I should disable it just for this ...

pjeweb: I'll just put a /* prettier-ignore */ in front of the jsx block!

apps/website/src/components/layout/footer/Footer.tsx
pjeweb: Would ask her, but going with your suggestion first

apps/website/src/components/layout/footer/Footer.tsx
MattIPv4: (I'll be honest, TIL that prettier was being used here, I don't think I've been running it)

PR #66, opened 2023-02-16
Port Alveus static content

MattIPv4:
Ports across the remaining "static" content pages from the old Alveus site (annual reports, collabs, donate, merch, contact). This should leave only the ambassador pages, events and the show & tell section left to implement.


apps/website/src/pages/merch-store.tsx
```diff
@@ -0,0 +1,84 @@
...
+      {/* Grow the last section to cover the page */}
+      <Section className="flex-grow" containerClassName="flex flex-wrap items-center">
+        {merch.map((items, idx) => (
+          <div key={idx} className="basis-full md:basis-1/2 p-4 flex flex-col items-center gap-8">
+            {Object.entries(items).map(([ key, item ]) => (
+              <Link
+                key={key}
+                href={item.link}
+                target="_blank"
+                rel="noreferrer"
```
pjeweb, line 61: Weird case but we might actually want to provide the referrer to the shops. I'm not sure if the shops have analytics for that but might be interesting to see. Maybe they even support some get params to track that more directly?

pjeweb (approved):
Looks great as alaways.

apps/website/src/pages/merch-store.tsx
MattIPv4: Hm, interesting point. If it gets raised as a concern we can switch this over to noopener :nod:

PR #69, opened 2023-02-18
Implement show and tell

### LINKED ISSUE #5: Show and tell
## Features

- [x] "Seen on stream" marker
- [x] Approval process (mods review posts)
- [x] Login => probably fine using twitch OAuth instead of manual name/password login
- [x] Allow posting without an account => tbd if needed
- [ ] Comments => rarely use, probably can be dropped

##  Each post will have
- [x] Name for the user (maybe make this something the user can change in their profile/settings)
- [x] Post Title
- [x]  ~~Feature Image/~~ Images (each with title, caption and description) => Upload to S3-compatible Storage
- [x] Richtext
  - [x] Sanitize HTML
- [x] Video-Links (Youtube?!) => Currently supports Streamables and Imgur videos aswell. Check URL, extract video id (https://gist.github.com/rodrigoborgesdeoliveira/987683cfbfcc8d800192da1e73adc486)

##  View/Actions
- [x] Public list of entries
  - [x]  Show image captions / alternative texts in carousel and lightbox
  - [ ]  Sync carousel and lightbox?
- [x] Public detail pages
- [ ] Public archive list without infinite loading for SEO
- [x] Presentation view optimized for showing on stream
  - [ ] Add buttons to move through the lightbox?
- For users:
  - [x] Entry form
    - [x] Implement video (embed) attachments (Youtube, Imgur video)
  - [x] List of own entries
  - [x] Edit entry
  - [x] Delete entry
  - [ ] Preview entry?
- For mods/admins:
     - [x] List all entries
     - [x] Approve post
     - [x] Remove/disable post
     - [x] Mark which post was last shown on stream
     - [ ] Assign anonymous posts to a user (esp. for migrated posts)
     - [ ] Preview entry?

##  Also consider:

- [ ] ~~Do we want to have a local dev stand-in replacement for S3?~~ => #81
- [x] Update README with S3 storage info (DO Spaces)
- [x] Regularly cleanup S3 that are not being used (check which files are referenced in db, send delete requests to S3 api)
- [ ] Migrating old content
- [ ] ~~Migrating old accounts~~ => Instead we migrate them as anonymous posts and allow admins to assign them to users later on

pjeweb:
Resolves #5


pjeweb:
This is currently also deployed at https://preview.alveus.gg/show-and-tell with working auth.

pjeweb:
There are still some things to fix here and there but I would like to get this giant PR out of the way ;)

apps/website/prisma/schema.prisma
```diff
@@ -241,4 +249,88 @@ model OutgoingWebhook {
...
+}
+
+model LinkAttachment {
+  id              String @id @default(cuid())
+  type            String
+  name            String
+  title           String
+  alternativeText String
+  caption         String
+  description     String
```
MattIPv4, line 263: What is the difference between these three fields? The UI seems to only show alt text + caption, and I wonder if those should be combined?

apps/website/prisma/schema.prisma
```diff
@@ -241,4 +249,88 @@ model OutgoingWebhook {
...
+  id                             String                          @id @default(cuid())
+  name                           String
+  title                          String
+  alternativeText                String
+  caption                        String
+  description                    String
+  url                            String
+  fileStorageObjectId            String                          @unique
+  fileStorageObject              FileStorageObject               @relation(fields: [fileStorageObjectId], references: [id])
+  showAndTellEntryLinkAttachment ShowAndTellEntryLinkAttachment?
```
MattIPv4, line 335: Through this relation, isn't there a relation to a LinkAttachment which already contains much of the metadata (title, description, etc.) that this model seems to be duplicating? Or am I misunderstanding the schema?

MattIPv4 (commented):
This looks great! Left a few thoughts from poking around.

One avenue for abuse I noticed -- if you add an image to a post and then remove it, it doesn't look like we delete it from S3? Meaning someone could sit on the submit page forever just adding images and removing them, using our S3 bucket as free storage? Equally, someone could add images in there and then close the tab, using our S3 bucket as free storage? Maybe we should only upload them to S3 when the post is actually submitted, or cleanup any images in the bucket that aren't tied to a post after a timeout (30 minutes maybe)?

apps/website/src/components/content/Section.tsx
```diff
@@ -1,40 +1,58 @@
...
-      </div>
+      {offsetParent && (
+        <div
+          className={`absolute inset-0 -z-10 ${opacity}`}
+          style={{
+            backgroundImage: `url(${texture})`,
+            backgroundSize: "32rem",
+          }}
+        />
+      )}
```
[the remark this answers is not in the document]
pjeweb: Ah yes, thats left-over. I had the show-an-tell at one point request native fullscreen for the whole document and not only for the section. Then I needed to make the entries container `fixed inset-0` but that did not work because the section was promoted to a offset parent because of the background image. That should be solved now, that I request native fullscreen for the entries container which has basically the same effect as `fixed inset-0` had.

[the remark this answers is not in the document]
pjeweb: Oh actually we still need this because you can get into the full screen mode without native fullscreen, e.g. if browsers do not support it or if gets closed by pressing ESC.

apps/website/prisma/schema.prisma
pjeweb: > One avenue for abuse I noticed -- if you add an image to a post and then remove it, it doesn't look like we delete it from S3? Meaning someone could sit on the submit page forever just adding images and removing them, using our S3 bucket as free storage? Equally, someone could add images in there and then close the tab, using our S3 bucket as free storage? Maybe we should only upload them to S3 when the post is actually submitted, or cleanup any images in the bucket that aren't tied to a post after a timeout (30 minutes maybe)?

Yep, thats on my todo list in #5 . That is also why I added the file storage object references to the database - to be able to clean them. One thing that hindered me from implementing this already is that I am unsure if its okay to use the newly introduced vercel cron jobs or if we would need another service (e.g. the long-running node server) for the cleanup task? 

If we were to upload all files on submit we would need to either orchestrate that in the client making the user wait or pass everything as multipart to the serverless function and relay the files to DO. I think its easier to upload them separately and clean up if they are not used.

For the possible abuse of the file storage until the file gets cleaned or approved I do not really have a solution. We can upload them without public access and only make them public after the post is submitted. We cant really wait for the approval though, because the reviewer would need to be able to see them to review them unless we use an api route to make  them viewable for authorized users.

This is all pretty new to me (that is serverless + s3), so feel free to suggest different approaches! :)

apps/website/prisma/schema.prisma
pjeweb: It's complicated and probably over-engineered, but there should not be any duplications.

- if a file upload is started, a `FileStorageObject` row also gets created to keep a reference for cleanup etc.
- an `ShowAndTellEntry` can have many attachments that are either image or links
- to implement this polymorphic m:m relation I've added the `ShowAndTellEntryLinkAttachment` table that should have been named `ShowAndTellEntryAttachment`. Prisma makes it a bit harder to keep the m:m table simple.
- link attachments and image attachments then have all the additional fields a files may have. to keep those entities generic I over-engineered them and added a bunch of fields we currently do not use. I basically copied what the old page did. but they did not really get used there either
- if a new image attachment is created, it is connected with the uploaded file storage object if it is valid
- at this point we also check the image to be valid and extract some info (mime type, width, height)
- if the image is not valid it gets deleted
- if the image meta data is valid, it gets written to the `ImageMetadata` connected to the `FileStorageObject`
- the `FileStorageObject` is generic though and can be any file type so I kept the metadata in a separate table

alveus_attachments_er

apps/website/src/components/content/Section.tsx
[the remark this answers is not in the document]
MattIPv4: Oh I see why this is needed now -- it makes more sense to me, but I worry that this is a fragile way to handle "fullscreen"? If any changes down the road happened to the layout etc. above this, it could easily end up not working as intended again?

To me, it feels like fullscreen would be better off being implemented in a way where it hoists itself far higher in the stack (basically right to the top of the page) -- maybe using a React portal?

apps/website/src/components/shared/VideoPlatformIcon.tsx
```diff
@@ -0,0 +1,30 @@
+import React from "react";
+import Image from "next/image";
+import youtubeIcon from "simple-icons/icons/youtube.svg";
+import imgurIcon from "simple-icons/icons/imgur.svg";
+import { VideoCameraIcon } from "@heroicons/react/20/solid";
+import streamableIcon from "@/assets/icons/streamable.svg";
```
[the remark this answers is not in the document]
MattIPv4: 👍 Seems fine to me, probably best to file an issue to standardise icons here and anywhere else that is doing it a bit different.

apps/website/prisma/schema.prisma
pjeweb: I renamed the m:m relation table from `ShowAndTellEntryLinkAttachment` to `ShowAndTellEntryAttachment`

MattIPv4: (fwiw, it might be worth dropping this explanation into a docs MD file somewhere in the repo, for future travellers?)

That diagram makes it much easier to follow -- I'd got tripped up by the fact the Prisma schema defines the relationship on both models, not just the "parent" model (e.g. the EntryAttachment), so was thinking an ImageAttachment was tied to a LinkAttachment somehow.

apps/website/src/components/content/Section.tsx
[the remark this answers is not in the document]
pjeweb: Yeah not super happy with that either, but I also did not really want to re-render all the content in the portal and handle the intersection observer twice etc. At that point I would just remove the "hot swapping" between views and make them separate completely and put them on different routes. I would wait and see if the feature even gets used. And if we change something in the layout that breaks this its not that bad to fix.

pjeweb:
Main points should be fixed now. File cleanup should work now, though vercel cron jobs wont work in the preview branch, they have to be triggered manually with a simple HTTP GET to https://preview.alveus.gg/api/cron

PR #70, opened 2023-02-20
Events page + collaborations cleanup

MattIPv4:


pjeweb (changes requested):
I love the video thumbnails and everything! Would never have thought to write my own ffmpeg loader.

Two things:
1) I would prefer not to have a different eslint parser config just for the one file. The eslint typescript parser parses the file fine only the cjs requires get flagged. I would propose using the `.cjs` file ending and adding an override for `**/*.cjs` in the main `.eslintrc.json`
```
  "overrides": [
    {
      "files": ["**/*.cjs"],
      "rules": {
        "@typescript-eslint/no-var-requires": "off"
      }
    }
  ]
```

2) I've used prettier for code formatting in this project. I like having something enforce code style and not having to think about it, so i run prettier on my auto-format shortcut in the IDE. I would prefer keeping it on, but that also means I always get diffs if i work in files from contributors that have not run prettier. Would you be open to set that up in your IDE as well? We could also set up pre-commit hooks or something like that to enforce use and document it better in the README ... (https://www.jetbrains.com/help/webstorm/prettier.html#ws_prettier_default_formatter)

apps/website/src/build/video-loader.js
```diff
@@ -0,0 +1,189 @@
+const { tmpdir } = require("os");
+const { randomBytes } = require("crypto");
+const { join } = require("path");
+const { exec: exexSync } = require("child_process");
```
pjeweb, line 4: Should this be `execSync`?

MattIPv4:
Makes sense to just have an override for all cjs files 👍 I removed an existing eslint disable from the tailwind cjs config.

I thought I had enabled prettier, but I guess not 🤦 Everything I've touched in this PR should now be formatted using it (it does make some interesting decisions that cause a lot of excessive nesting).

I think a pre-commit hook would be neat, and running linting as part of that too.

pjeweb (approved):
lgtm

PR #72, opened 2023-02-21
Assorted site optimisations and cleanup

MattIPv4:


pjeweb (approved):
Looks great. Love the img placeholders. Feels a lot smoother. I did not know next/vercel sends `cache-control: public, max-age=0, must-revalidate` for the icons and manifest. So I guess the cache busting is not needed?!

MattIPv4:
Yeah, it looks like the ask browsers to always revalidate, and tbh they shouldn't be changing very often anyway.

PR #78, opened 2023-03-02
Fix typo in 2021 annual report heading

NerdyTechy:
Only a small typo, just thought it would be easier to fix it then PR rather than create an issue for it


pjeweb:
thank you

PR #85, opened 2023-03-04
Make website installable (PWA / "install to home screen")

### LINKED ISSUE #84: Progressive Web App basics
To get web push notifications (see #1) on iOS we will need to support "Add to Home screen" meaning the website should be a progressive web app.

Minimal requirements:

- [x] The website must be served over HTTPS.
- [ ] The website must have a valid web app manifest file that includes the following properties:
  - "name": the name of the app
  - "short_name": a short name for the app
  - "start_url": the URL that should be loaded when the app is launched
  - "display": "standalone" or "fullscreen"
- [ ] The website must have at least one service worker registered and activated.

pjeweb:
Resolves #84 

- generate web manifest dynamically at build time
- add basic service worker with fetch handler
- refactor/cleanup build scripts
- move build scripts from src/build/ to build-scripts as they are not source files that get built, but separate build scripts that will be used run in node without being transpiled through from typescript


pjeweb:
This is deployed to https://preview.alveus.gg/ for easier testing. Works for me on iOS 16.3.1, Desktop Chrome (Mac) and Chrome on Android.

PR #88, opened 2023-03-05
Show and Tell: migration script

pjeweb:
parses html from the live page, transfers source images to the file storage and create database entries


pjeweb:
hopefully we wont need this anymore. closing for now, but keeping the branch just in case

PR #89, opened 2023-03-06
Ambassador clips + collabs/home/about YouTube lightbox

MattIPv4:


apps/website/src/components/content/YouTubeLightbox.tsx
```diff
@@ -0,0 +1,240 @@
...
+};
+
+type TriggerProps = {
+  videoId: string;
+  caption?: string;
+  className?: string;
+  children: React.ReactNode;
+};
+
+const getTrigger = (id: string) => {
```
pjeweb, line 44: nitpick but i would suggest naming this `createTrigger` to make clear its a factory

pjeweb (approved):
looking great as always

PR #92, opened 2023-03-10
Update socials content in footer

pjeweb:


pjeweb:
ahh have some pngs too many, i'll remove them

MattIPv4:
On mobile, can we change the padding for the stay updated block to:

```
padding-top: 0;
padding-bottom: 6rem;
```

I think that looks more balanced?

image

pjeweb:
> On mobile, can we change the padding for the stay updated block to

should be fixed. i had added some negative margin so the video has some more space on desktop. removed that for mobile

MattIPv4 (approved):
(for future me if I ever get super bored, look at what more the video loader can do to optimize here -- 5mb for 16s at 720p doesn't seem great)

pjeweb:
interesting. when i tested it was 2.7mb in production.

i think the biggest "problem" with the video not being reduced much is that it's not 16:9 so it wont get scaled much with the 720p preset. but i did not really know how to implement loader parameters. also might be an idea to go even further down in fps? cinematic 24fps?

MattIPv4:
Yeah, looking at loader parameters is probably the way to go -- probably default to 730p30 at a decent bitrate, but allow toggling to a "standard" quality that is 540p24 and a lower bitrate or smth

PR #93, opened 2023-03-10
feat(nav): change styles for dropdown and mobile nav

pjeweb:


MattIPv4:
👀 I'm not sure I'm a fan of using a background on the top-level items -- it feels rather out of place to me. I think it looks great for the dropdowns though, and on mobile.

pjeweb:
> 👀 I'm not sure I'm a fan of using a background on the top-level items -- it feels rather out of place to me. I think it looks great for the dropdowns though, and on mobile.

was not supposed to be on top-level, uppsi, will check

MattIPv4:
image

Sign in boi is still unhappy, but the rest looks better now

MattIPv4 (approved):

PR #99, opened 2023-03-11
fix(ambassadors): Oliver arrived in June 2021

MattIPv4:
https://twitter.com/AlveusSanctuary/status/1404886840830676995 / https://www.alveussanctuary.org/ambassadors/oliver/


pjeweb (approved):
i mean, i cant really say what is right, but i agree it was after nugget (april 2021) and before the twitter post in june 2021...

Issue #98, opened 2023-03-11
Prefer words for months in Dates to prevent date format confusion ISO/US 


pjeweb:

pjeweb:
fixed in #100 100

PR #100, opened 2023-03-11
feat(dates): use long date format for end users, keep short format fo…

pjeweb:
…r admins

this is to prevent confusion between date formats


pjeweb:
fixes #98

PR #104, opened 2023-03-11
Add PO Box page + link on Donate page

### LINKED ISSUE #96: Website: short link / page with P.O. box info

MattIPv4:
Resolves #96


MattIPv4:
Ended up going with having a dedicated page and linking to it from donate, rather than having it on donate itself. Figured that's more future-proof if additional guidance etc. needs to be put on the page in the future (size restrictions, etc.)

pjeweb (approved):
beautiful and good call with the separate page imho

Issue #52, opened 2023-02-12
Replace wordpress site (alveussanctuary.org)


pjeweb:
# Pre-Checklist

- [x] configure .org domain in vercel => vercel wont set up the domain unless the dns points to it ...
- [x] ~~google analytics tag~~
- [x] invite space to all teams (vercel, DO, planetscale, github)
- [x] make a list of services with their costs
- [x] lower TTL in DNS
- [x] `s/Alveus.gg/Alveus/` => #107
- [x] Sitemap?
- [ ] Alveus/Space has to set up a twitch dev account to use for OAuth
- [x] Replace alveussanctuary.org link with copyright smallprint in page footer
- [x] use .org privacy policy and check whats missing
- [x] Add .org to OAuth app

# Go-Live-Checklist

- [x] Add .org in Vercel
- [x] Unset `NEXT_PUBLIC_NOINDEX` for production
- [x] Update `NEXT_PUBLIC_BASE_URL`, `NEXT_PUBLIC_VERCEL_URL`, `VERCEL_URL` for production
- [x] Change DNS records

| Type | Name | Value |
| --- | --- | --- |
| `CNAME` | `www` | `cname.vercel-dns.com.` |
| `A` | `@` | `76.76.21.21` |
| `A` | `www` | **remove if set** |
| `AAAA` | `www` | **remove if set** |
| `AAAA` | `@` | **remove if set** |

- [ ] Check OAuth and S3 uploads work with .org domain
- [x] Check robots.txt, noindex etc.
- [x] Check top google search results still resolve correctly

# Move over website content

- [x] Home page
- [x] About pages
- [x] Ambassadors -> Replace with something based on the Twitch Extension?
- [x] Show and Tell ->  in progress, #5 
- [x] Events -> This probably needs to be reconsidered. The last "big" event was 2021 ... Maybe we can have some sort of blog-ish documentation for events still? Needs more thinking
- [x] Contact Us -> Does this need to a main menu item? Can we move this to the footer?
- [x] Donate -> Needs to be redone. I would prefer not to have so many different looking embeds, also amazon smile is sunsetting.
- [x] Merch -> Would be nice to have a nice Merch scroller of the SE Shopify merch as wenn as the plushies

MattIPv4:
After a bit of back and forth, _our_ current idea is:

```
Home (will be similar to the current Alveus homepage imo, but with a player on the right and a CTA to /live?)
Watch
Explore (for community/educational items)
  - Ambassadors
[...]

PR #107, opened 2023-03-11
Replace Alveus.gg with Alveus/Alveus Sanctuary (Inc.)

pjeweb:
see #52

### DIFF
--- a/apps/website/build-scripts/create-manifest.mjs
+++ b/apps/website/build-scripts/create-manifest.mjs
@@ -3,8 +3,8 @@ import "./env.mjs";
 import { env } from "../src/env/client.mjs";
 
 const manifest = {
-  name: "Alveus.gg",
-  short_name: "Alveus.gg",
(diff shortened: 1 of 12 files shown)


MattIPv4:
👀 `site.webmanifest` needs updating. There's also an alveus.gg email in the privacy policy, but that's probably a separate thing to address.

MattIPv4:
Oh, and another item for pre-launch, update the footer (replace the alveussanctuary.org link with copyright)

pjeweb:
> 👀 `site.webmanifest` needs updating.
`site.webmanifest` will get regenerated on `build`

> There's also an alveus.gg email in the privacy policy, but that's probably a separate thing to address.
yeah would probably just use the old policy or diff them => #52

> Oh, and another item for pre-launch, update the footer (replace the alveussanctuary.org link with copyright)
I've added that to the checklist => #52

MattIPv4 (approved):

PR #109, opened 2023-03-12
Use image + icon for YouTube preview, no iframe

MattIPv4:


pjeweb:
can we maybe extract the Preview component into a separate component? i would love to use that for show and tell as well.

pjeweb (approved):

PR #124, opened 2023-03-23
Remove WIP explore page

MattIPv4:
Removes the WIP explore page, and all its components/assets, as it relies on pulling data from the now private/archived data repository. This is blocking the site building (e.g. #123).


pjeweb (approved):
Did not think about that. Oops

PR #125, opened 2023-03-23
Cleanup icons, update deps, misc cleanup

pjeweb:


MattIPv4 (commented):
Is it worth moving `IconProps` to a dedicated file, and maybe even having a dedicated typedef for Icons generally (those props + specifically returning an SVG)? We currently have it duped in every file which is probably not best (my doing).

MattIPv4 (approved):

PR #131, opened 2023-03-26
Add copyright, updated privacy policy

MattIPv4:


MattIPv4:
Note to self, figure out nice wrapping on mobile (or change the heading)

MattIPv4:
Other note to self, probably needs legal review

pjeweb:
> Note to self, figure out nice wrapping on mobile (or change the heading)

hyphens-auto should do the trick. has good browser support now. https://tailwindcss.com/docs/hyphens

MattIPv4:
Ended up just going with standard word breaking -- hyphens causes all words to be hyphenated, rather than preferring standard breaking rules.

pjeweb (approved):

PR #140, opened 2023-04-01
Consistent styling for in-content links

MattIPv4:


pjeweb (commented):
If you could figure how to prevent wrapping right before the icon (maybe using a non-breaking space `\u00A0`/`&nbsp;`).

ambassador iucn link wrapping just before icon

MattIPv4:
Hm, I'm not sure it's possible to stop it wrapping? Even if you `display: inline` it, it will still wrap?

apps/website/src/components/content/Link.tsx
```diff
@@ -0,0 +1,46 @@
...
+
+  return (
+    <NextLink {...props}>
+      {children}
+      {external && !custom && (
+        <IconExternal
+          size="0.75em"
+          className="ml-1 mr-0.5 inline-block align-baseline"
+        />
+      )}
```
pjeweb, line 41: Try this:
```typescript
      {external && !custom && (
        <>
          {"\u00A0"}
          <IconExternal
            size="0.75em"
            className="ml-1 mr-0.5 inline-block align-baseline"
          />
        </>
      )}
```

MattIPv4: Yah, no dice with that, it still wraps

pjeweb: wait, formattig broke it again

pjeweb: no that works locally for me

pjeweb: i'm dumb i resized the browser

pjeweb (approved):

PR #142, opened 2023-04-01
Improve homepage initial load 

MattIPv4:


MattIPv4:
This still isn't much of a gain, having Twitch on the homepage is a massive hit to lighthouse scores etc., but I think its probably worth it to promote the stream for folks landing on the site

MattIPv4:
@pjeweb open to other ideas for optimizing the first load if you have any

pjeweb (approved):
I don not really have any more ideas right now. Lighthouse looks pretty good for what we have on the page. You could see if we can optimize the JS bundle size, but i dont know if there is much to gain.

PR #149, opened 2023-04-02
Update deps, simplify trpc base url

pjeweb:
- update deps (biggest one is typescript 5)
- some more small fixes


pjeweb:
tailwind autocompletion is broken for me. will have to check

pjeweb:
> tailwind autocompletion is broken for me. will have to check
I dropped the commit that moves the tailwind config to typescript as it is not supported by the jetbrains IDEs.

MattIPv4 (approved):
LGTM, apologies for the delay

PR #150, opened 2023-04-02
Rework nav structure, remove Live page

pjeweb:
- move live link to utility nav
- make utility nav more compact
- move user menu button to utility nav
- move Show and Tell to first level in nav


MattIPv4:
Hm, I'm not sure I'm a fan of all the changes here:

It feels like the utility nav is getting rather cumbersome, it was much cleaner visually when it was just icons.

I'm also not sure it feels right to put show and tell as a top-level item. If anything, I think Animal Quest would be better to surface as it has far more educational value to site visitors who might not know about Alveus. Those that're wanting to view/contribute to show and tell already know about it and have intent, so they'll find it in a dropdown.

pjeweb:
I am fine with [this version](https://alveusgg-website-m2ugl9fsd-alveus.vercel.app/), but I do not really fancy the external link in the main nav. Also we will run into the same problem with the crammed utility nav if we add the notifications button there. So sooner or later we will have to find a solution for all those features.

How important do we think the socials are in the header? Its nice as a "linktree" alternative, but we also have them on each page in the footer …

MattIPv4:
> but I do not really fancy the external link in the main nav.

I would have it still go to /live (and the other places that link to it to), and have that be a redirect defined in the Next.js config. That feels cleaner to me, and ensures folks who already knew about that page etc. get sent to the right place and whatnot. And I feel that sits nicely on the line between being interval vs. external

> we will run into the same problem with the crammed utility nav if we add the notifications button there

I think there are ways to design around that so it's far less of an issue, for example have it just be an bell icon in the nav, and when you hover we show a tooltip or slide the icon to the side and show text if needed.

MattIPv4:
> How important do we think the socials are in the header?

We don't have click tracking data, but I imagine pretty important. Users don't generally scroll on pages, and even less so all the way down to the footer. Anything above the fold is generally super valuable.

pjeweb:
> > but I do not really fancy the external link in the main nav.
> 
> I would have it still go to /live (and the other places that link to it to), and have that be a redirect defined in the Next.js config. That feels cleaner to me, and ensures folks who already knew about that page etc. get sent to the right place and whatnot. And I feel that sits nicely on the line between being interval vs. external

yeah i understand the appeal and would be fine with that but on the other hand its worse UX imho to have a different URL show on mouseover then what you end up at. but maybe thats just me. not something i am that particular about though.

> > we will run into the same problem with the crammed utility nav if we add the notifications button there
> 
> I think there are ways to design around that so it's far less of an issue, for example have it just be an bell icon in the nav, and when you hover we show a tooltip or slide the icon to the side and show text if needed.

yeah maybe. i feel like uniformity hurts UX for both socials and other features. it looks cleaner, but its also harder to see what is what at a glance. do you like the vertical separator lines?

MattIPv4:
> its worse UX imho to have a different URL show on mouseover then what you end up at. but maybe thats just me. not something i am that particular about though.

I generally find that to be okay, assuming the URL shown is still obvious in what it's doing (e.g. /live is probably going to take me to the live content) -- it also allows folks to copy the "friendly"/short link from our side and use that, allowing us to change the destination down the road again if needed

> do you like the vertical separator lines?

Yah, I think splitting the socials from other items with a vertical separator makes sense

PR #151, opened 2023-04-03
Remove Live page, move login to utility nav

pjeweb:


apps/website/src/pages/index.tsx
```diff
@@ -240,6 +240,9 @@ const Home: NextPage = () => {
               <Link
                 className="inline-block rounded-full border-2 border-white px-4 py-2 text-lg transition-colors hover:border-alveus-tan hover:bg-alveus-tan hover:text-alveus-green"
                 href="/live"
+                target="_blank"
+                rel="noreferrer"
+                title="Watch Live on Twitch.tv"
```
MattIPv4, line 245: I don't think this needs a title, as the link already has text content?

MattIPv4 (approved):

PR #157, opened 2023-04-06
Implement custom consent handling

### LINKED ISSUE #148: Build custom consent solution
- Build popup/banner for consent that tracks a state
  - EU should get fullscreen popup on first load
  - Everyone else should get banner experience
- Build a hook that can be used to get consent for an item
- Lock Twitch and YouTube embeds behind consent

MattIPv4:
Resolves #148


MattIPv4:
@pjeweb lmk what you think of this so far!

(You can reset your consent at any time by running `consent.updateConsent(Object.keys(consent.consent).reduce((obj, key) => ({ ...obj, [key]: false }), {}))` in console)

apps/website/src/hooks/consent.tsx
```diff
@@ -0,0 +1,106 @@
...
+            : (existing || defaultConsent)[key],
+      }),
+      {}
+    )
+  ) as Consent;
+
+export type ConsentContext = Readonly<{
+  consent: Consent;
+  updateConsent: (consent: Partial<Consent>) => void;
+  consentLoaded: boolean;
```
MattIPv4, line 41: I never know with hooks whether the pattern should be that properties have names that indicate where they are from (as they are currently), or whether they should have clean names (e.g. `status, update, loaded` here) -- thoughts?

MattIPv4: Went with clean names, it was getting messy with them all having the word consent in it

MattIPv4:
TODO: YouTube popups need to prompt for consent like the Twitch embed does

MattIPv4:
TODO: You can't tab to the checkboxes

MattIPv4:
TODO: set title attr on the corner button, so it shows on hover

pjeweb (commented):
beautiful design both visually and code-wise! just a few things:
- our privacy policy should be linked in the modal and readable without having to decide, usually you would not show the modal on that page to make sure its readable
- we probably want to link the privacy policies of google/yt and twitch right in the modal as well
- do we still need the global window.consent?

MattIPv4:
> do we still need the global window.consent?

I exposed it just in case we ever have anything non-React that needs access to it. Happy to remove it if you disagree with having it exposed though?

pjeweb (approved):

PR #158, opened 2023-04-07
Disable animations/autoplay for reduced motion

### LINKED ISSUE #155: Accessiblity: Disable autoplay and animations if `prefers-reduced-motion` is set to `reduce`
https://developer.mozilla.org/en-US/docs/Web/CSS/@media/prefers-reduced-motion

MattIPv4:
Resolves #155


MattIPv4:
Lmk if there is anything else this should capture -- I haven't done the Twitch embed on the homepage here as I didn't want to cause a conflict with #157

pjeweb (approved):
nice

PR #160, opened 2023-04-07
Show ambassador images on animal quest page

### LINKED ISSUE #159: Add ambassador images to animal quest
Idea from @pjeweb:

image

If we do it, we need to consider how it handles episodes with multiple ambassadors (I'm thinking picture-in-picture bubbles of different sizes), and what to do for the Falcon episode (I'm thinking just show a translucent version of the thumbnail image).

MattIPv4:
Resolves #159


pjeweb (approved):
I like it a lot

PR #163, opened 2023-04-07
Fix tab navigation

### LINKED ISSUE #153: Accessiblity: Keyboard tab navigation
Problems I have encountered:
- nav dropdown items cant be tabbed to (or any other typical key like arrow keys)
- the twitch embed on the homepage can be tabbed into while clicks are redirected to twitch.tv

MattIPv4:
Resolves #153

### DIFF
--- a/apps/website/src/components/layout/navbar/DesktopMenu.tsx
+++ b/apps/website/src/components/layout/navbar/DesktopMenu.tsx
@@ -1,8 +1,7 @@
 import Link from "next/link";
 import { Menu, Transition } from "@headlessui/react";
 import { signIn, signOut, useSession } from "next-auth/react";
-import type { AnchorHTMLAttributes, ReactNode } from "react";
-import React, { Fragment } from "react";
+import React, { Children, cloneElement, forwardRef, Fragment } from "react";
 
 import {
   mainNavStructure,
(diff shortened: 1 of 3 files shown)


apps/website/src/components/layout/navbar/DesktopMenu.tsx
```diff
@@ -36,13 +35,22 @@ const DropdownMenuItems: typeof Menu.Items = ({ ...props }) => (
...
 DropdownMenuItems.displayName = "DropdownMenuItems";
 
+const DropdownMenuItem: React.FC<{ children: React.ReactElement }> = forwardRef(
+  ({ children, ...props }, ref) => (
+    <li {...props}>
+      {Children.map(children, (child) => cloneElement(child, { ref }))}
+    </li>
+  )
+);
+DropdownMenuItem.displayName = "DropdownMenuItem";
```
MattIPv4, line 52: This feels terrible, but it looks like the ref needs to make it all the way down to the `a` element for the keyboard navigation to work, and the native `li` element doesn't pass the ref down (duh)

pjeweb: yeah not great but unless we drop headlessui there i dont see many other options

pjeweb (approved):

PR #164, opened 2023-04-07
Remove root service worker

pjeweb:
Apple does not require a full PWA for web push so we do not need to have a SW. The manifest is enough and the push service worker is registered separately.

### DIFF
--- a/apps/website/src/pages/_app.tsx
+++ b/apps/website/src/pages/_app.tsx
@@ -6,10 +6,10 @@ import { SSRProvider } from "react-aria";
 import { trpc } from "@/utils/trpc";
 import "@/styles/globals.css";
 import Layout from "@/components/layout/Layout";
-import registerServiceWorker from "@/utils/sw";
+import { unregisterServiceWorker } from "@/utils/sw";
 import { ConsentProvider } from "@/hooks/consent";
 
-registerServiceWorker("/RootServiceWorker.js");
+unregisterServiceWorker("/RootServiceWorker.js");
 
 const AlveusGgWebsiteApp: AppType<{ session: Session | null }> = ({
   Component,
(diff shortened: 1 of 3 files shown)


apps/website/src/pages/_app.tsx
```diff
@@ -6,10 +6,10 @@ import { SSRProvider } from "react-aria";
 import { trpc } from "@/utils/trpc";
 import "@/styles/globals.css";
 import Layout from "@/components/layout/Layout";
-import registerServiceWorker from "@/utils/sw";
+//import registerServiceWorker from "@/utils/sw";
 import { ConsentProvider } from "@/hooks/consent";
 
-registerServiceWorker("/RootServiceWorker.js");
+//registerServiceWorker("/RootServiceWorker.js");
```
MattIPv4, line 12: Do we want a quick bit of code to unregister it for existing folks?

pjeweb: yeah probably best

apps/website/src/utils/sw.ts
```diff
@@ -12,3 +12,13 @@ export default function registerServiceWorker(path: string, scope = "/") {
       });
   }
 }
+
+export function unregisterServiceWorker(path: string, scope = "/") {
+  if (typeof navigator !== "undefined" && "serviceWorker" in navigator) {
+    navigator.serviceWorker.getRegistration(scope).then((registration) => {
+      if (registration) {
```
MattIPv4, line 19:
```suggestion
      if (registration?.active && new URL(registration.active.scriptURL).pathname === path) {
```
If we want to enforce the path?

pjeweb: I think just using the scope is enough?! Ij ust forgot to remove the path param while switching between the branches.

MattIPv4 (approved):

PR #181, opened 2023-04-08
Add social + ambassador redirects

### LINKED ISSUE #178: Redirects for ambassadors
I think it would be nice to have `/x` redirect to `/ambassadors/x` so we could link them in SE commands etc.

With the .gg domain: `https://alveus.gg/fenn`

MattIPv4:
Resolves #178


MattIPv4:
@pjeweb I do believe the answer may be no, doing redirects this way does not work

pjeweb:
i'm at a loss then. i would probably just keep them in two places and maybe add a bit of jsdoc to at least somewhat use types to check they match?!

```javascript
/** @type {import("./data/ambassadors.mjs").AmbassadorKey[]} */
const ambassadorNames = ["abbott", "appa", "barbaraBakedBean", "bb", …

/** @type {import("next").NextConfig} */
const config = {
  …
  redirects: async () => [
    ...ambassadorNames.map((name) => ({
      source: "/" + name,
      destination: "/ambassadors/" + name,
      permanent: false,
    })),
```

pjeweb (approved):
nice solution!

MattIPv4:
It's not the greatest as it adds more build tooling, but esbuild was already in the toolchain for tsup afaik, so not the worst.

pjeweb:
yeah does not feel too bad. i was thinking we might also use what tailwind introduced in 3.3 to compile JIT:
https://github.com/unjs/jiti

PR #185, opened 2023-04-10
Use shared data for builds

MattIPv4:


pjeweb:
what do you think about publishing the data as `@alveusgg/data` on npm, transpilied as js+d.ts? then we could skip the build step and git checkout?

MattIPv4:
That makes senes to me -- do you want to take a look at doing that over in the data repo, and I'll fix up this build issue?

pjeweb:
yeah i'll take a stab at it

pjeweb (approved):
fabulous

Issue #59, opened 2023-02-13
Forms (Giveaways)


pjeweb:
- [x] Basic implementation
- [x] Allow users to view~~edit~~ their entry
- [x] encrypt PII => #97
- [x] Create/edit/remove giveaways
- [x] Make rules editable
- [x] Admin CRUD giveaway entries (remove test entries etc.)
- [x] Add privacy/gdpr checkbox
- [ ] Make individual checks configurable
   - [ ] enable/disable checks
   - [ ] change URLs/Account to follow/sub per Check?
- [x] Implement better checks (actually checking follow/subscribe status)
  - [x] Twitch follow
  - [ ] ~~YT subscription~~ (is against YT ToS)
  - [ ] ~~Twitter follow~~ (API is too expensive);  alternative: OAuth or [ follow intent link + ask for username + check manually ]
  - [ ] ~~IG follow~~ (No API); alternative: OAuth for user name
  - [ ] ~~TikTok follow~~ (No API); alternative: use login kit (auth) or ask for username instead? https://developers.tiktok.com/doc/login-kit-web/
- [ ] Add a captcha or question to make free entry harder? (e.g. What type of animal is Stompy -> Emu, Ostrich, Elephant)
- [ ] Improve caching

Possible bugs:

- Check webhooks work with unencrypted data
- Check mailing addresses actually being deleted when anonymising entries

pjeweb:
Will close this since its been used as is without any need for the open points

PR #192, opened 2023-04-12
Encrypt PII, finalize Giveaways implementation

### LINKED ISSUE #97: Encrypt PII at rest in the DB
We should be encrypting PII in the DB to ensure that if the DB were popped there'd be no usable PII to extract.

https://www.npmjs.com/package/prisma-field-encryption might be a good candidate for doing this?

### LINKED ISSUE #191: Testing
Do we want unit/integration tests?

I think we don't need automated tests for most parts of the website, but maybe for:
- web push (vapid, content encryption etc.)
- data encryption for database

If so, i would probably go with vitest (see https://github.com/vercel/next.js/tree/canary/examples/with-vitest ).

if we want want end-to-end tests i would probably go with playwright (see https://github.com/vercel/next.js/tree/canary/examples/with-playwright ).

pjeweb:
This is a big PR, happy to split it up for easier review, but it is quite intermingled. It might be easier to review commit by commit.

I pushed this branch to preivew (https://preview.alveus.gg/) so we can test the admin functionality.

- Introduce vitest for automated tests (manually run with `pnpm vitest`) vitest
- Adds encryption utils (AES256)
- Adds utils to encrypt database records (used for giveaways and mailing addresses)
- Support ordered lists in Markdown component (used for giveaways rules)
- Tweaks some shared form components (for giveaway admin)
- Adds component and util for `datetime-local` input fields (used for giveaway admin)
- Adds ModalDialog component (used for giveaway admin)
- Adds util to slugify strings (used for the public giveaway url)
- Some misc fixes and cleanups
- Minor/path level dep updates
- Giveaways: Implement at-rest encryption
- Giveaways admin: Create, Edit, Delete, Purge entries, Anonymize entries (remove PII)

delete-purge-anonymize-giveaway

create-edit-giveaway

Resolves #97
Resolves #191
See #59


apps/website/src/server/db/giveaways.ts
```diff
@@ -0,0 +1,130 @@
...
+
+const entryEncryptFields = ["givenName", "familyName"];
+
+const mailingAddressEncryptFields = [
+  "addressLine1",
+  "addressLine2",
+  "postalCode",
+  "city",
+  "state",
+];
```
pjeweb, line 16: add country?!

MattIPv4: 👀 Yeah I would assume so, anything identifiable about the user

pjeweb:
@MattIPv4 if you want to test this on preview, you'd need to login once so i can give you perms in the new DB branch.

apps/website/src/components/layout/Layout.tsx
```diff
@@ -25,10 +24,10 @@ export const ptSerif = PT_Serif({
...
 
 const Layout: React.FC<LayoutProps> = ({ children }) => {
-  const router = useRouter();
-  if (router.pathname.endsWith("/attachment")) {
-    return <>{children}</>;
-  }
+  // Add fonts to body for portals that do not attach to #app
+  useEffect(() => {
+    document.body.classList.add(...fonts.split(" "));
+  }, []);
```
MattIPv4, line 30: Could you update the consent overlay to not set these now, as we're setting them here globally?

pjeweb: Oh I forgot we already have that dialog. Sure. Or I should we use your way for the ModalDialog? Not sure whats better

MattIPv4: Not sure I have much of a preference either way -- this is probably easier so we don't have to think about it for anything else in the future that is portal'd

apps/website/src/server/utils/encryption.ts
```diff
@@ -0,0 +1,124 @@
...
+
+    const decryptedBuffer = await crypto.subtle.decrypt(
+      { name: ENCRYPTION_ALGORITHM_NAME, iv, tagLength: 128 },
+      saltedEncryptionKey,
+      cipherText
+    );
+    return decoder.decode(decryptedBuffer);
+  } catch (e) {
+    return "";
+  }
```
MattIPv4, line 123: Do we not want to surface an error if a bad decryption happens? I worry this risks some level of data loss if a key were accidentally changed etc.?

apps/website/src/server/db/encryption.ts
```diff
@@ -0,0 +1,61 @@
...
+  const salt = await generateSalt();
+  const key = await createSaltedEncryptionKey(salt);
+
+  const result: RecordType & { salt: string } = Object.assign({}, input, {
+    salt: encodeSalt(salt),
+  });
+
+  for (const fieldName of fields) {
+    const value = input[fieldName];
+    if (typeof value === "string") {
```
MattIPv4, line 52: Should we instead throw an error if encryption is requested on a non-string?

apps/website/src/server/db/encryption.ts
```diff
@@ -0,0 +1,61 @@
...
+      const encrypted = await encrypt(value, key);
+      if (encrypted) {
```
MattIPv4, line 54: In what situation would encrypted be falsy?

apps/website/src/server/db/encryption.ts
```diff
@@ -0,0 +1,61 @@
...
+  EncryptionField extends string,
+  RecordType extends RecordToEncryptWithSalt<EncryptionField>
+>(input: RecordType, fields: Array<EncryptionField>) {
+  const salt = await decodeSalt(input.salt);
+  const key = await createSaltedEncryptionKey(salt);
+
+  const result = Object.assign({}, input);
+  for (const fieldName of fields) {
+    const encryptedValue = input[fieldName];
+    if (typeof encryptedValue === "string") {
```
MattIPv4, line 28: Ditto

apps/website/src/server/db/encryption.ts
```diff
@@ -0,0 +1,61 @@
...
+      const decrypted = await decrypt(encryptedValue, key);
+      if (decrypted) {
```
MattIPv4, line 30: Ditto

apps/website/src/pages/giveaways/[giveawayId]/rules.tsx
```diff
@@ -6,33 +6,21 @@ import type {
...
-  });
-}
+import { findActiveGiveaway } from "@/server/db/giveaways";
+import Markdown from "@/components/content/Markdown";
 
 export type GiveawayPageProps = InferGetServerSidePropsType<
   typeof getServerSideProps
 >;
 
 export const getServerSideProps: GetServerSideProps<{
```
Comment (MattIPv4) [inline apps/website/src/pages/giveaways/[giveawayId]/rules.tsx:21]: Maybe not for this PR, but are we still thinking about caching for this?

apps/website/src/pages/giveaways/[giveawayId]/index.tsx
```diff
@@ -7,33 +7,19 @@ import type {
...
-    },
-  });
-}
-
 export type GiveawayEntryWithAddress = GiveawayEntry & {
-  mailingAddress: MailingAddress;
+  mailingAddress: MailingAddress | null;
 };
 
 export const getServerSideProps: GetServerSideProps<{
```
Comment (MattIPv4) [inline apps/website/src/pages/giveaways/[giveawayId]/index.tsx:25]: Ditto

MattIPv4:
Haven't pulled down to poke around (I trust it actually works) -- read through everything though and it looks sound other than the few comments I left.

What's the plan for encrypting existing stuff?

Are we going to do encryption for user-account-related tables too (maybe a separate PR after)?

apps/website/src/pages/giveaways/[giveawayId]/rules.tsx
Comment (pjeweb) [reply apps/website/src/pages/giveaways/[giveawayId]/rules.tsx:21]: Yeah I have a few things I want to add to giveaways like a gdpr/privacy policy checkbox, but wanted to get this PR done first.

apps/website/src/pages/giveaways/[giveawayId]/index.tsx
Comment (pjeweb) [reply apps/website/src/pages/giveaways/[giveawayId]/index.tsx:25]: Ditto on the caching I assume?

apps/website/src/server/utils/encryption.ts
pjeweb: > What's the plan for encrypting existing stuff?
For the giveaway entries i wanted to just let the decryption util allow empty values (+ empty salt), and because we removed all the fields that would be encrypted (= PII) that would be the case for all of them.

> Are we going to do encryption for user-account-related tables too (maybe a separate PR after)?
Sure. Right now that would only be the email address though. Not sure if we need encryption for that at the cost of making auth potentially slower.

apps/website/src/server/db/encryption.ts
pjeweb: don't know, looks like refactoring leftover, will remove

apps/website/src/server/db/encryption.ts
pjeweb: ditto

pjeweb:
> What's the plan for encrypting existing stuff?

For the giveaway entries i wanted to just let the decryption util allow empty values (+ empty salt), and because we removed all the fields that would be encrypted (= PII) that would be the case for all of them.

> Are we going to do encryption for user-account-related tables too (maybe a separate PR after)?

Sure. Right now that would only be the email address though. Not sure if we need encryption for that at the cost of making auth potentially slower.

apps/website/src/pages/giveaways/[giveawayId]/index.tsx
Comment (MattIPv4) [reply apps/website/src/pages/giveaways/[giveawayId]/index.tsx]: Indeed

MattIPv4:
> Sure. Right now that would only be the email address though. Not sure if we need encryption for that at the cost of making auth potentially slower.

We probably should -- don't we also have access tokens from Twitch etc. that should be encrypted too?

Can be done once this is in though

MattIPv4 (approved):
All the logic seems sane, I trust it works as you intend it to

pjeweb:
> We probably should -- don't we also have access tokens from Twitch etc. that should be encrypted too?

Hmm maybe. I feel like with OAuth the user and us can revoke access anytime in case it gets compromised and an attacker would also need the client secret to use it. Also not as trivial as the entry data as we would need to customize the auth-db adapter. Doable but I don't really think its worth it. But I will open a new issue.

pjeweb:
(Note that the first deplyoment may fail as I also need to update the DB)

PR #197, opened 2023-04-14
Allow ambassadors to be grouped by enclosure

MattIPv4:


MattIPv4:
@pjeweb would appreicate a review on this (and the data PR) before I finalise them (updating the version in data + merging, and then updating the version used here)

pjeweb (commented):
Looks so good already! Would have no problems merging as is.

A few observations:
- if i switch between the tabs then click an ambassador then navigate back with browser ui the page does not switch back (the url and title changes, but not the content). there is no obvious error and its finicky to reproduce
- i would like the active tab highlight border to be a bit thicker
- i would prefer some more spacing between the groups
- Serranos image is very grainy/high contrast especially in the grouped view as the image is smaller there, but it does look subpar on the live site as well. i would also prefer a different hover style instead of the filter. we are not very consistent with that. personally i prefer the slight zoom we have on the donate buttons or the socials in the footer. does not have to be in this PR though, just something i noticed here

MattIPv4:
> if i switch between the tabs then click an ambassador then navigate back with browser ui the page does not switch back (the url and title changes, but not the content). there is no obvious error and its finicky to reproduce

That sounds almost like an issue w/ Next.js rather than anything we're doing, unless you mean the ambassador page loads but doesn't switch to the enclosure tab you were on before? If you can get a reliable reproduction definitely happy to dig either way.

> i would like the active tab highlight border to be a bit thicker

Sounds good, will do

> i would prefer some more spacing between the groups

Ditto

> Serranos image is very grainy/high contrast especially in the grouped view as the image is smaller there, but it does look subpar on the live site as well. i would also prefer a different hover style instead of the filter. we are not very consistent with that. personally i prefer the slight zoom we have on the donate buttons or the socials in the footer. does not have to be in this PR though, just something i noticed here

I am somewhat a fan of the filter that's applied (I'm generally a fan of higher saturation images, I find they look rather nice). That being said, I agree we should have a more consistent approach to it. Maybe we keep the filter, and add the zoom as well? And maybe this also comes back to the wider issue that we want to get better photos for some ambassadors? Either way, probably worth filing a dedicated issue for so we can think about it more?

pjeweb:
> If you can get a reliable reproduction definitely happy to dig either way.

We can look at the navigation issues later! I want to finish the giveaways today, but will try to reproduce.


> I'm generally a fan of higher saturation images, I find they look rather nice

Generally I like the "vibrant" images as well, but this looks broken to me, almost like pixel art or 8bit or something. It does look way better on my retina display. I think its just how its scaled.
Serrano

> Maybe we keep the filter, and add the zoom as well?

Sure. I'll open an issue later, maybe ;)

pjeweb (dismissed):

pjeweb:
btw the serrano pic looks way better in firefox

MattIPv4:
Ah okay, so sounds like a browser/color rendering issue. Weird

pjeweb (approved):

PR #210, opened 2023-04-18
Rename giveaways to forms

pjeweb:
Rename "Giveaways" to "Forms" as we use the forms for collection addresses outside of giveaways as well.

This PR is deployed to https://preview.alveus.gg/


pjeweb:
I tested this on preview. I think it works. We'll have to rename the DB schema at some point, but I did not really want to do this now.

pjeweb:
Okay, db is prepared and schema is renamed.

MattIPv4:
👀 `GiveawayChecks` still exists?

MattIPv4:
(Also as an unrelated note to here, does Prisma have a proper way to do in-code migrations?)

pjeweb:
> 👀 `GiveawayChecks` still exists?

yeah on purpose, those checks only make sense for giveaways imo

MattIPv4:
Ah I see, cool

MattIPv4 (approved):

pjeweb:
> (Also as an unrelated note to here, does Prisma have a proper way to do in-code migrations?)

yes, but they are not recommended with planetscale. I did not think it was worth doing it properly. planetscale suggests doing table renames in two steps:
1. add new table
2. in app code copy content to new schema / write to both tables
3. remove old table

i did that manually for this PR. I might do it more properly the next time. Most changes just work by running `pnpm prisma db push` though. I have not had to do many migrations otherwise yet.

Issue #198, opened 2023-04-15
Small styling inconsistencies


pjeweb:
- [x] External links in Show and Tell should be styled like other external links
- [x] Hover style: Use slight zoom for ambassadors
- [ ] Show and Tell dates are formatted based on the users locale, other dates are always formatted in US style

MattIPv4:
👀 Tackling the hover style issue

MattIPv4:
> Show and Tell dates are formatted based on the users locale, other dates are always formatted in US style

What's the preferred thing here, should everything use the localdate component, or should we just stick to doing US style everywhere (which I imagine is actually easier for folks to compensate for than being unsure if something is actually local or not)

pjeweb:
Yeah probably the same long form US format everywhere.

PR #214, opened 2023-04-20
Tweaks to show and tell

MattIPv4:
cc #198

### DIFF
--- a/apps/website/src/components/show-and-tell/ShowAndTellEntry.tsx
+++ b/apps/website/src/components/show-and-tell/ShowAndTellEntry.tsx
@@ -8,8 +8,16 @@ import type {
   ShowAndTellEntryAttachment,
 } from "@prisma/client";
 import Image from "next/image";
+import parse, {
+  domToReact,
+  Element,
+  type HTMLReactParserOptions,
+} from "html-react-parser";
+
 import { parseVideoUrl } from "@/utils/video-urls";
 import { notEmpty } from "@/utils/helpers";
+
+import Link from "@/components/content/Link";
 import { LocalDate } from "@/components/shared/LocalDate";
 import { ShowAndTellGallery } from "@/components/show-and-tell/gallery/ShowAndTellGallery";
 import { SeenOnStreamBadge } from "@/components/show-and-tell/SeenOnStreamBadge";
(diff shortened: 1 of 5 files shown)


apps/website/src/components/show-and-tell/ShowAndTellEntry.tsx
```diff
@@ -35,6 +43,18 @@ type ShowAndTellEntryProps = {
   showPermalink?: boolean;
 };
 
+const parseOptions: HTMLReactParserOptions = {
+  replace: (node) => {
```
MattIPv4, line 47: (As a note for the future, this lets us far more easily implement any sanitization if we need it)

apps/website/src/components/show-and-tell/ShowAndTellEntry.tsx
```diff
@@ -169,10 +189,9 @@ export const ShowAndTellEntry = forwardRef<
               isPresentationView ? "scrollbar-none max-h-[66vh] pb-6" : ""
             }`}
           >
-            <div
-              className="alveus-ugc max-w-[1100px] hyphens-auto leading-relaxed md:text-lg xl:text-2xl"
-              dangerouslySetInnerHTML={{ __html: entry.text }}
-            />
+            <div className="alveus-ugc max-w-[1100px] hyphens-auto leading-relaxed md:text-lg xl:text-2xl">
+              {parse(entry.text, parseOptions)}
```
MattIPv4, line 193: I did wonder if we wanted to put an error boundary around this specifically, to avoid breaking the whole page if someone gets malformed HTML into the system? Or are we happy with the manual review process catching that?

MattIPv4:
Noticed Mik's Stompy post in production had no content and so looked a bit odd, pushed a commit to address that (it's a bit of a hack).

I think the logic for checking the text content is sound, but would appreciate a double-check that I'm not inadvertently hiding legitimate posts that do have content.

pjeweb (approved):
that's beautiful

apps/website/src/components/show-and-tell/ShowAndTellEntry.tsx
pjeweb: That does sound good.  While I don't think it's likely to happen but it does definitely not hurt.

MattIPv4: https://github.com/alveusgg/alveusgg/issues/216

PR #221, opened 2023-04-20
Show and Tell: hide fullscreen button in small viewports

### LINKED ISSUE #215: Show and Tell: Fullscreen presentation mode on mobile
I think we should either improve scaling/styling for smaller viewports in the presentation mode (smaller text, hide sidebar) or just not show the toggle button.

pjeweb:
Resolves #215

### DIFF
--- a/apps/website/src/pages/show-and-tell/index.tsx
+++ b/apps/website/src/pages/show-and-tell/index.tsx
@@ -221,7 +221,7 @@ const ShowAndTellIndexPage: NextPage<ShowAndTellPageProps> = ({
           )}
 
           {isPresentationView && (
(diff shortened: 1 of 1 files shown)


MattIPv4 (approved):
Seems like a sensible fix, not letting a user get into the fullscreen mode on mobile

pjeweb:
> Seems like a sensible fix, not letting a user get into the fullscreen mode on mobile

I mean I would like to have the presentation mode for mobile akin to tiktok or reels, but i am not sure its worth and until then this an easy fix

MattIPv4:
Agree that seems like a good long-term goal but this is a good fix for now

PR #227, opened 2023-04-22
Show and Tell: sort by not seen on stream first

### LINKED ISSUE #225: Sort show and tell by seen on stream, then date
Sort the show and tell posts so that those that haven't been seen on stream are shown first, then those that have. Within each of those two groups, sort by date as usual.

pjeweb:
Resolves #225

### DIFF
--- a/apps/website/prisma/schema.prisma
+++ b/apps/website/prisma/schema.prisma
@@ -3,7 +3,7 @@
 
 generator client {
   provider        = "prisma-client-js"
-  previewFeatures = ["fieldReference", "fullTextSearch"]
+  previewFeatures = ["fieldReference", "fullTextSearch", "orderByNulls"]
 }
 
 datasource db {
--- a/apps/website/prisma/schema.prisma
+++ b/apps/website/prisma/schema.prisma
@@ -3,7 +3,7 @@
 
 generator client {
   provider        = "prisma-client-js"
-  previewFeatures = ["fieldReference", "fullTextSearch", "orderByNulls"]
+  previewFeatures = ["fieldReference", "fullTextSearch"]
 }
 
 datasource db {
--- a/apps/website/prisma/schema.prisma
+++ b/apps/website/prisma/schema.prisma
@@ -280,6 +280,7 @@ model ShowAndTellEntry {
   updatedAt      DateTime  @default(now())
   approvedAt     DateTime?
   seenOnStreamAt DateTime?
+  seenOnStream   Boolean   @default(false)
 
   attachments ShowAndTellEntryAttachment[]
   user        User?                        @relation(fields: [userId], references: [id])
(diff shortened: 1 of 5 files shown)


apps/website/src/server/db/show-and-tell.ts
```diff
@@ -27,6 +27,25 @@ export const whereApproved = {
...
+          { approvedAt: { lt: prisma.showAndTellEntry.fields.updatedAt } },
+        ],
+      }
+    : filter === "approved"
+    ? { approvedAt: { gte: prisma.showAndTellEntry.fields.updatedAt } }
+    : {};
+}
+
+const postOrderBy = [
+  { seenOnStreamAt: { sort: "desc", nulls: "first" } },
```
MattIPv4, line 44: Is there a way to sort just by whether this is null or not, rather than by the value in the column? It think its a bit confusing to have the posts that've been seen be stored by when they were seen, as that's not the date shown?

pjeweb: I did not find any way. Maybe we could manage our own sort value as a separate column. The only way to do this would probably be not using prisma (or using a raw query). They sadly do not support SQL syntax for the order by https://github.com/prisma/prisma/issues/5848

For now I would just try clicking "seen" in the right order...

MattIPv4: Ah, that's a shame -- I guess the other solution would be to make the column a boolean instead of a nullable timestamp?

pjeweb: that should work! i would not mind that

pjeweb: i would just add it as a second col though

MattIPv4: Works for me :)

apps/website/prisma/schema.prisma
```diff
@@ -3,7 +3,7 @@
 
 generator client {
   provider        = "prisma-client-js"
-  previewFeatures = ["fieldReference", "fullTextSearch"]
+  previewFeatures = ["fieldReference", "fullTextSearch", "orderByNulls"]
```
MattIPv4, line 6: Is this needed now?

pjeweb: yeah no, thanks for reviewing so diligently

MattIPv4 (approved):

MattIPv4:
I assume this will need a tiny data migration in prod to set the column for existing stuff?

pjeweb:
yeah I've done that already

pjeweb:
for the record:
- create branch from main on planetscale
- push new schema (with seenOnStream defaulting to false)
- create deploy request to main
- apply deploy request
- run `UPDATE ShowAndTellEntry SET seenOnStream = 1 WHERE seenOnStreamAt IS NOT NULL;`
- merge PR

PR #232, opened 2023-04-22
Use consistent date/time formatting

### LINKED ISSUE #198: Small styling inconsistencies

MattIPv4:
Resolves #198


apps/website/src/utils/datetime.ts
```diff
@@ -72,9 +44,29 @@ export function formatDateUTC(
...
+    ]
+      .filter(Boolean)
+      .join(":") + (showTimezone ? " UTC" : "")
+  );
+}
+
+export function formatDateTimeUTC(
+  dateTime: Date,
+  format: "short" | "long" = "short",
+  { showSeconds = false, showTimezone = false } = {}
```
MattIPv4, line 65: Should these be combined into a single settings object, or is it fine as-is?

pjeweb (commented):
I need to think about the timezones a bit. I feel like just using UTC might not be the best after all. I will review this tomorrow.

pjeweb:
My brain hurts from thinking about this, but I feel for the admin pages and for "my posts" I would prefer local timezones and for the show and tell index i think either CT or local time would be the best. I feel like UTC is too hard for people to understand. Assuming a user in CT submits a post at 9 PM, using UTC it would show as the next day in their my posts list. That feels wrong. Also I think it would be easier for the Alveus team to see CT times/dates in the admin area.

I would also use the users localized formatting for admin pages and "my posts" and only use the long US date with ordinals for the "public" pages.

MattIPv4:
> for the show and tell index i think either CT or local time would be the best. I feel like UTC is too hard for people to understand.

I think either way there is going to be confusion, as we only show a date. We don't show a timestamp, so we can't show the timezone its in, so either we do it in local so it always makes sense to the user, but then looks odd when Maya streams and the dates don't match everyone, or we use a fixed timezone and risk a mismatch with individual users.

> I would also use the users localized formatting for admin pages and "my posts" and only use the long US date with ordinals for the "public" pages.

That seems fine, as long as we force showing what timezone the timestamp is in, to reduce confusion about whether its UTC/CT/local

pjeweb:
> We don't show a timestamp, so we can't show the timezone its in

I would not mind putting a "CT" after the date to make it super clear. But I feel like the date on the show and tell index wont cause much confusion. It is not even clear if its the create/update/approve/publish date to the user, so it does not really matter that much if its off by one day. I only really care in their own submissions list and for the admins where you see the time as well.

> That seems fine, as long as we force showing what timezone the timestamp is in, to reduce confusion about whether its UTC/CT/local

that sound great!

MattIPv4:
@pjeweb What do you think of this?

My only pain is that we no longer have ordinals on the dates, but I don't think there's a great way to solve that unless we monkey-patch Intl formats

pjeweb:
> @pjeweb What do you think of this?
> 
> My only pain is that we no longer have ordinals on the dates, but I don't think there's a great way to solve that unless we monkey-patch Intl formats

I think something like this would be okay, and has decent browser support [1]:

```typescript
const getOrdinal = (n: number) => {
  const j = n % 10;
  const k = n % 100;
  if (j === 1 && k !== 11) return "st";
  if (j === 2 && k !== 12) return "nd";
  if (j === 3 && k !== 13) return "rd";
  return "th";
};

function formatDateWithDayOrdinal(date: Date, formatter: Intl.DateTimeFormat) {
  return formatter
    .formatToParts(date)
    .map(({ type, value }) => {
      switch (type) {
        case "day":
          return value + getOrdinal(parseInt(value, 10));
        default:
          return value;
      }
    })
    .join("");
}


const date = new Date(Date.UTC(2023, 3, 23, 0, 1, 0, 0));
const df = new Intl.DateTimeFormat(undefined, {day: "2-digit"});

console.log(formatDateWithDayOrdinal(date, df));
```

1) https://caniuse.com/mdn-javascript_builtins_intl_datetimeformat_formattoparts

MattIPv4:
Hm, let me play with that and Intl.PluralNumbers

MattIPv4:
Turns out Intl.PluralNumbers doesn't actually provide locale-specific translations, just mappings. I think it'll stick without ordinals so we don't mix `en` specific words into locale-formatted stuff.

pjeweb (approved):

PR #235, opened 2023-04-24
Fix scrolling in show and tell presentation view

### LINKED ISSUE #224: Disable scrolling in show and tell fullscreen
- Disable the ability to scroll when in fullscreen on show and tell
- Add new click regions at the top/bottom of the screen (where you can just see the next/previous posts), which change post like the existing up/down buttons
- Extra points: When the user tries to scroll up/down, make the click regions and/or the buttons highlighted as a visual prompt

MattIPv4:
Resolves #224

### DIFF
--- a/apps/website/src/pages/show-and-tell/index.tsx
+++ b/apps/website/src/pages/show-and-tell/index.tsx
@@ -180,7 +180,7 @@ const ShowAndTellIndexPage: NextPage<ShowAndTellPageProps> = ({
           className={
             "scrollbar-none flex flex-col transition-colors duration-200 " +
             (isPresentationView
-              ? "fixed inset-0 z-[100] snap-y snap-mandatory gap-5 overflow-y-auto overflow-x-hidden bg-black p-5"
+              ? "fixed inset-0 z-[100] snap-y snap-mandatory gap-5 overflow-hidden bg-black p-5"
               : "gap-20 bg-white/0")
           }
         >
@@ -215,7 +215,7 @@ const ShowAndTellIndexPage: NextPage<ShowAndTellPageProps> = ({
           )}
 
           {isPresentationView && (
-            <div className="fixed right-[20px] top-[20px] z-10 z-20 ml-auto flex w-[calc(20%-2em)] flex-col items-center gap-4 text-alveus-green">
+            <div className="fixed right-[20px] top-[20px] z-20 ml-auto flex w-[calc(20%-2em)] flex-col items-center gap-4 text-alveus-green">
               <div className="flex flex-row items-center gap-1">
                 <Heading level={1} className="pt-3">
                   Show and Tell
--- a/apps/website/src/pages/show-and-tell/index.tsx
+++ b/apps/website/src/pages/show-and-tell/index.tsx
@@ -1,4 +1,4 @@
-import React, { useCallback, useRef, useState } from "react";
+import React, { useCallback, useMemo, useRef, useState } from "react";
 import type { InferGetStaticPropsType, NextPage } from "next";
 import Image from "next/image";
 import Link from "next/link";
@@ -121,7 +121,11 @@ const ShowAndTellIndexPage: NextPage<ShowAndTellPageProps> = ({
     },
     [checkPosition]
   );
-  const registerObserveElement = useIntersectionObserver(onEntryIntersection);
+  const intersectionObserverOpts = useMemo(() => ({ threshold: 0.55 }), []);
+  const registerObserveElement = useIntersectionObserver(
+    onEntryIntersection,
+    intersectionObserverOpts
+  );
 
   // Scroll an element into the center of the viewport
   const scrollTo = useCallback((element: HTMLElement) => {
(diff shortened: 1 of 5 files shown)


pjeweb:
Works well, but I do think we would need to make the button more obvious. Either like you suggested when trying to scroll or just make them bigger / more obvious. You can click the next "slide"/post, but they are too small to click, I think.

MattIPv4:
I think I'm going to play around with listening to the wheel event and jumping to the next post if the user scrolls past a threshold -- I see that YT shorts just allow natively scrolling and it is _incredibly_ buggy for me, not an experience I want to replicate.

MattIPv4:
@pjeweb How does this feel to you to use?

MattIPv4:
> but with Chrome and Safari under macOS with a track pad it's scrolling very jerky and does often not scroll to the right end position.

Interesting -- there is no actual scrolling going on here at all, it just calls the scrollIntoView on the element once the scroll event reaches a threshold, so I'm not sure where that jerky motion or misalignment would be coming from

pjeweb:
https://github.com/alveusgg/alveusgg/pull/236

Can you compare how this feels for you? I think the biggest mistake was using snap-mandatory which forces the slide to change even with tiny amounts of wheel scroll distance. This feels decently good in FF and Chrome on Mac. I'm booting windows right now.

MattIPv4:
👀 So in #236, I never get snapped in Firefox, it just scrolls as normal always. No matter how far into a post I scroll, it doesn't snap the alignment ever. (Also, removing the padding/margin has caused the shadows to clip)

pjeweb:
Yeah the shadow clipping should be fixed in none-presentation mode for sure. Was just a quick fix to prevent each slide from having overflow scrolling in the presentation mode.

It does snap for me in FF both on Windows and Mac, but only very very gently if you are within like 100px of the snap point and only if the scroll is decently long and in the right direction. But yeah basically no snapping.

I think that would be the save version so it works everywhere without jenky behavior, but less snappy :)

MattIPv4:
Let me see if I can get snap debouncing working correctly this evening so we get the best of both worlds (free-flowing scroll, and then snapping properly when you stop)

pjeweb:
> Let me see if I can get snap debouncing working correctly this evening so we get the best of both worlds (free-flowing scroll, and then snapping properly when you stop)

Yes please. Let me know if I can help or test anything. It might be worth adding a polyfill for windows/chrome: https://github.com/iamdustan/smoothscroll

MattIPv4:
Does this feel better to you? And did you want me to include the margin/padding removal here, or do you want to do that in a separate PR (ideally that also keeps a fix in place for shadows in the regular view). I started looking at it, but my mind is kinda melting trying to parse how the presentation view actually is laid out

pjeweb:
> Does this feel better to you? 

Hard to tell.

Chrome/Windows is quite bad still because it hard jumps to the snap point (no smoothing). And thats what maya is using. Might be fixed by just throwing in the polyfill.

Otherwise it's not really that much better or worse in Chrome then the live version. But I am biased. Sometimes it's hard to get over the initial scroll threshold and I get snapped back to the old one, sometimes it scrolls to the next/prev with a small scroll distance. Browser scrolling is _hard_. Firefox is behaving well with pretty much all your versions though.

Maybe we need to ask different people for feedback which version they prefer? I really cant tell anymore.

> And did you want me to include the margin/padding removal here, or do you want to do that in a separate PR (ideally that also keeps a fix in place for shadows in the regular view). I started looking at it, but my mind is kinda melting trying to parse how the presentation view actually is laid out

I can tacke that in a separate PR. Yeah its kind of complicated trying to keep the layout. What works for me is just lowering the padding a bit in presentation mode. I cant see the shadow clipping with that:
```typescript
      // Negative margin to allow padding for shadows
      wrapperClassName={`
          flex-1
          ${isPresentationView ? "pb-6 -mb-4" : "pb-16 -mb-12"}
        `}
```

MattIPv4:
> Chrome/Windows is quite bad still because it hard jumps to the snap point (no smoothing). And thats what maya is using. Might be fixed by just throwing in the polyfill.

I tried the polyfill, but that didn't work -- I believe this is re-enabling snapping colliding with the scrollIntoView. Really, I need a way to wait for the scroll to finish, then re-enable snapping. Will look at introducing a dependency to manage that

apps/website/src/pages/show-and-tell/index.tsx
```diff
@@ -122,87 +123,95 @@ const ShowAndTellIndexPage: NextPage<ShowAndTellPageProps> = ({
...
-    start: HTMLElement;
-    distance: number;
+  // Scroll an element into the center of the viewport
+  const scrollTo = useCallback((element: HTMLElement) => {
+    // We use a library to do this, because not every browser supports it
+    // And even in Chrome, smooth breaks for the first/last element of things
+    scrollIntoView(element, {
+      behavior: "smooth",
+      block: "center",
+    });
```
MattIPv4, line 133: My only gripe with this now is that this animation-based scrolling has no way to abort if the user starts scrolling again, so you somewhat end up fighting it briefly. I'm not sure that's really that much of a problem though

pjeweb (approved):

PR #250, opened 2023-04-29
Add more clips to some ambassadors, tweak highlights section

pjeweb:
- add a `classes` util for conditional classes (maybe we can use https://www.npmjs.com/package/classnames instead?)
- update data with more clips
- make ambassador highlights section slightly darker to offset it from the socials footer
- make highlights section linkable (`#highlights`)

### DIFF
--- a/apps/website/src/pages/ambassadors/[ambassadorName].tsx
+++ b/apps/website/src/pages/ambassadors/[ambassadorName].tsx
@@ -311,7 +311,7 @@ const AmbassadorPage: NextPage<AmbassadorPageProps> = ({
       </div>
 
       {ambassador.clips.length > 0 && (
-        <Section dark>
+        <Section dark className="bg-alveus-green-800">
           <Heading level={2} className="text-center text-4xl">
             {ambassador.name}
             &apos;s Highlights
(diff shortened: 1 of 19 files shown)


apps/website/src/pages/ambassadors/[ambassadorName].tsx
```diff
@@ -311,8 +311,8 @@ const AmbassadorPage: NextPage<AmbassadorPageProps> = ({
       </div>
 
       {ambassador.clips.length > 0 && (
-        <Section dark>
-          <Heading level={2} className="text-center text-4xl">
+        <Section dark className="bg-alveus-green-800">
+          <Heading id="highlights" level={2} className="text-center text-4xl">
```
Comment (MattIPv4) [inline apps/website/src/pages/ambassadors/[ambassadorName].tsx:315]: Is it worth also setting the `link` prop here, so folks can click the heading to get the anchor?

MattIPv4:
(and ty for the classes util, i've written that filter logic so many times and yet never thought to move it into one 🤦  -- I'm not too worried about moving to a dep for it, unless you are?)

pjeweb:
> (and ty for the classes util, i've written that filter logic so many times and yet never thought to move it into one 🤦 -- I'm not too worried about moving to a dep for it, unless you are?)

(nah i'm fine with the util for now)

apps/website/src/pages/ambassadors/[ambassadorName].tsx
Comment (pjeweb) [reply apps/website/src/pages/ambassadors/[ambassadorName].tsx]: sure

MattIPv4 (approved):

PR #256, opened 2023-05-02
Remove duplicate video loader calls for server build

MattIPv4:
We already didn't actually emit the generated files during the server build, so why bother even running ffmpeg to generate them.

The only change to make this work was to switch the asset hash used in the files to be that of the input video, rather than the ffmpeg output. This should be fine, as the output should be deterministic based on the input (per Webpack loader requirements, and per the actual logic in the loader).

### DIFF
--- a/apps/website/build-scripts/video-loader.cjs
+++ b/apps/website/build-scripts/video-loader.cjs
@@ -274,11 +274,15 @@ const videoLoader = async (context, content) => {
 };
 
 module.exports = function (content) {
-  console.log(`Processing video ${this.resourcePath}...`);
+  console.log(`Processing video ${this.resourcePath} ...`);
   const callback = this.async();
   videoLoader(this, content)
     .then((res) => {
-      console.log(`Processed video ${this.resourcePath}: ${res}`);
+      console.log(
+        ` ... ${this.resourcePath} completed${
+          this.getOptions().isServer ? " (render skipped)" : ""
+        }`
+      );
       callback(null, res);
     })
     .catch(callback);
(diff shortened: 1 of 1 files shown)


apps/website/build-scripts/video-loader.cjs
```diff
@@ -149,18 +157,21 @@ const videoResized = async (
  *
  * @param {Object} context Webpack compilation context
  * @param {string} format Output filename format
+ * @param {string|Buffer} content Input video content
+ * @param {boolean} isServer Is the loader being run for the server build
  * @param {number} [width] Optional width for the filter
  * @param {number} [height] Optional height for the filter
- * @return {Promise<{ name: string, content: Buffer }>}
+ * @return {Promise<{ name: string, content: string|Buffer }>}
```
MattIPv4, line 164:
```suggestion
 * @return {Promise<{ name: string, content: Buffer|null }>}
```

apps/website/build-scripts/video-loader.cjs
```diff
@@ -262,9 +274,13 @@ const videoLoader = async (context, content) => {
 };
 
 module.exports = function (content) {
+  console.log(`Processing video ${this.resourcePath}...`);
   const callback = this.async();
   videoLoader(this, content)
-    .then((res) => callback(null, res))
+    .then((res) => {
+      console.log(`Processed video ${this.resourcePath}: ${res}`);
```
MattIPv4, line 281: @pjeweb how do you feel about this logging? Is logging the full output too noisy? Is any level of logging too noisy?

pjeweb: I dont mind it. We can remove it later if it bothers us. (or maybe add a environment variable to allow building w or w/o verbose build logs?)

pjeweb (approved):

PR #258, opened 2023-05-02
Set slideshow quality based on screen size

MattIPv4:
Default to a quality of 50 for mobile devices to reduce payload size, and then increase the quality as the device screen gets larger.

Use 90 rather than 100 so that the file still gets processed (100 appears to skip processing, there is a large file size jump from 99 to 100).

### DIFF
--- a/apps/website/src/components/content/Slideshow.tsx
+++ b/apps/website/src/components/content/Slideshow.tsx
@@ -1,7 +1,9 @@
 import type { ImageProps } from "next/image";
 import Image from "next/image";
-import React, { type CSSProperties, useId, useMemo } from "react";
-import { camelToKebab } from "../../utils/string-case";
+import React, { type CSSProperties, useCallback, useId, useMemo } from "react";
+
+import { camelToKebab } from "@/utils/string-case";
+import { createImageUrl, type ImageLoaderProps } from "@/utils/image";
 
 type SlideshowProps = {
   images: {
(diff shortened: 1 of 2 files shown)


MattIPv4:
https://pagespeed.web.dev/analysis/https-alveusgg-website-btk83lju0-alveus-vercel-app/dpjmydv3u8?form_factor=desktop is even happier than my last PR

pjeweb (approved):
Just to be clear, this means tablet and up will fetch the 50 quality first and then 75 or 90 quality, possibly canceling the first requests. So more requests and more bytes for bigger viewports in total. But if it still performs better, it might be worth.

MattIPv4:
Yeah, on desktop you'll load 50 and then the 75 -- I was hoping Next would be lazy enough that it'd only load the 50 for the first, but it loads it for all. Thoughts on forcing the others to wait until we've determined the quality?

pjeweb:
You can test that, for sure. I have a different idea how to make the browser decide on its own:

```typescript
import { createImageUrl } from "@/utils/image";

<Image
  …
  quality="50"
  loader={(props) => {
    let quality = 50;
    if (props.width >= 768) quality = 75;
    else if (props.width >= 1280) quality = 90;

    return createImageUrl({ ...props, quality });
  }}
  …
/>
```

MattIPv4:
Ohhh, that's smart af. Completely forgot you could override the loader... let me try it.

pjeweb (approved):
seems to work well now, I'll let you check lighthouse

MattIPv4:
Pagespeed seems to be pretty inconsistent, but overall it seems like this is doing okay: https://pagespeed.web.dev/analysis/https-alveusgg-website-635btqdvs-alveus-vercel-app/1nw3ues5tw?form_factor=mobile it's complaining far less about image sizes than prod

PR #264, opened 2023-05-09
feat(stats): set up simple view stats using self-hosted plausible

pjeweb:


MattIPv4:
👀 Where are we at with analytics? Are we happy with what Vercel provides, or did we want to move forward with Plausible still?

pjeweb:
Still would like to try plausible in addition. I was hoping to get the chatbot running sooner, but if you are up for it, i'll get this PR ready first. I guess we need alveus approval again?!

MattIPv4:
I think the chatbot is probably priority still? And yah, if Alveus hasn't +1'ed this policy change yet, we'll want someone to do that before this goes live.

MattIPv4:
👀 Checking in on what the plan is here, now that we've got the notification stuff mostly sorted? Are we still wanting more detailed analytics, or is what Vercel gives us enough at this point? I've mostly just been looking at the traffic over time in there for specific pages, and what geos are hitting us.

PR #268, opened 2023-05-12
Investigating Webpack cache size

MattIPv4:

PR #269, opened 2023-05-12
Implement custom video loader caching, disable Webpack caching

MattIPv4:
A solution to the continued Webpack cache issues investigated in #268

https://github.com/alveusgg/alveusgg/pull/269/commits/0d3ab4ab5ad53a6465ec74d6cc16e0389658bba6 (no existing cache): https://vercel.com/alveus/alveusgg-website/BAijGm2Wv9N9Taw64HCvz3SG1soE 366.29 MB cache in ~7m

https://github.com/alveusgg/alveusgg/pull/269/commits/37bde3a5843e31e561c18d9f9f756c7eb68f5e45 (cache from above): https://vercel.com/alveus/alveusgg-website/3WDhweMUvEYMhoE7iuTqqrTppjtB 366.29 MB cache in ~3m


MattIPv4:
@pjeweb what do you think of this as a solution?

apps/website/build-scripts/video-loader.cjs
```diff
@@ -266,26 +343,77 @@ const videoLoader = async (context, content) => {
...
+  };
 };
 
+/**
+ * Webpack hook to clean up old cache files
+ *
+ * @return {Promise<void>}
+ */
+const cacheCleanup = async () => {
+  // Get all files in the cache dir
```
pjeweb, line 374: this should handle the case if the cacheDir does not exist yet. that might happen when doing a clean `pnpm build`.
Something like `if (!fs.existsSync(cacheDir)) return;`

pjeweb (dismissed):
The solution makes sense to me. I would not want to debug webpack any further. So unless you want to, I am happy with this.

apps/website/build-scripts/video-loader.cjs
MattIPv4: 🤔 It should never be possible for this directory to not exist -- the loader always runs before this, and always writes to the cache dir?

MattIPv4: I guess I should handle the case that something manages to remove that directory between the loader running and the cleanup plugin running 👍

pjeweb: It happened to me. I 'rm -rf'ed .next and ran 'pnpm build'

pjeweb (approved):

apps/website/build-scripts/video-loader.cjs
MattIPv4: Oh right yes, the server build happens first that doesn't emit assets 🤦

PR #271, opened 2023-05-19
Set up Vercel Web Analytics

### LINKED ISSUE #266: Implement privacy-friendly analytics/view statistics
# Proposal

My proposal is to set up our own instance of Plausible (https://github.com/plausible/analytics). It is available as OSS for self-hosting, designed to be privacy friendly. It does not use cookies and does not track beyond a very basic anonymized hash that is used to count unique visitors per day.  https://plausible.io/privacy-focused-web-analytics

# Frontend

The tracking consists of a small ~1kb script that can be cached and an event endpoint. Both can be proxied through the edge network. (see https://github.com/alveusgg/alveusgg/pull/264/commits/e2b7a5eb61d77e4446224dcce55d004e2d309390) 

# Backend

We can either use the plausible UI directly with user/password logins or we can embed the dashboard for one site in the admin panel (see https://github.com/alveusgg/alveusgg/pull/264/commits/53afef4dfd719ae5f7f92e145bfa6f42714b8aa9). The authentication is through a secret get parameter that is provided as a server environment variable and passed to the admin page, if the user has appropriate permissions and is authenticated through the site auth (currently using Twitch OAuth). This also means that the secret can be extracted by a malicious or compromised user with permissions to view the stats. If that comes to be known, we would need to revoke the share link in the plausible UI and generate a new secret. Considering there are no personal or PII data accessible through the plausible embed, that risk seems tolerable.

#  Hosting

Right now I have set up a docker compose deployment on a Hetzner Cloud instance, hosting plausible and its dependencies (postgres and clickhouse) as suggested at https://plausible.io/docs/self-hosting. Additionally I have set up nginx, certbot and watchtower in docker compose for https, caching and automatic updates: https://github.com/alveusgg/plausible-hosting

I am not sure how much traffic we should expect and how big to scale the server. The Hetzner cloud instance (Ampere arm64, 8 core, 16 gig, 160 gb nvme ssd, Ubuntu 22.04) can be scaled to 16 cores/32 gig. If that is not enough we would have to move hosting/servers or split the system into multiple machines. Firewall, snapshots and backups are set up in hetzner cloud.

These instances are only available in their german data center, which is good for GDPR, but not the best for access times from the US. It is however a very small async payload that should not block any user interaction, so i do not think that really matters.

I am open for any other suggestions for hosting.

We can also buy this service or any other analytics SaaS, but that comes with its own cost and privacy concerns.

# Privacy Policy / Consent

Imho we should not need explicit consent as we do not collect any PII and the raw visit stats should fall under "legitimate interest". We should however mention it in the privacy policy.  (see https://github.com/alveusgg/alveusgg/pull/264/commits/eaf8b6da8790841a31a528d7652ebad578348dcf)

# Future ideas

Some of the next things would be to look into tracking redirects to external sites without a page visit, possibly with an edge middleware using the plausible API. https://plausible.io/docs/events-api

-------------------------

PR #264

pjeweb:
Use vercel web analytics

Resolves #266

### DIFF
--- a/apps/website/src/data/privacy-policy.md
+++ b/apps/website/src/data/privacy-policy.md
@@ -151,7 +151,7 @@ Personal Data is collected for the following purposes and using the following se
 - ### Analytics
 
   The services contained in this section enable the Owner to monitor and analyze web traffic and can be used to keep track of User behavior.
-  
+
   #### Vercel Web Analytics
 
   This website utilizes Vercel Web Analytics to measure views. Vercel Web Analytics is designed to align with
@@ -389,4 +389,4 @@ This privacy statement has been prepared based on provisions of multiple legisla
 
 This privacy policy relates solely to this Website, if not stated otherwise within this document.
 
-Latest update: April 23rd, 2023
+Latest update: May 19th, 2023
--- a/apps/website/src/data/privacy-policy.md
+++ b/apps/website/src/data/privacy-policy.md
@@ -148,6 +148,31 @@ Personal Data is collected for the following purposes and using the following se
   This processing constitutes a sale based on the definition under the CCPA. In addition to the information in this clause, the User can find information regarding how to opt out of the sale in the section detailing the rights of Californian consumers
 
 
+- ### Measuring views (Vercel Web Analytics)
+
+  This website utilizes Vercel Web Analytics to measure views. Vercel Web Analytics is designed to align with
+  international data protection guidelines. It does not collect personal identifiers that track and cross-check
+  users' data across different applications or websites. The service deals with only aggregated data which does not
+  allow identification or re-identification of end users.
+  Data points like page views and custom events are recorded anonymously. They are not tied or associated with any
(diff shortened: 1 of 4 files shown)


MattIPv4 (changes requested):
policy date needs updating please

apps/website/src/data/privacy-policy.md
```diff
@@ -148,6 +148,31 @@ Personal Data is collected for the following purposes and using the following se
   This processing constitutes a sale based on the definition under the CCPA. In addition to the information in this clause, the User can find information regarding how to opt out of the sale in the section detailing the rights of Californian consumers
 
 
+- ### Measuring views (Vercel Web Analytics)
```
MattIPv4, line 151:
```suggestion
- ### Analytics

  The services contained in this section enable the Owner to monitor and analyze web traffic and can be used to keep track of User behavior.
  
  #### Vercel Web Analytics
```
To be consistent with other sections her

MattIPv4 (approved):

PR #273, opened 2023-05-20
Add pages for each animal quest episode

### LINKED ISSUE #182: Animal quest episode pages
It would be great to have per-episode pages available on the website that have far more information about the episode within them. Perhaps a full transcript, or some notes, and a embed/link for the presentation used for the episode.

We should check with Maya and the Alveus team before implementing to make sure they're on-board with publishing content like that, and ideally to get access to the presentations for ease of creating the notes.

MattIPv4:
Resolves #182 

cc https://github.com/alveusgg/data/pull/16


MattIPv4:
TODO: We should probably have some route redirects, maybe for /animal-quest/{episodeNumber} and /animal-quest/episode-{episodeNumber} -- anything else you think would be worth handling @pjeweb?

MattIPv4:
TODO: Meet the ambassadors slice on each page, using the same ambassador carousel component as the homepage

pjeweb (approved):

PR #278, opened 2023-05-27
Consent: add max width

pjeweb:
Limit the width of the consent box for readability

### DIFF
--- a/apps/website/src/components/Consent.tsx
+++ b/apps/website/src/components/Consent.tsx
@@ -52,7 +52,7 @@ const Consent: React.FC<ConsentProps> = ({
           <>
             <p className="m-6 mb-3 text-2xl">Loading {item}...</p>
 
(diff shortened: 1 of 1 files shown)


MattIPv4:
Hm, this feels very constrained on the homepage now, with no real visual reason for it being that constrained:

image
image

What's the reason for doing this, is the box becoming too wide on very large monitors?

I think `max-w-2xl` would constrain it in a way that matches what I think looks to be more comfortable visually?

image

pjeweb:
> What's the reason for doing this, is the box becoming too wide on very large monitors?

Yeah the text lines get a bit too long to read comfortable and it does not look that coherent/balanced to me.

> I think `max-w-2xl` would constrain it in a way that matches what I think looks to be more comfortable visually?

yeah that looks good. I'll push an update later

MattIPv4 (approved):

PR #280, opened 2023-05-31
T3 env

tomheaton:
Replace old env schema validation with T3 Env (https://env.t3.gg/).

Could merge client and server schemas into one schema but its shipped to the client so variable names are visible (shouldn't matter as site is open source but I left it as is).


pjeweb:
> Could merge client and server schemas into one schema but its shipped to the client so variable names are visible (shouldn't matter as site is open source but I left it as is).

Oh variable **names**! Yeah that should not matter.

tomheaton:
I'll merge the schemas!

apps/website/sw.tsup.config.ts
```diff
@@ -20,7 +20,10 @@ export default defineConfig((options) => ({
     options.define = {
       ...Object.fromEntries(
         Object.entries(env)
-          .filter(([, value]) => value !== undefined)
+          .filter(
+            ([, value]) =>
+              value !== undefined && value.startsWith("NEXT_PUBLIC_")
```
tomheaton, line 25: added this here because this was originally client only env, wasnt entirely sure how this code runs so added in additional check to make sure client only env is accessed

pjeweb:
@tomheaton Looking good. Would you mind rebasing or merging main?

Unless @MattIPv4 has any concern, I would merge this.

MattIPv4:
The diff here seems fine to me

pjeweb:
Hi @tomheaton, wanted to see how things are going. Do you need any support from us to finish this up? We merged a big PR for the notification system, so if you need me to help merge those changes, let me know.

tomheaton:
I'll get right on it!

tomheaton:
think thats it 🚀

pjeweb (changes requested):
the old `schema.mjs` is left over, otherwise this looks great

pjeweb:
thanks again for you contribution, @tomheaton

PR #285, opened 2023-06-03
Fix issue link in README

joar:
Seems like the `#9` shorthand doesn't work in README

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,7 +41,7 @@ Hosting (production):
 
 ## Systems overview
 
-For a more complete overview see: #9
+For a more complete overview see: https://github.com/alveusgg/alveusgg/issues/9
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -41,7 +41,7 @@ Hosting (production):
 
 ## Systems overview
 
-For a more complete overview see: #9
+For a more complete overview see: https://github.com/alveusgg/alveusgg/issues/9
```
MattIPv4, line 44:
```suggestion
For a more complete overview see: [#9](https://github.com/alveusgg/alveusgg/issues/9)
```
For cleanliness, I think it'd be better to keep the text as `#9` still?

joar: It renders as `#9` on github, but it wouldn't hurt to be portable.

MattIPv4 (approved):

PR #286, opened 2023-06-03
Update node/pnpm versions in README.md

pjeweb:


MattIPv4:
Is it worth just updating the text in the readme to say check the package.json engines? So we don't have the info duplicated (so we don't forget to update it again lol)

README.md
```diff
@@ -14,8 +14,8 @@ This project uses pnpm workspaces. The main app is the website package (`apps/we
 
 For development:
 
-- node v16
-- pnpm 7 with workspaces
+- node v18
+- pnpm 8 with workspaces
```
MattIPv4, line 18:
```suggestion
- Node.js
- pnpm with workspaces
```

README.md
```diff
@@ -61,7 +61,7 @@ TODO
 
 ### Local development
 
-1. Install Node.js v16 and pnpm
+1. Install Node.js and pnpm
```
MattIPv4, line 64:
```suggestion
1. Install Node.js and pnpm (see `engines` in `package.json`)
```

pjeweb:
> Is it worth just updating the text in the readme to say check the package.json engines? So we don't have the info duplicated (so we don't forget to update it again lol)

yeah that makes sense. i also tried to spell all the names properly

MattIPv4 (approved):

PR #288, opened 2023-06-05
Add option to sort ambassadors by classification

mforsb:
I felt this was general enough to replace the 'all' option, so I removed that (which also meant the code could be simplified since it was special-cased). If anyone disagrees with this I'd be more than happy to undelete it.

Suggestions are welcome on the naming. I also considered 'type' + a few others, but nothing felt obviously better.

Also made some style changes:
- left align everything so ambassadors are centered rather than pushed slightly to the right
- make grid denser (is that the right word?, more columns) on large displays
- make ambassador names smaller than the group heading

Depends on https://github.com/alveusgg/data/pull/19

### DIFF
--- a/apps/website/src/pages/ambassadors/[ambassadorName].tsx
+++ b/apps/website/src/pages/ambassadors/[ambassadorName].tsx
@@ -20,6 +20,7 @@ import {
 import enclosures, { type Enclosure } from "@alveusgg/data/src/enclosures";
 import { getIUCNStatus } from "@alveusgg/data/src/iucn";
 
+import { getClassification } from "@alveusgg/data/src/ambassadors/classification";
 import animalQuestImage from "@/assets/animal-quest/full.png";
 
 import Section from "@/components/content/Section";
@@ -37,6 +38,7 @@ import {
 } from "@/utils/string-case";
 import { getDefaultPhotoswipeLightboxOptions } from "@/utils/photoswipe";
 import { typeSafeObjectKeys } from "@/utils/helpers";
+import { convertToSlug } from "@/utils/slugs";
 
 const parseDate = (date: string | null): string => {
   if (!date) return "Unknown";
@@ -197,7 +199,15 @@ const AmbassadorPage: NextPage<AmbassadorPageProps> = ({
                 <div className="ml-4">
                   <p className="text-xl">{ambassador.species}</p>
                   <p className="text-xl italic text-alveus-green-700">
-                    {ambassador.scientific}
+                    {ambassador.scientific} (
+                    <Link
+                      href={`/ambassadors#classification:${convertToSlug(
+                        getClassification(ambassador.class)
+                      )}`}
+                    >
+                      {getClassification(ambassador.class)}
+                    </Link>
+                    )
                   </p>
                 </div>
               </div>
--- a/apps/website/src/pages/ambassadors/[ambassadorName].tsx
+++ b/apps/website/src/pages/ambassadors/[ambassadorName].tsx
@@ -19,8 +19,8 @@ import {
 } from "@alveusgg/data/src/animal-quest";
 import enclosures, { type Enclosure } from "@alveusgg/data/src/enclosures";
 import { getIUCNStatus } from "@alveusgg/data/src/iucn";
-
 import { getClassification } from "@alveusgg/data/src/ambassadors/classification";
+
 import animalQuestImage from "@/assets/animal-quest/full.png";
 
 import Section from "@/components/content/Section";
--- a/apps/website/src/pages/ambassadors/index.tsx
+++ b/apps/website/src/pages/ambassadors/index.tsx
@@ -21,6 +21,7 @@ import { camelToKebab } from "@/utils/string-case";
 import { typeSafeObjectEntries } from "@/utils/helpers";
 import { classes } from "@/utils/classes";
 import { parseAmbassadorDate, sortAmbassadorDate } from "@/utils/datetime";
+import { convertToSlug } from "@/utils/slugs";
 
 import Section from "@/components/content/Section";
 import Heading from "@/components/content/Heading";
@@ -31,7 +32,6 @@ import leafRightImage1 from "@/assets/floral/leaf-right-1.png";
 import leafLeftImage1 from "@/assets/floral/leaf-left-1.png";
 import leafRightImage2 from "@/assets/floral/leaf-right-2.png";
 import leafLeftImage2 from "@/assets/floral/leaf-left-2.png";
-import { convertToSlug } from "@/utils/slugs";
 
 // We don't want to show retired ambassadors on the page
 const activeAmbassadors = typeSafeObjectEntries(ambassadors).filter(
@@ -75,9 +75,9 @@ const sortByOptions = {
   enclosures: {
     label: "Enclosures",
     result: activeAmbassadors.reduce<AmbassadorsByGroup>((map, [key, val]) => {
-      const group = val.enclosure;
+      const group = camelToKebab(val.enclosure);
       map.set(group, {
-        name: enclosures[group].name,
+        name: enclosures[val.enclosure].name,
         items: [...(map.get(group)?.items || []), key],
       });
       return map;
@@ -181,12 +181,7 @@ const AmbassadorGroup: React.FC<{
 
   return (
     <div ref={scroll}>
-      <Heading
-        level={2}
-        className="mb-8"
-        id={`${type}:${camelToKebab(group)}`}
-        link
-      >
+      <Heading level={2} className="mb-8" id={`${type}:${group}`} link>
         {name}
       </Heading>
       <AmbassadorItems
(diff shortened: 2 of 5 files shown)


MattIPv4:
I'll take a full look at this tomorrow, but I would like to keep the default sorting as All please -- it follows the specific order set in the data repo :)

apps/website/src/pages/ambassadors/index.tsx
```diff
@@ -155,24 +163,30 @@ const AmbassadorGroup: React.FC<{
...
+        {ambassadors.map((key) => (
+          <AmbassadorItem key={key} ambassador={key} level={3} />
+        ))}
+      </div>
     </div>
   );
 };
 
-const AmbassadorsPage: NextPage = () => {
+export const getStaticProps: GetStaticProps = async () => {
```
MattIPv4, line 181: Is there a reason to start using static props here? Seems like we're just going to be generating extra data in the initial payload as this will be in the JS data as well as the JSON data now?

MattIPv4:
👀 You might want to include a temporary commit here that switches your data dependency over to your PR branch, so that this can be deployed to a preview environment: `pnpm remove @alveusgg/data && pnpm add github:mforsb/alveus-data\#class`

apps/website/src/pages/ambassadors/index.tsx
```diff
@@ -34,29 +33,52 @@ const activeAmbassadors = typeSafeObjectEntries(ambassadors).filter(
...
+    case Class.Mammalia: return "Mammals";
+    case Class.Aves: return "Birds";
+    case Class.Reptilia:
+    case Class.Amphibia: return "Reptiles & Amphibians";
+    case Class.Arachnida:
+    case Class.Diplopoda:
+    case Class.Insecta:
+    case Class.Malacostraca: return "Invertebrates";
+  }
+}
```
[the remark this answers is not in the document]
mforsb: Sure! I didn't realize the data repo was in charge of presentation, should have double checked. Would you prefer to keep `Class` and `Classification` separate or simplify it (but make it less scientifically accurate I guess) to just `Classification`?

apps/website/src/pages/ambassadors/index.tsx
```diff
@@ -155,24 +163,30 @@ const AmbassadorGroup: React.FC<{
...
         link
       >
         {name}
       </Heading>
-      <AmbassadorItems
-        ambassadors={ambassadors}
-        className="md:ml-8"
-        level={3}
-      />
+      <div className="grid gap-8 sm:grid-cols-2 md:grid-cols-3 lg:grid-cols-4 xl:grid-cols-5 2xl:grid-cols-6">
```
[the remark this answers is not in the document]
mforsb: Done! I made it 5 per row, the difference between 4 and 6 felt quite jarring when switching.

apps/website/src/pages/ambassadors/index.tsx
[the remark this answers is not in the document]
mforsb: Done

MattIPv4:
> does vercel pick up on overrides?

Let's find out! If not, you can just replace the actual dependency with `cd apps/website && pnpm remove @alveusgg/data && pnpm add github:mforsb/alveus-data\#class`

MattIPv4:
`ERROR  Unexpected end of JSON input while parsing near "...ata.git#class\"\n  }\n}\n" in /vercel/path0/package.json`

mforsb:
There, things should work now.

> eyes It looks like the deep links have broken -- for example, `/ambassadors#enclosures:reptiles` no longer works

This is also broken on main, at least on my machine.

MattIPv4:
> This is also broken on main, at least on my machine.

Hm, interesting. It should switch to the correct sort by option, and scroll down to the relevant heading. It looks like your PR has broken that because the heading IDs have changed, you've now got spaces etc. in the headings.

MattIPv4:
Is it worth also including the classification on the individual ambassador page, and perhaps even linking it back like we do for enclosure? Maybe put it in brackets after the species scientific name?

mforsb:
> > This is also broken on main, at least on my machine.
> 
> Hm, interesting. It should switch to the correct sort by option, and scroll down to the relevant heading. It looks like your PR has broken that because the heading IDs have changed, you've now got spaces etc. in the IDs.

True that my IDs were wrong, but linking was broken anyways (didn't change `in` to `.has()` when switching `AmbassadorsByGroup` from `object` to `Map`) in an earlier PR.

> Is it worth also including the classification on the individual ambassador page, and perhaps even linking it back like we do for enclosure? Maybe put it in brackets after the species scientific name?

Which do you prefer?
Reed Ambassadors Alveus Sanctuary
image

MattIPv4:
> True that my IDs were wrong, but linking was broken anyways (didn't change `in` to `.has()` when switching `AmbassadorsByGroup` from `object` to `Map`) in an earlier PR.

Oh, good catch -- ty for fixing!

> Which do you prefer?

I think probably the `(Mammals)` -- thoughts on putting it in italics on the same link as the scientific name, to keep it a bit more compact?

mforsb:
> thoughts on putting it in italics on the same link as the scientific name, to keep it a bit more compact?

I think it looks good! Should I include the parenthesis in the link?
image

MattIPv4:
> Should I include the parenthesis in the link?

Probably not, just the text inside? That feels proper to me but I have no idea why.

apps/website/src/pages/ambassadors/[ambassadorName].tsx
```diff
@@ -20,6 +20,7 @@ import {
 import enclosures, { type Enclosure } from "@alveusgg/data/src/enclosures";
 import { getIUCNStatus } from "@alveusgg/data/src/iucn";
 
+import { getClassification } from "@alveusgg/data/src/ambassadors/classification";
```
Comment (MattIPv4) [inline apps/website/src/pages/ambassadors/[ambassadorName].tsx:23]:
```suggestion
import { getClassification } from "@alveusgg/data/src/ambassadors/classification";

Nitpick, import grouping

File: apps/website/src/pages/ambassadors/index.tsx [hunk 5]
```diff
@@ -27,6 +31,7 @@ import leafRightImage1 from "@/assets/floral/leaf-right-1.png";
 import leafLeftImage1 from "@/assets/floral/leaf-left-1.png";
 import leafRightImage2 from "@/assets/floral/leaf-right-2.png";
 import leafLeftImage2 from "@/assets/floral/leaf-left-2.png";
+import { convertToSlug } from "@/utils/slugs";
```
MattIPv4, line 34: Nitpick, import grouping -- can this be moved up to the other utils, please?

apps/website/src/pages/ambassadors/index.tsx
```diff
@@ -155,15 +183,15 @@ const AmbassadorGroup: React.FC<{
     <div ref={scroll}>
       <Heading
         level={2}
-        className="mb-4"
+        className="mb-8"
         id={`${type}:${camelToKebab(group)}`}
```
MattIPv4, line 187: Given we're now generating the slugs in `sortByOptions`, is this `camelToKebab` needed -- or rather, should this be in the `sortByOptions` logic for the existing options in there (given that the equivalent `kebabToCamel` was removed below)?

MattIPv4:
This is super close, thank you for all the work on this 💚 just a couple of nitpicks and I think we should be good to go

apps/website/src/pages/ambassadors/index.tsx
mforsb: Sure! That would probably be better. All group keys except for those in classification are already a single lower-case word I think, should I slug them as anyways to future-proof?

MattIPv4: Yeah, I think having it in there is probably best to future-proof (I assume past me added it in here for that reason)

mforsb: Also should it be based on the key or the actual name of the enclosure, i.e. should it be `#enclosures:critters` or `enclosures:critter-cave`?

MattIPv4: I think the key, so it remains backwards compatible

mforsb:
I think that was everything?

mforsb:
> I did notice that your lock file seems to have a massive diff, which seems odd? It looks to be changing all the quotes?

Because this is the first time the file has changed since #284, no?

MattIPv4:
Oh yes, that makes sense, didn't realise I hadn't caught that in that PR -- would you mind reverting that file to match `main`, and then committing with `--no-verify` to avoid prettier doing that for now? I'll create a follow-up PR to drop the lockfile into an ignore.

MattIPv4 (approved):
LGTM, thank you for all the work on this 💚

_If you've got any feedback for the contributing experience, please feel free to drop it in a comment here, or an issue, or a PR to improve the docs 😉_

mforsb:
It's been great! :) There are a couple doc things I would like to change. Can probably do that later this week.

PR #295, opened 2023-06-13
Tweak consent button styling

tomheaton:
Restyle the consent button (the one to open the consent dialog) to have an "enabled" background color.

**Before**

https://github.com/alveusgg/alveusgg/assets/50220137/c33eef95-848b-4e41-814c-3acb3b90a7cd

**After**

https://github.com/alveusgg/alveusgg/assets/50220137/5d22dc14-46f8-4eef-ac7f-8c9243252667

**Note:** I also tried using the exact same styling as the actual toggle switches, so if you'd like that more I can do that instead.


MattIPv4:
Hm -- the issue here is that there is now not a hover/focus state when the toggle is on, which is why I had avoided doing this originally.

tomheaton:
> Fwiw, I also feel like having this filled in permanently is far more visually "loud" than the outline, and I don't think it makes sense to be wanting to draw the user's eye to this button continually.

that makes sense, but on the other hand it does look odd being empty when "enabled"

MattIPv4:
Perhaps we leave it an outline by default, and trying a lighter shade of green when it's enabled? That might solve the issue of it looking out of place, while also avoiding it drawing the eye?

MattIPv4:
I don't think so, they are fine to use our standard colors, as we _do_ want to draw the user's eye to those when they encounter them.

MattIPv4:
👋 @tomheaton just wanted to check in here on if you were able to make the changes discussed ^^?

tomheaton:
will get on with this right now

tomheaton:
`hasConsented` background color
image

`!hasConsented` no background color
image

MattIPv4 (approved):

MattIPv4:
(Not really sure what those failures are, I think Vercel is having a moment:tm: I couldn't approve the build so I had to trigger a manual redeploy.)

MattIPv4:
Thank you for your work on this 💚

PR #306, opened 2023-06-20
Implement Twitch Chatbot for Notifications

### LINKED ISSUE #294: Notifications: Allow sending notifications from chat using command(s)
## TODO

- [x] Implement chat bot (long-running node server)
  - [x] Logged in twitch (bot) user with OAuth auth flow to allow feedback (aka send chat messages)
  - [x] Listen to chat rooms (`alveussanctuary`, `alveusgg`, …)
  - [x] Check user perms (only a list of users + channel owner)?
  - [x] Call Notification API from chat bot
  - [x] Feedback in Chat (possibly as dm)
    - [x] Require confirmation
    - [ ] ~~Allow cancelling in the first 20 seconds?~~
- [x] Implement Notification API

## Chat Commands

- `!notification [url] title ... [| message text ...]` Create a stream notification
  - `[url]` Link, optional, defaults to https://twitch.tv/alveussanctuary
- `!announcement [url] [start] [end] title ... [| message text ...]` Create an announcement
  - `[url]` Link, optional, defaults to empty (no link)
  - `[start]` Start Date/Time in the format `YYYY-MM-DD[ HH:II]`, optional, defaults to empty
  - `[end]` End Date/Time in the format `YYYY-MM-DD[ HH:II]`, only if `[start]` is present, optional, defaults to empty

### Aliases

- `!notify`
- `!announce`

### Flags

- `--no-push` to disable push
- `--no-discord` to disable discord
- `--no-site` to disable showing on website (when implemented)
- `--image=[url]` to attach an image by URL (has to be png/jpg/gif)

pjeweb:
Resolves #294

## Website - Twitch channel management and tokens

agg-twitch-api

- this allows users with the `twitchApi` role to manage twitch credentials and supply moderator/broadcaster credentials for a configured twitch channels

## Chatbot

<table>
<tr>
<td>  </td>
<td>  </td>
<td>  </td>
</tr>
<tr>
<td>Successful workflow stream notification</td>
<td>Denied </td>
<td>Denied confirm </td>
</tr>
<tr>
<td>  </td>
<td>  </td>
</tr>
<tr>
<td>Announcement</td>
<td>Trying to confirm without pending</td>
</tr>
</table>

### Implementation

- using twurple to handle most twitch things (api & chatbot)
- using drizzle ORM to get data from the site. it might be cleaner to implement a API and keep the db solely controlled by the next app or use prisma like the site does. this was basically a test balloon if drizzle might be a replacement better for prisma for the site as it is smaller and works in edge envioronments.

### Deployment

- using Node.js and `tsx` to run the server without any pre-transpilation or bundling keeping the setup easy. We might want to use some sort of transpiler and/or bundler (webpack/tsc/esbuild etc.) at some point and build the chatbot using gh actions, especially if we want to take advantage of lockfiles and/or packages from the workspace, sharing code with the site.
- the chatbot runs on the hetzner machine as a docker service as defined in the docker-compose.yaml using the basic node docker image
- the env config is currently in a `.env` file directly on the server; there are currently two deployment environments (preview and production)
- using a ssh gh action it checks out the repo on the server, rebuilds and re-ups the service. Secrets are set in GH
- the gh action workflow is set up such that only pushes to `main` and `preview` trigger a deployment
- but: the site always gets deployed. it might be worth at some point to figure out how to prevent vercel from deploying if the site has not changed


.github/workflows/update-chatbot.yml
```diff
@@ -0,0 +1,71 @@
...
+          port: ${{ secrets.CHATBOT_PORT }}
+          passphrase: ${{ secrets.CHATBOT_PASSPHRASE }}
+          script_stop: true
+          script: |
+            cd $CHATBOT_PATH
+            git checkout -b tmp-dispose-local
+            git branch -D $REF_NAME
+            git fetch origin $REF_NAME
+            git checkout -f $REF_NAME
+            git branch -D tmp-dispose-local
```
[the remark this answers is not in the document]
MattIPv4: Would it not be more efficient to do `git fetch origin $REF_NAME && git reset --hard origin/$REF_NAME`?

apps/chatbot/src/bot/commands/notification.ts
```diff
@@ -0,0 +1,302 @@
...
+    const { restParams: paramsWithoutChannelOptions, ...options } =
+      parseOptions(params);
+    const { linkUrl, restParams } = parseOptionalLinkUrl(
+      broadcasterName,
+      paramsWithoutChannelOptions
+    );
+    let { title, text } = parseTitleAndText(restParams);
+
+    if (!title) {
+      reply("Waiting Creating notification...");
```
pjeweb, line 180: seems unnecessary to send this. it can however take a bit to fetch the channel info

apps/chatbot/src/bot/commands/notification.ts
```diff
@@ -0,0 +1,302 @@
...
+    const pendingNotification = pendingNotifications.get(key);
+
+    if (!pendingNotification || pendingNotification.expiresAt < Date.now()) {
+      if (pendingNotification) pendingNotifications.delete(key);
+
+      reply("mojjcheck no pending notification for user");
+      return;
+    }
+
+    reply(`Waiting sending notification …`);
```
pjeweb, line 274: seems unnecessary to send this

apps/chatbot/src/env.ts
```diff
@@ -0,0 +1,70 @@
...
+    NODE_ENV: z
+      .enum(["development", "test", "production"])
+      .default("production"),
+
+    // Chatbot
+    BOT_USER_ID: z.string().default("858050963"),
+    BOT_CHANNEL_NAMES: namesSchema.default(["AlveusGG"].join(",")),
+
+    // Users
+    MODERATOR_USER_NAMES: namesSchema.default(
```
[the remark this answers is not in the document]
pjeweb: i think this is fine with the simplified git checkout

.github/workflows/update-chatbot.yml
```diff
@@ -0,0 +1,71 @@
...
+            cd apps/chatbot
+            docker compose up --build --detach
+
+  update-chatbot-preview:
```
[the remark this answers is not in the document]
pjeweb: i think this is fine with the simplified git checkout

MattIPv4:
> using drizzle ORM to get data from the site. it might be cleaner to implement a API and keep the db solely controlled by the next app or use prisma like the site does. this was basically a test balloon if drizzle might be a replacement better for prisma for the site as it is smaller and works in edge environments.

Could we open an issue for this, so we don't forget about it? I'm always a fan of lighter-weight dependencies :)

.github/workflows/update-chatbot.yml
```diff
@@ -0,0 +1,62 @@
...
+            cd $CHATBOT_PATH
+            git fetch origin $REF_NAME && git reset --hard origin/$REF_NAME
+            cd apps/chatbot
+            docker compose up --build --detach
+
+  update-chatbot-preview:
+    name: Update Chatbot (Preview)
+    runs-on: ubuntu-latest
+    environment: preview
+    if: github.ref != 'refs/heads/main'
```
MattIPv4, line 46: Could I suggest gating this specifically to preview, just to be safer?

apps/chatbot/lint-staged.config.cjs
```diff
@@ -0,0 +1,9 @@
+const mainConfig = require("../../lint-staged.config.js");
+
+const buildEslintCommand = (filenames) =>
+  `pnpm run lint --fix ${filenames.join(" ")}`;
```
MattIPv4, line 4: `pnpm run lint` is an alias to `eslint .`, so it will always run against all files, even if you're passing extra file names here. This should probably just interface with eslint directly.

MattIPv4:
> We might want to use some sort of transpiler and/or bundler (webpack/tsc/esbuild etc.) at some point and build the chatbot using gh actions, especially if we want to take advantage of lockfiles and/or packages from the workspace, sharing code with the site.

Could we also open an issue for this? I think using `tsx` is fine, I dislike having to transpile server-side code prior to running it, but it'd be great to figure out ensuring we have the protection of the lockfile.

apps/chatbot/Dockerfile
```diff
@@ -0,0 +1,13 @@
+FROM node:18-alpine
```
MattIPv4, line 1:
```suggestion
FROM node:18.15.0-alpine
```
Just to match what's defined in our `.nvmrc`

apps/chatbot/lint-staged.config.cjs
```diff
@@ -0,0 +1,9 @@
...
+
+module.exports = {
+  ...mainConfig,
+  "*.{js,jsx,ts,tsx}": buildEslintCommand,
```
MattIPv4, line 8:
```suggestion
  "*.{js,ts}": buildEslintCommand,
```
I don't think we're going to have any JSX in here?

apps/chatbot/prettier.config.cjs
MattIPv4, line 1: Is this file needed, it seems like an empty stub?

apps/chatbot/src/bot/commands/notification.ts
```diff
@@ -0,0 +1,302 @@
...
+  const notificationCommand: CommandHandler = async (
+    params,
+    { broadcasterId, broadcasterName, userName, userId, reply }
+  ) => {
+    const isMod = await checkUserIsAllowedToSendNotifications(userName);
+    if (!isMod) {
+      reply("mayaHalt you are not allowed to send notifications");
+      return;
+    }
+
```
MattIPv4, line 170: Could we add a help flag here, that explains the flags and message format?

apps/chatbot/src/bot/commands/notification.ts
```diff
@@ -0,0 +1,302 @@
...
+  const announcementCommand: CommandHandler = async (
+    params,
+    { broadcasterId, broadcasterName, userName, userId, reply }
+  ) => {
+    const isMod = await checkUserIsAllowedToSendNotifications(userName);
+    if (!isMod) {
+      reply("mayaHalt you are not allowed to send announcements");
+      return;
+    }
+
```
MattIPv4, line 216: Ditto

apps/chatbot/src/bot/commands/version.ts
```diff
@@ -0,0 +1,18 @@
+import { readPackage } from "read-pkg";
+
+import type { CommandHandler } from "@/bot/commands/base";
+import { createBotCommand } from "@/bot/commands/base";
+
+const versionPromise = readPackage().then(({ version }) => version);
```
MattIPv4, line 6: Would the commit hash from git be more useful here? We can probably just shell out to `git rev-parse --short HEAD` to get it?

apps/website/src/server/trpc/router/forms.ts
```diff
@@ -119,12 +112,17 @@ export const formsRouter = router({
       if (config.checks) {
         // TODO: Make the form config granular. Right now the channel follow check is hard-coded here:
         // NOTE: Does not work, twitch removed API access :(
```
MattIPv4, line 114: Is this comment outdated now?

MattIPv4:
> but: the site always gets deployed. it might be worth at some point to figure out how to prevent vercel from deploying if the site has not changed

Worth creating an issue for this as well? https://vercel.com/docs/concepts/projects/overview#ignored-build-step

MattIPv4:
From my understanding of the PR, it seems like we're now allowing admin users to auth more scopes, and then that auth gets used for features like follower checking, twitch event sub etc.?

I see that we're displaying when that token expires -- do we have logic that's refreshing tokens automatically in the background, so they don't actually ever expire?

apps/chatbot/src/bot/commands/notification.ts
pjeweb: I added some help, but its not super readable because we can not add line breaks. Maybe we can condense the help a bit?

gg-help

apps/chatbot/src/bot/commands/version.ts
pjeweb: Right now the chatbot runs in the docker without git installed. I am not sure if it makes sense to install git just for this or maybe we can write the ref to a file with GH actions instead?!

pjeweb:
> From my understanding of the PR, it seems like we're now allowing admin users to auth more scopes, and then that auth gets used for features like follower checking, twitch event sub etc.?

It could be used for that, yes. Right now that just uses the long-lived app auth (OAuth).

> I see that we're displaying when that token expires -- do we have logic that's refreshing tokens automatically in the background, so they don't actually ever expire?

The chatbot does that for the token used there, because `twurple` has that built in. It should be fairly easy to use the same library to refresh mod/broadcaster tokens using a cron job, but thats not implemented yet.

apps/chatbot/src/bot/commands/version.ts
MattIPv4: Ah yeah -- I would pass it in as a build arg and then expose it as an env var for the bot?

MattIPv4:
Ah okay, so is the new auth logic with more scopes just for the chatbot currently? If so, twurple handling refresh seems great for now.

apps/chatbot/src/bot/commands/version.ts
pjeweb: I got it running using a runtime env variable, but i think that should be fine as well?! 4157d95dfbcb6ae3b55598fe2d45dd2e127e1ad2

MattIPv4: Works for me 👍

pjeweb:
> Ah okay, so is the new auth logic with more scopes just for the chatbot currently? If so, twurple handling refresh seems great for now.

Yeah thats the state with this PR.

Prisma to Drizzle issue is opened (#320). I'll open issues for the lockfile situation and for token refreshing after this PR.

MattIPv4 (dismissed):
All seems fine to me reading through it, I trust it actually works when running.

Looks like you've got some conflicts with dependencies

MattIPv4 (approved):

PR #317, opened 2023-06-30
Use typescript checking for some js config files, remove unnecessary react aria Proivder

pjeweb:
- `react-aria` does not need a SSRProvider with React 18
- add `@ts-check` directives to some js config files to enable typescript type checking, add `@types/prettier` for prettier config

### DIFF
--- a/package.json
+++ b/package.json
@@ -19,6 +19,7 @@
   },
   "packageManager": "pnpm@8.6.0",
   "devDependencies": {
+    "@types/prettier": "^2.7.3",
     "husky": "^8.0.3",
     "lint-staged": "^13.2.2",
     "prettier": "^2.8.8"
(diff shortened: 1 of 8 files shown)


package.json
```diff
@@ -19,6 +19,7 @@
   },
   "packageManager": "pnpm@8.6.0",
   "devDependencies": {
+    "@types/prettier": "^2.7.3",
```
MattIPv4, line 22: Should this be a dependency of the website? Typescript isn't a global dependency, and we only have a typescript config in the website?

pjeweb: yeah makes sense. i had it there because i was working on both the site and chatbot, but keeping it separate makes sense

MattIPv4 (approved):

PR #326, opened 2023-07-16
Fix merch scroller sizing on mobile

MattIPv4:
On mobile, the PiP images were overflowing and colliding with other items. Each item has a padding of 2, so we can safely move the image -2 into the corner without it overlapping. This also ensures the PiP remains half the width of the main image.


pjeweb (approved):
hmm i tested on very smol mobile but not the in-between. Looks great now

PR #334, opened 2023-07-19
Add promos section to entry forms

### LINKED ISSUE #333: Promote merch + updates on giveaways
On the giveaways, it'd be nice to be able to promote the merch (maybe similar to how we do it on the homepage), and also to promote the updates page and being able to get push notifications (not sure how to promote this, maybe just a one-liner with a button, clean + simple [given its also in the footer]).

In both cases, these shouldn't be "entry" items in the giveaway, more just promotional content on the pages to get more eyes on them, and to add some more content to the pages so they aren't just a form.

MattIPv4:
Resolves #333 

Also, I did a small refactor to the giveaway checks as I was noticing some odd behaviour with them not checking themselves locally, and I refactored/restyled the entry form a bit generally, to make the promos fit better in the form flow.

### DIFF
--- a/apps/website/src/components/forms/EntryForm.tsx
+++ b/apps/website/src/components/forms/EntryForm.tsx
@@ -3,23 +3,26 @@ import { useSession } from "next-auth/react";
 import type { Form } from "@prisma/client";
 
 import { LoginWithTwitchButton } from "@/components/shared/LoginWithTwitchButton";
-import { Headline } from "@/components/shared/Headline";
 import { Button } from "@/components/shared/Button";
 import { MessageBox } from "@/components/shared/MessageBox";
+
+import Markdown from "@/components/content/Markdown";
+import Heading from "@/components/content/Heading";
+import Section from "@/components/content/Section";
 
 import { getCountryName } from "@/utils/countries";
 import { trpc } from "@/utils/trpc";
 import { calcFormConfig } from "@/utils/forms";
-import Markdown from "@/components/content/Markdown";
+
 import type { FormEntryWithAddress } from "@/pages/forms/[formId]";
 
-import { ConsentFieldset } from "@/components/forms/ConsentFieldset";
-
+import { ConsentFieldset } from "./ConsentFieldset";
 import { GiveawayChecks } from "./GiveawayChecks";
 import { ShippingAddressFieldset } from "./ShippingAddressFieldset";
 import { NameFieldset } from "./NameFieldset";
 import { ContactFieldset } from "./ContactFieldset";
 import { EntryRulesFieldset } from "./EntryRulesFieldset";
+import { Promos } from "./Promos";
 
 export const EntryForm: React.FC<{
   form: Form;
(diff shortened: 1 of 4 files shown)


apps/website/src/components/forms/EntryForm.tsx
```diff
@@ -55,92 +58,182 @@ export const EntryForm: React.FC<{
...
+              <p className="m-2">Check the details you submitted below.</p>
+            </MessageBox>
+          </Section>
+
+          <Section dark>
+            <Promos />
+          </Section>
+
+          <Section containerClassName="max-w-xl">
+            <Heading level={2} className="mb-6 text-2xl">
```
pjeweb, line 119: I do not really the promo here and i don't think its that necessary.

```suggestion
            <Heading level={2} className="mb-6 mt-12 text-2xl">
```

You could also make the section extra dark (`className="bg-alveus-green-green-900"`) and move it below the entry, but then we already have the socials section there.

apps/website/src/components/forms/GiveawayChecks.tsx
```diff
@@ -15,8 +16,46 @@ const GiveawayCheck: React.FC<{
...
+      )}
+    >
+      <span className="sr-only">{label}</span>
+      <input
+        name={`req-${name}`}
+        value="yes"
+        type="checkbox"
+        defaultChecked={hasLink && isClicked}
+        required={true}
+        disabled={disabled}
```
pjeweb, line 55: The problem with disabling them is that the browser form validation ignores the checkboxes here. Thats why i did the weird `preventDefault()` thing.

apps/website/src/components/forms/EntryForm.tsx
```diff
@@ -55,92 +58,182 @@ export const EntryForm: React.FC<{
...
+              <Section containerClassName="max-w-lg">
+                {/* Surface errors at the start */}
+                {enterForm.error && (
+                  <MessageBox variant="failure">
+                    Error: {enterForm.error.message}
+                  </MessageBox>
+                )}
+
+                <Heading level={2} className="mb-6 text-2xl">
+                  Steps to Enter
```
pjeweb, line 189: To make the split form easier to understand, can add numbers to the parts? I'm using actions instead of steps so its less confusing.

```suggestion
                  1) Actions to Enter
```

apps/website/src/components/forms/EntryForm.tsx
```diff
@@ -55,92 +58,182 @@ export const EntryForm: React.FC<{
...
+          <Section containerClassName="max-w-lg">
+            {/* Surface errors at the start, if there was no checks section */}
+            {!config.checks && enterForm.error && (
+              <MessageBox variant="failure">
+                Error: {enterForm.error.message}
+              </MessageBox>
+            )}
+
+            <Heading level={2} className="mb-6 text-2xl">
+              Your Details
```
pjeweb, line 215: see above

```suggestion
              {config.checks ? "2)" : ""}
              Your Details
```

pjeweb (changes requested):

apps/website/src/components/forms/GiveawayChecks.tsx
MattIPv4: Ah that makes sense, didn't notice it wasn't validating it when disabled

apps/website/src/components/forms/EntryForm.tsx
MattIPv4: Yah, that's why I'd placed it in the middle here, as it felt a bit weird to have an updates promo right above another mention of updates again.

MattIPv4: Can drop it from here though, I think having it on the form with checks should be enough

pjeweb (approved):

PR #338, opened 2023-07-21
Forms: Last minute fixes

pjeweb:
- implement backend for the optional shipping address and marketing consent
- Improve consent wording and only include shipping address if applicable
- Make link in intro look better


apps/website/src/components/forms/EntryForm.tsx
```diff
@@ -80,7 +80,7 @@ export const EntryForm: React.FC<{
           <Heading>{form.label}</Heading>
         </header>
 
-        {config.intro && <Markdown content={config.intro} />}
+        {config.intro && <Markdown content={config.intro} dark={true} />}
```
MattIPv4, line 83:
```suggestion
        {config.intro && <Markdown content={config.intro} dark />}
```
`={true}` on props is redundant

apps/website/src/components/forms/EntryForm.tsx
```diff
@@ -219,7 +219,9 @@ export const EntryForm: React.FC<{
               />
               {config.requireShippingAddress && <ShippingAddressFieldset />}
               {config.hasRules && <EntryRulesFieldset form={form} />}
-              <ConsentFieldset />
+              <ConsentFieldset
+                withShippingAddress={config.requireShippingAddress}
+              />
```
MattIPv4, line 224: I would think it makes more sense to have the marketing consent be part of the consent section, rather than part of the section where you provide the details?

pjeweb: Sure, i don't have strong feelings, just felt very checkbox heavy down there plus it might be more clear what the email address entered is used for if the consent is right there?!

MattIPv4: In some ways I agree it's more apparent, but at the same time, I think I'd expect to see checkboxes relating to data processing being listed under data processing.

MattIPv4: With moving it to the bottom, it might also make sense to word it more like:

```
Additionally, I am happy for Alveus Sancatury Inc. to use my email for direct marketing relating to the Sanctuary, including from its partners, such as for news, fundraisers, and other promotions.
```

pjeweb: Okay that sounds good. Will do. We should probably still mention the partners since I understand Alveus wants to share the data with the merch company.

MattIPv4: Yah, I kept the partners mention in there

pjeweb: Oh im dumb. So hard to read in the scroll box :)

MattIPv4:
With shipping disabled locally, I'm unable to submit the form:

```
Error: [ { "code": "custom", "message": "Invalid country code", "fatal": true, "path": [ "country" ] } ]
```

MattIPv4 (approved):

PR #351, opened 2023-07-30
Update contact us page

MattIPv4:


MattIPv4:
This does potentially put the opensource@ email in the firing line a bit, but I hope it's buried enough on the page and contextual enough it shouldn't be a problem. If it is, we can always pull it and just link it to the security policy to abstract it a bit more.

apps/website/src/pages/contact-us.tsx
```diff
@@ -20,43 +19,87 @@ const ContactUsPage: NextPage = () => {
...
+            </p>
 
-        <p className="my-3 text-xl font-semibold">We are not openly hiring.</p>
+            <p className="my-3 text-xl font-semibold">
+              For plushie inquiries: <br />
+              <Link href="https://youtooz.com/pages/contact-us" external>
+                youtooz.com/pages/contact-us
+              </Link>
+            </p>
+          </div>
```
pjeweb, line 52: this aligns weird on mobile, probably needs some sort of center aligning or flex-grow or smth. i'm too lazy to open the code though Smile

pjeweb (changes requested):

pjeweb (approved):

PR #353, opened 2023-08-01
Discord notifications: wrap links in text to prevent multiple embeds (#313)

### LINKED ISSUE #313: Discord notifications: wrap links in text to prevent multiple embeds
as suggested by @MattIPv4 find links (using https://github.com/markdown-it/linkify-it) in notification text and wrap them in `<>` to prevent additional embeds in discord

nslee333:
## Describe your changes

Closes #313 

Added linkify-it to project, and used it to parse discord notification and wrapped in angle brackets to prevent multiple embeds.

This is my first attempt at it.

I'm not sure how I can test this locally. I went off of the #313 description, is there a way that I can test this locally?


pjeweb:
> ## Describe your changes
> Closes #313
> 
> Added linkify-it to project, and used it to parse discord notification and wrapped in angle brackets to prevent multiple embeds.
> 
> This is my first attempt at it.

Thanks for working on this!

Some notes on the draft:
I would try to not explicitly type variables if they are already inferred from the right hand side (I'll add a comment on the code).

Also i would prefer if we could extract the link marking into a helper function (and file). That also will help with testing (see below).

> I'm not sure how I can test this locally. I went off of the #313 description, is there a way that I can test this locally?

Two ideas.

1) This is a great use case for unit tests imho. You could put a `discord.test.ts` or something like this in the `__tests__` directory and test a few cases of text with "links". It would be great to test both fully qualified URLS `https://google.com` but also more importantly partial links like `twitch.tv/alveussanctuary` without a scheme etc. Maybe add some text that could be falsely identified as a link like a full stop between sentences.

2) If you want to actually test this, you can set up your own Discord server, add a channel for testing and create a webhook URL for that channel (https://support.discord.com/hc/en-us/articles/228383668). Add the url to `DISCORD_CHANNEL_WEBHOOK_URLS_STREAM_NOTIFICATION` and `DISCORD_CHANNEL_WEBHOOK_URLS_ANNOUNCEMENT` in your `.env`. You should be able to send discord notifications now from you local server at `http://localhost:3000/admin/notifications`. This assumes that you already set up authentication (including the OAuth process at dev.twitch.tv) and added yourself as a super user in the `.env` or through the role system. If you want to skip the auth setup it might be possible to replace `getAdminSSP` in `server/utils/admin.ts` with a stub like this: `export const getAdminSSP = async () => ({isSuperUser: true, menuItems});`. But of course do not commit this change. If you don't have all the push env variables set up you might also need to uncheck the push checkbox. If this does not work, let me know and we can try to find a way to test this for you.

apps/website/src/server/notifications.ts
```diff
@@ -200,6 +201,24 @@ async function createDiscordNotifications({
     default:
   }
 
+  const linkify = new LinkifyIt();
+
+  const matches: LinkifyIt.Match[] | null = linkify.match(message);
```
pjeweb, line 206:
```suggestion
  const matches = linkify.match(message);
```

apps/website/src/server/notifications.ts
```diff
@@ -200,6 +201,24 @@ async function createDiscordNotifications({
...
+
+  let offset = 0;
+  if (matches != null) {
+    for (const match of matches) {
+      const linkStart: number = match.index + offset;
+      const linkEnd: number = match.lastIndex + offset;
```
pjeweb, line 212:
```suggestion
      const linkStart = match.index + offset;
      const linkEnd = match.lastIndex + offset;
```

apps/website/src/server/notifications.ts
```diff
@@ -200,6 +201,24 @@ async function createDiscordNotifications({
...
+
+      const linkified = `<${match.text}>`;
+      message =
+        message.slice(0, linkStart) + linkified + message.slice(linkEnd + 1);
+
+      offset += linkified.length - (match.lastIndex - match.index + 1);
+    }
+  }
```
pjeweb, line 220: As noted in the comment, I would suggest moving this into a separate helper function like `escapeLinksForDiscord(text: string)` and move it in its own util file.

nslee333:
Alright I got it extracted to it's own util file, and removed the explicit types. Is it common for variables to just be inferred? I haven't really used TS in serious project before. 

I'm going to add a few unit tests first and then go from there.

MattIPv4:
One quick note, it looks like you've installed the dependencies in the root package, rather than the website package :)

nslee333:
Oh whoops, I'll get that fixed.

pjeweb:
>  Is it common for variables to just be inferred?

Yeah i would say so. There might be other opinions, but i think its fairly common to only use type annotations when necessary or also when writing non-trivial function signatures where they act as sort of contract or API. Although opinions differ, whether you should let typescript infer return types or whether it should always be explicitly set in the function signature.

I would argue it makes the code less verbose and easier to read. Your IDE or editor should be able to show you the inferred type, either inline or when hovering the variable. Using explicit type annotation can also widen the type unnecessary, but i don't think thats as big of a problem or it might actually widen it on purpose in which case you should of course have the annotation.

nslee333:
> I would argue it makes the code less verbose and easier to read. 

That makes sense, I'm all for cleaner / less verbose code. Especially since TS doesn't require explicit variables to enforce types, why not keep it cleaner.

> Maybe add some text that could be falsely identified as a link like a full stop between sentences

Kinda like `... day.We're ...`? Where there isn't a space after the full stop?

nslee333:
@pjeweb When you get a minute, I could use some clarification on above.

pjeweb:
> @pjeweb When you get a minute, I could use some clarification on above.

the tests you have look good to me

apps/website/src/server/notifications.ts
```diff
@@ -200,6 +202,26 @@ async function createDiscordNotifications({
...
+
+  message = escapeLinksForDiscord(message);
```
pjeweb, line 223: that should be able to go now (also the `import LinkifyIt from "linkify-it";`)

```suggestion
  message = escapeLinksForDiscord(message);
```

pjeweb (approved):

PR #363, opened 2023-08-06
Fixed small gap in social footer below video

tomoyukiy:
## Describe your changes
resolve #358 

alveusgg_pr

...

## Notes for testing your change

...

PR #364, opened 2023-08-06
Remove reliance on negative margin in socials

MattIPv4:
## Describe your changes

Follow-up to #363, just so that we don't continue to have reliance on a negative margin that breaks if the size of the content changes.

This change should keep the video pinned to the bottom no matter how large the content on the right is:

image

And should keep the content on the right pinned to the center no matter how small it is compared to the video:

image

### DIFF
--- a/apps/website/src/components/layout/footer/Socials.tsx
+++ b/apps/website/src/components/layout/footer/Socials.tsx
@@ -27,9 +27,9 @@ const Socials: React.FC = () => {
 
   return (
     <Section dark className="z-0 pb-0">
-      <div className="flex flex-wrap-reverse items-center">
-        <div className="basis-full md:basis-1/2 md:pr-8 md:pt-0">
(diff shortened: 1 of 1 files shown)


pjeweb:
I mean the negative margin was not supposed to do any layout work, just allow the video to take up a bit more of the top of the section:
socials-margin

Its not super important but i prefer the smaller gap there. Using margin auto or `self-start` or something is a good idea though. The reversed flex wrap makes the flex-start and flex-end a bit confusing, so using top margin looks good to me. I should've kept the other PR open for you to review :)

MattIPv4:
Ah I see -- why wasn't the padding just reduced rather than using a negative margin?

pjeweb:
> Ah I see -- why wasn't the padding just reduced rather than using a negative margin?

Can't remember. I don't mind negative margins in general, so it probably just seemed like the easiest solution at that point? I'll try not to use them anymore ;)

pjeweb (approved):

PR #365, opened 2023-08-07
Fix host name on Animal Quest pages

MattIPv4:
## Describe your changes

#353 accidentally bumped `@alveusgg/data` to `0.18.0`, which included a change to how Animal Quest hosts are stored (https://github.com/alveusgg/data/pull/31).

This PR fixes how we render those hosts, so that the page uses the new `hosts` object to get the host data (name, etc.).

Access an Animal Quest page, confirm host's name shows as expected.

### DIFF
--- a/apps/website/src/components/content/Link.tsx
+++ b/apps/website/src/components/content/Link.tsx
@@ -1,4 +1,4 @@
-import React from "react";
+import React, { useMemo } from "react";
 import NextLink from "next/link";
 
 import { classes } from "@/utils/classes";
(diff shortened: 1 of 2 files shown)


MattIPv4:
FYI @jinitsuga, I ended up just doing this as needed to fix it due to the version being bumped :)

pjeweb (approved):
sorry that i missed that in the review 😄

jinitsuga:
> FYI @jinitsuga, I ended up just doing this as needed to fix it due to the version being bumped :)

Hey! That's awesome. Sorry I couldn't get to it earlier, life got a bit crazy. Guess I'll continue with the cards :D

PR #371, opened 2023-08-13
Update dependencies

MattIPv4:
## Describe your changes

This updates all the dependencies throughout the stack, and makes the appropriate code changes to support the updates.

The one exception to the update is `@types/node` that I have now properly pinned to `18.15.0`, to match the Node.js version we're using.

Please test this thoroughly, given the substantial number of updates, including to core dependencies like Prisma.


MattIPv4:
@pjeweb I haven't pushed this to preview yet just in case you were using it for something, but feel free to push this branch there when you're around to test :) Would appreciate a thorough set of eyes looking over this!

MattIPv4:
(Fwiw, once this is in I'm going to look at cleaning up some of the old WIP stuff we probably don't need in main now (can always revive it from git history), and do a general audit of dependencies 'cause it does feel like we have quite a few)

pjeweb (changes requested):
Tested pretty much everything. Looks good.

https://github.com/alveusgg/alveusgg/blob/f54f005480d0b2b2911d50564dd3354a86e89bf8/apps/chatbot/src/bot/index.ts#L11 should be
```typescript
const bot = new Bot({
```
dropping the first arguments with twurple 7 (https://github.com/twurple/twurple/commit/0e6ca6bd1680e572ffa5cc3b84480ae9ef0567c0)

pjeweb:
> @pjeweb I haven't pushed this to preview yet just in case you were using it for something, but feel free to push this branch there when you're around to test :) Would appreciate a thorough set of eyes looking over this!

oh right yeah pushed it rn

MattIPv4:
Alright, this should be good for another round of testing. Apologies for the delay.

PR #372, opened 2023-08-14
Add bell peepo animation to updates page

MattIPv4:
## Describe your changes

Adds the new bell peepo animation from Dion to the updates page header, on the right.

Check mobile + desktop.


pjeweb:
is it me or does the blurred inline svg ginormous? its loading super slow on my m1 pro macbook and iphone

MattIPv4:
Yeah, noticed the same, the blur placeholder was making things much worse and looked terrible. Removed it. I suppose if we really wanted we could do some custom logic to load a freeze-frame first and then load the full thing, but I think this is probably fine as-is now? It's still 8mb, but seems to load pretty quickly given it's not going through optimization.

pjeweb (dismissed):
still somehow needs 30+ seconds on ios, but i guess thats me. loads fine on the laptop with the same internet

pjeweb:
loading-slow
 its consistently bad with browser cache disabled. it somehow blocks and inflates the js bundle?!

MattIPv4:
After a bunch of source diving and debugging, it seems that this is a bug in Next.js 13.4.7 that is partially fixed in 13.4.8 and future versions -- I've not managed to track down the actual PR that fixed this, but updating locally definitely fixes the bundle size.

If you use next/image the blurDataURL still inflates the URL, but it looks like in 13.4.8 when using a pure image tag, the blurDataURL gets tree-shaken correctly.

I've included the dependency update here for now, but really I think it'd probably be better to land #371 first, and then I can rebase this and clean it up.

MattIPv4:
I've filed https://github.com/vercel/next.js/issues/54012 to report this to the Next.js folks

MattIPv4:
@pjeweb this should be good for another review now -- my partial fix for this issue was merged into Next.js, so the blur data URL is never generated for animated images in the first place :)

pjeweb (approved):

PR #377, opened 2023-08-20
Add keyboard navigation and focus management to Show and Tell

pjeweb:
## Describe your changes

This will move the focus to the current ShowAndTellEntry when scrolling to it, allowing better keyboard accessibility and moving the focus to within the gallery component so we can bind a `keydown` handler to handle arrow keys. For the infinite loading we see if a new page has been added and move the focus to the first new entry, once its rendered. For this we save the id in a ref when the new data has loaded and use an DOM attribute to find it (instead of keeping a ref to every item).

Go to `/show-and-tell` and test that scrolling and arrow buttons work as expected and also, once focus has been set by tabbing into the gallery, clicking an entry or entering presentation mode (fullscreen), check that arrow keys will move the focus to the next/prev entry and scroll it into view.

### DIFF
--- a/apps/website/package.json
+++ b/apps/website/package.json
@@ -111,7 +111,7 @@
     "npm-run-all": "^4.1.5",
     "postcss": "^8.4.27",
     "postcss-nesting": "^12.0.1",
-    "prettier-plugin-tailwindcss": "^0.5.2",
+    "prettier-plugin-tailwindcss": "^0.5.3",
     "prisma": "^5.1.1",
     "tailwindcss": "^3.3.3",
     "tsup": "^7.2.0",
--- a/apps/website/src/components/show-and-tell/ShowAndTellEntry.tsx
+++ b/apps/website/src/components/show-and-tell/ShowAndTellEntry.tsx
@@ -177,7 +177,7 @@ export const ShowAndTellEntry = forwardRef<
     <article
       key={entry.id}
       className={
-        "relative flex flex-shrink-0 flex-col transition-opacity delay-500 duration-500 " +
+        "relative flex flex-shrink-0 flex-col transition-opacity delay-500 duration-500 focus:outline-none " +
         (isPresentationView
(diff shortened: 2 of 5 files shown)


MattIPv4:
👀 In non-full-screen mode, its pretty easy to lose the focus and then no longer be able to keyboard navigate -- should we lock this behind full-screen only?

Also, we should probably remove the native browser focus styling?

pjeweb:
> 👀 In non-full-screen mode, its pretty easy to lose the focus and then no longer be able to keyboard navigate -- should we lock this behind full-screen only?

I might be missing something, but you should only loose focus if you click outside the gallery or tab outside of it. We can't really lock that, can we?! My suggestions would be attaching the key handler to the body/document or maybe we could make the key event target the full width of the viewport so you can at least not click out of it on the sides.

> Also, we should probably remove the native browser focus styling?

I was okay with focus-visible styles since they'll only show when users use the arrow keys and I would consider those advanced users that understand focus rings?! But I would also be happy to remove them since the focus is pretty obvious with the entry taking up most of the viewport.

MattIPv4:
> I might be missing something, but you should only lose focus if you click outside the gallery or tab outside of it. We can't really lock that, can we?! My suggestions would be attaching the key handler to the body/document or maybe we could make the key event target the full width of the viewport so you can at least not click out of it on the sides.

Yeah I don't think it's that much of an issue thinking through it some more 👍 

> I was okay with focus-visible styles since they'll only show when users use the arrow keys and I would consider those advanced users that understand focus rings?! But I would also be happy to remove them since the focus is pretty obvious with the entry taking up most of the viewport.

Yeah I think it should be clear enough what's happening without them, and its intuitive and results in a cleaner looking page.

MattIPv4:
Hm, I'm still seeing the outline when I navigate with the keys?

pjeweb:
> Hm, I'm still seeing the outline when I navigate with the keys?

oops yeah i did not check firefox. should be fixed now. `outline-width: 0` is not enough for FF

MattIPv4 (approved):

PR #381, opened 2023-08-24
Add native to information on ambassador pages

MattIPv4:
## Describe your changes

cc https://github.com/alveusgg/data/pull/22

Adds "native to" information to ambassador pages, and restyles the stats/facts section of the ambassador page to more closely resemble a table rather than a side-by-side section which made the information feel rather cluttered.

All ambassadors' native to information seems correct.


pjeweb (dismissed):
The layout is definitely better. Would you consider going a step further and downgrade the headings to something less big, maybe no serifs? And at that point maybe use a `<dl><dt><dd>` list? That way you can make it semantic markup but and also do not need to make a custom heading? But otherwise this is good enough as is :)

agg-ambassador-specs

MattIPv4:
Good call on switching it to a proper definition list!

PR #382, opened 2023-08-25
Cleanup WIP pages and related code

MattIPv4:
## Describe your changes

We had assorted WIP pages over time in the codebase and it looks like we were down to just one, so I think we're probably safe to remove it at this point. The magic of git being that we can always reference back to here if we do need it.

This PR also cleans up assorted components and dependencies that appear to be unused as well, as a result of this WIP page being removed and likely others in the past.

Make sure nothing was removed that is required (build would fail if that were the case).

Make sure nothing unused is left in the codebase.


pjeweb:
I think we can remove my wip push canvas image too. I do not think we want a separate twitter and IG for updates at the moment. Should I do that? Its a few files here and there, the `canvas-hypertxt` dep afaik and the `NEXT_PUBLIC_FEATURE_NOTIFICATIONS_ADVANCED` env var.

MattIPv4:
Ah good call, might be easier for you to do it if you have the context on it? Welcome to push to this branch or I can merge this and you follow up?

pjeweb:
> Ah good call, might be easier for you to do it if you have the context on it? Welcome to push to this branch or I can merge this and you follow up?

then i would try to push it here. There are also a few twitch (API) features that we do not use (using the event sub etc.). We do not really use them. That might be another PR though as I might take a moment to cut that out cleanly.

pjeweb:
The push image should be removed.

pjeweb (approved):
Looks good to me now Smile

PR #385, opened 2023-08-25
Cleanup React imports + FC usage

MattIPv4:
## Describe your changes

We shouldn't need to be importing React into every JSX context as Next.js uses the modern react-jsx transform (it would've also handled importing React for us anyway prior to this afaik).

The use of React.FC is also discouraged, with the preferred approach just being to let the typing be implied from the functional components themselves, so this removes all usage of the FC type.

Everything builds, all React imports and FC usages cleaned up.


pjeweb (approved):
ty! looks good and everything works

PR #387, opened 2023-08-28
Add Esfand collab

pjeweb:
## Describe your changes

Add esfand collab

Check name, date and thumbnail are correct and that the video plays


MattIPv4 (approved):
Ty for doing this, completely slipped my mind before I disappeared for the last couple of days

PR #392, opened 2023-08-31
Restrict TheGivingBlock embed width to full

### LINKED ISSUE #389: [Bug]: GivingBlock embed overflows on mobile

On mobile, on the donate page, the giving block embed causes the page to horizontally overflow:

image

It should be constrained to the width of the page.

### Steps to reproduce

https://www.alveussanctuary.org/donate on mobile, consented to loading the embed.

### Browser specs

iPhone 11 simulated in Firefox

### Anything else?

_No response_

MattIPv4:
## Describe your changes

Resolves #389

On small viewports, the giving block embed is restricted to the width of the page and does not overflow the page.


pjeweb (approved):
fancy tailwind class

PR #393, opened 2023-08-31
Wait for images to load before animating slideshow

MattIPv4:
## Describe your changes

While the slideshow seemed to behave great in Chrome, in Firefox I had noticed it sometimes really struggles if the images aren't cached, trying to render and animate the images at the same time.

This PR updates the logic in the slideshow, making use of the [`onLoadingComplete`](https://nextjs.org/docs/pages/api-reference/components/image#onloadingcomplete) prop from `next/image`, to require that the first two images of the slideshow have loaded fully before we start the animation that will transition between them. Hopefully, before we get to the third image and so on they've had time to naturally load.

This also removes the `will-change` declarations for the animations, as it appears this was also causing issues with Firefox due to it trying to allocate too much memory for all the images. Instead, we'll just let the browser figure out what to optimize.

In both Chrome and Firefox, the homepage slideshow continues to visually look correct, and doesn't appear to stress the browser out.

### DIFF
--- a/apps/website/src/components/content/Slideshow.tsx
+++ b/apps/website/src/components/content/Slideshow.tsx
@@ -1,6 +1,12 @@
 import type { ImageProps } from "next/image";
 import Image from "next/image";
-import { type CSSProperties, useCallback, useId, useMemo } from "react";
+import {
+  type CSSProperties,
+  useCallback,
+  useId,
+  useMemo,
+  useState,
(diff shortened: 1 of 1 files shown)


MattIPv4:
In conclusion, none of these ideas seem to make it better, Firefox is just really bad at rendering large images. Was worth trying though.

PR #403, opened 2023-09-16
Remove links to merch store, replace merch carousel with plushies

pjeweb:
## Describe your changes

- Removes links and redirects to the merch store.
- Replace apparel merch in merch carousel on homepage with plushies

- pages should still look okay (home page and general footer)
- menu should not have apparel merch link anymore
- the merch carousel on the home page should show the plushies and link to the plush store
- redirects should redirect to the merch carousel on the home page:
   - /merch
   - /merch-store
   - /apparel
   - /store
   - /shop


pjeweb:
I'm not 100% if we should disable the redirects. Maybe its enough to make them temporary redirects?!

MattIPv4:
> I'm not 100% if we should disable the redirects. Maybe its enough to make them temporary redirects?!

I would probably remove them? Or make them temporary to the plushies page? Seems odd to just silently go to the homepage.

apps/website/src/pages/index.tsx
```diff
@@ -320,17 +320,13 @@ const Home: NextPage = () => {
...
-              <Heading level={2}>Alveus Merch!</Heading>
+              <Heading level={2}>Alveus Plushies!</Heading>
               <p className="my-4">
-                Grab yourself some high-quality merch from Alveus in our
-                limited-time exclusive drops!
-              </p>
-              <p className="my-4">
-                Or, have your very own ambassador in your home with our
+                Grab yourself your very own ambassador in your home with our
                 plushies!
```
MattIPv4, line 330:
```suggestion
                Grab yourself a high-quality plushie and have your very own ambassador in your home!
```

pjeweb:
> > I'm not 100% if we should disable the redirects. Maybe its enough to make them temporary redirects?!
> 
> I would probably remove them? Or make them temporary to the plushies page? Seems odd to just silently go to the homepage.

sure, our 404 is just very barren. redirecting to youtooz seems confusing too. 🤔 can you redirect to the plush section on the home page `/#plushies`?

MattIPv4:
> can you redirect to the plush section on the home page `/#plushies`?

That seems sensible -- do we have that anchor, or does it need to be added?

pjeweb:
> > can you redirect to the plush section on the home page `/#plushies`?
> 
> That seems sensible -- do we have that anchor, or does it need to be added?

it needs added, but it looks like we can't redirect to an anchor from the next config. probably would just remove them then completely and maybe add something to the 404?

pjeweb:
> > > can you redirect to the plush section on the home page `/#plushies`?
> > 
> > 
> > That seems sensible -- do we have that anchor, or does it need to be added?
> 
> it needs added, but it looks like we can't redirect to an anchor from the next config. probably would just remove them then completely and maybe add something to the 404?

no, it was just slow next dev server. sorry. it works!

MattIPv4 (approved):

PR #416, opened 2023-10-02
Do not default Form.config field to empty string

kian:
## Describe your changes

The Form model defines the config field with an empty string as the default value which MySQL rejects.  Remove this default.

see:
```
rpc error: code = InvalidArgument desc = BLOB, TEXT, GEOMETRY or JSON column 'config' can't have a default value (errno 1101) (sqlstate 42000) 
```

full error:
```
alveusgg/apps/website on  main via  v18.15.0 took 31s
❯ npx prisma db push
Environment variables loaded from .env
Prisma schema loaded from prisma/schema.prisma
Datasource "db": MySQL database "alveus-dev" at "aws.connect.psdb.cloud"
Error: target: alveus-dev.-.primary: vttablet: rpc error: code = InvalidArgument desc = BLOB, TEXT, GEOMETRY or JSON column 'config' can't have a default value (errno 1101) (sqlstate 42000) (CallerID: sm24f0g6ixvc8xm6vgzx): Sql: "create table Form (\n\tid VARCHAR(191) not null,\n\tlabel VARCHAR(191) not null,\n\tslug VARCHAR(191) null,\n\tactive BOOLEAN not null default false,\n\tstartAt DATETIME(3) not null default current_timestamp(3),\n\tendAt DATETIME(3) null,\n\toutgoingWebhookUrl VARCHAR(720) null,\n\tshowInLists BOOLEAN not null default true,\n\tconfig TEXT not null default '',\n\tcreatedAt DATETIME(3) not null default current_timestamp(3),\n\tupdatedAt DATETIME(3) not null,\n\tUNIQUE INDEX Form_slug_key (slug),\n\tPRIMARY KEY (id)\n) charset utf8mb4,\n  COLLATE utf8mb4_unicode_ci", BindVars: {REDACTED}
   0: sql_schema_connector::apply_migration::migration_step
           with step=CreateTable { table_id: TableId(15) }
             at schema-engine/connectors/sql-schema-connector/src/apply_migration.rs:21
   1: sql_schema_connector::apply_migration::apply_migration
             at schema-engine/connectors/sql-schema-connector/src/apply_migration.rs:10
   2: schema_core::state::SchemaPush
             at schema-engine/core/src/state.rs:436
```

 - to repro follow local development setup doc from scratch
 - after removing default I was able to push schema and start the dev server, browse the site, log in via twitch, etc.


apps/website/prisma/schema.prisma
```diff
@@ -237,7 +237,7 @@ model Form {
   endAt              DateTime?
   outgoingWebhookUrl String?   @db.VarChar(720)
   showInLists        Boolean   @default(true)
-  config             String    @default("") @db.Text
+  config             String    @db.Text
```
kian, line 240: wasn't sure if the correct definition would be `String?` to indicate it is optional

pjeweb:
@kian just wanted to check in say i appreciate your PR but i have not found the time to check it out yet. I'll try to get to it on the weekend.

apps/website/prisma/schema.prisma
pjeweb: No, having it not optional should be the right thing. This way the config is required to be set on create.

pjeweb (approved):
Thank you! I did not really find out when this broke but i could reproduce the error and your PR fixes it.

PR #420, opened 2023-10-03
Update 2021 Annual Report

MattIPv4:
## Describe your changes

Replaces the PNG with a proper SVG, using minified raster images where needed to keep the file size sensible. Also, adds accessible text, similar to the 2022 report in #419.

Accessible text matches the content of the report image.


pjeweb (approved):
Drake

PR #422, opened 2023-10-06
Support `t` query param on AQ episodes

### LINKED ISSUE #400: [Idea]: Support `t` query param on Animal Quest episodes

Allow the `t` query param to be set on Animal Quest pages, and pass it down to the Twitch embed. Bonus points for parsing it and applying a minimum, so it cannot be set before the initial timestamp the data provides.

Allowing folks to deep-link to specific points within the episode, while using the Animal Quest page rather than directly linking to the Twitch VoD.

MattIPv4:
## Describe your changes

Resolves #400 by adding logic to parse the `t` query param on Animal Quest pages and use it as the start position for the Twitch embed, only if it is larger than the default start time for the VoD.

- When `t` is larger than the default start time, it is used.
- When `t` is smaller than the default start time, it is not used.
- When `t` is a partial time string (e.g. `10m00s` instead of `00h10m00s`), it is still used.

### DIFF
--- a/apps/website/src/pages/animal-quest/[episodeName].tsx
+++ b/apps/website/src/pages/animal-quest/[episodeName].tsx
@@ -1,5 +1,6 @@
 import type { GetStaticPaths, GetStaticProps, NextPage } from "next";
 import Image from "next/image";
+import { useRouter } from "next/router";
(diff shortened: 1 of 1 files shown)


pjeweb (approved):
amazing

PR #428, opened 2023-10-08
Open video for collaboration deep-links

MattIPv4:
## Describe your changes

To improve visibility and user experience, open the video popup by default when a deep-link anchor is used for a specific collaboration.

Valid deep-links open the popup reliably. Invalid deep-links do not.

### DIFF
--- a/apps/website/src/components/content/YouTube.tsx
+++ b/apps/website/src/components/content/YouTube.tsx
@@ -211,9 +211,39 @@ export const Lightbox = ({ id, className, children }: LightboxProps) => {
       }
     });
 
+    // Add the open attribute to the current slide
+    let lastElm: HTMLElement;
+    lightbox.on("change", () => {
+      // If there was a previously open slide, remove the attribute
+      if (lastElm) lastElm.removeAttribute("data-lightbox-open");
+
+      // If there is a current slide (there always should be), add the attribute
+      const elm = lightbox.pswp?.currSlide?.data?.element;
+      if (elm) {
+        elm.setAttribute("data-lightbox-open", "true");
+        lastElm = elm;
+      }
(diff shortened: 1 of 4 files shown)


MattIPv4:
I'm not very happy with this approach currently, simulating clicks. Trying to figure out what methods are exposed for directly triggering lightbox.

apps/website/src/components/content/YouTube.tsx
```diff
@@ -56,19 +58,35 @@ type TriggerProps = {
...
+    ({ videoId, caption, className, children }, ref) => {
+      // Expose a method to open the lightbox as the ref
+      const elm = useRef<HTMLAnchorElement>(null);
+      const open = useCallback(() => {
+        if (elm.current) elm.current.dispatchEvent(new Event("lightbox"));
+      }, []);
+      useEffect(() => {
+        if (typeof ref === "function") ref(open);
+        else if (ref) ref.current = open;
+      }, [ref, open]);
```
MattIPv4, line 71: My only question is whether it makes sense to expose a custom method here, or whether I should just expose the element and let the consumer dispatch the event?

pjeweb:
I am not sure what i am missing, but would it not be easier to open the lightbox using the method provided by photoswipe? Thats what we are using for show and tell also. I drafted a quick alternative PR #429 

I'm not sure about the refs approach. I think having a method in refs is not as common and so its harder for people to get maybe, but the encapsulation does make some sense. Not sure. Maybe it would feel nice if you package the event trigger thingy in a separate hook and give it the element ref instead. Hope that makes sense. Hard to grok react if it gets this entangled.

pjeweb (approved):
Great PR 👏

PR #429, opened 2023-10-09
Switch to controlled prop for lightbox open

pjeweb:
## Describe your changes

...

...

### DIFF
--- a/apps/website/src/components/content/YouTube.tsx
+++ b/apps/website/src/components/content/YouTube.tsx
@@ -5,12 +5,14 @@ import {
   useId,
   useMemo,
   useState,
-  useRef,
   type ReactNode,
 } from "react";
 import PhotoSwipeLightbox from "photoswipe/lightbox";
 
-import { getDefaultPhotoswipeLightboxOptions } from "@/utils/photoswipe";
+import {
+  getDefaultPhotoswipeLightboxOptions,
+  resolvePhotoswipeElementProvider,
+} from "@/utils/photoswipe";
 import { camelToKebab } from "@/utils/string-case";
 import { createImageUrl } from "@/utils/image";
 import { classes } from "@/utils/classes";
@@ -52,18 +54,31 @@ export const parseUrl = (url: string) => {
 type TriggerProps = {
   videoId: string;
   caption?: string;
+  triggerId?: string;
   className?: string;
   children: ReactNode;
 };
 
 const createTrigger = (id: string) => {
-  const Trigger = ({ videoId, caption, className, children }: TriggerProps) => (
+  const Trigger = ({
+    videoId,
+    caption,
+    triggerId,
+    className,
+    children,
+  }: TriggerProps) => (
     <a
       href={`https://www.youtube.com/watch?v=${encodeURIComponent(videoId)}`}
       target="_blank"
       rel="noreferrer"
       className={classes("group/trigger", className)}
-      {...{ [`data-lightbox-${id}`]: JSON.stringify({ videoId, caption }) }}
+      {...{
+        [`data-lightbox-${id}`]: JSON.stringify({
+          videoId,
+          caption,
+          triggerId: triggerId ?? videoId,
+        }),
+      }}
     >
       {children}
     </a>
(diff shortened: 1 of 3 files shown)


apps/website/src/components/content/YouTube.tsx
```diff
@@ -58,34 +57,23 @@ type TriggerProps = {
...
-          {...{ [`data-lightbox-${id}`]: JSON.stringify({ videoId, caption }) }}
-        >
-          {children}
-        </a>
-      );
-    },
-  );
+  const Trigger = ({ videoId, caption, className, children }: TriggerProps) => {
+    // Expose a method to open the lightbox as the ref
+    const elm = useRef<HTMLAnchorElement>(null);
```
pjeweb, line 62: forgot to remove this

apps/website/src/components/content/YouTube.tsx
```diff
@@ -143,33 +131,25 @@ type LightboxCtxProps = {
 
 type LightboxProps = {
   id?: string;
-  onInit?: () => void;
-  onDestroy?: () => void;
   className?: string;
   children: ReactNode | ((ctx: LightboxCtxProps) => ReactNode);
+  defaultOpen?: number;
```
pjeweb, line 136: maybe there is another name. its not a controlled value so i tried to use the defaultX naming convention like its common for html elements?!

MattIPv4:
Perhaps we update `Trigger` to require a unique key, and then can have the prop on `Lightbox` take that key and handle doing the DOM lookup for the index itself, so the consumer doesn't have to worry about how many `Trigger`s they're using, and in what order they're being rendered to the DOM etc.?

apps/website/src/components/content/YouTube.tsx
MattIPv4: I very much am not a fan of non-stateful props like this -- if we wanted to drive this by a prop, I'd probably want an `open` prop as well as an `onChange`, and expect them to always be in sync. If we just want something we use on demand, it should be a method not a prop imo.

pjeweb: sure. that should be quite possible to make it synced/controlled

apps/website/src/components/content/YouTube.tsx
```diff
@@ -244,44 +219,22 @@ export const Lightbox = ({
...
+  // Open the lightbox on load if we have a default
+  useEffect(() => {
+    if (defaultOpen && defaultOpen > -1) {
+      const gallery = document.getElementById(
+        photoswipeId,
+      ) as HTMLElement | null;
+      if (!gallery) return;
+      lightboxRef.current?.loadAndOpen(defaultOpen, { gallery });
+    }
+  }, [defaultOpen, photoswipeId]);
```
MattIPv4, line 237: We probably want to wait until the lightbox is initialized before doing this, and only handle if the useEffect for the lightbox gets re-run? The ref is probably fine, but we probably also still want a state to track once its initialized?

pjeweb:
> Perhaps we update `Trigger` to require a unique key, and then can have the prop on `Lightbox` take that key and handle doing the DOM lookup for the index itself, so the consumer doesn't have to worry about how many `Trigger`s they're using, and in what order they're being rendered to the DOM etc.?

Yeah that would probably make it more robust. The trigger could get the container/parent selector in `createTrigger` and all the trigger elements would need some attribute to find them. Makes sense.

apps/website/src/components/content/YouTube.tsx
pjeweb: I would've assumed the effects are done in series and they the init is synchronous, so it should be mostly guaranteed to be init'ed at that point?

MattIPv4: I would think so, but given they're different hooks with different dependencies, I would prefer to make this robust?

pjeweb: sure, also i dont think your approach is bad. with this one the arrow buttons in the lightbox does not seem to work. its probably not as easy as i thought ;)

MattIPv4:
@pjeweb pushed a commit here that I think does what we discussed above. Feel free to hack around or w/e, lmk if you're happy to merge this into the other branch

pjeweb:
> @pjeweb pushed a commit here that I think does what we discussed above. Feel free to hack around or w/e, lmk if you're happy to merge this into the other branch

I would be happy to merge this! Looks good and works very well.

PR #433, opened 2023-10-14
Add Filian collab

MattIPv4:
## Describe your changes

Adds the most recent collaboration, Filian

Video is correct, date is correct, name/link are correct

### DIFF
--- a/apps/website/src/components/content/YouTube.tsx
+++ b/apps/website/src/components/content/YouTube.tsx
@@ -30,7 +30,8 @@ const iframeAttrs = {
   title: "YouTube video",
   referrerpolicy: "no-referrer",
   allow: "encrypted-media",
-  sandbox: "allow-same-origin allow-scripts",
+  sandbox:
+    "allow-same-origin allow-scripts allow-popups allow-popups-to-escape-sandbox",
   loading: "lazy",
   class: "aspect-video w-full rounded-2xl shadow-xl bg-alveus-green-800",
 };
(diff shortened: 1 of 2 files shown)


pjeweb:
No way we can get all the content id music muted. I don't think we can embed this collab ...

MattIPv4:
Hm, yeah. How do you think we should handle it? Should we essentially just leave it as-is and allow the YouTube frame to open the 'Watch on YouTube' link, or should we add a prop to actively note this can't be embedded? If the prop, what should it do?

pjeweb:
I don't think the link "Watch on YT" actually works? I think the event get caught somewhere.

> If the prop, what should it do?

We could add our own "Video can't be embedded, watch on youtube.com/v/xxx" thingy, but that sounds like a lot of work for pretty much no benefit. Other idea would be to disable the lightbox trigger and just open it directly on youtube. Not great UX, but i might prefer that and it should be fairly easy?!

MattIPv4:
> I don't think the link "Watch on YT" actually works? I think the event get caught somewhere.

I just need to adjust the sandbox for the iframe and it'll work.

> Other idea would be to disable the lightbox trigger and just open it directly on youtube. Not great UX, but i might prefer that and it should be fairly easy?!

Yeah, that'd also work. My only hesitation is it might be confusing to the user that some show in a modal and some don't. I'm leaning towards the current behaviour, but with the sandbox updated so the link works?

pjeweb (dismissed):

pjeweb (approved):

PR #434, opened 2023-10-14
Add audio interfaces to network diagram

Echoskope:
Added the microphone interfaces.


MattIPv4:
I think the "w/ Axis P8221 I/O Audio Module" strings are probably outdated now, or are there two I/O modules for some cams?

Echoskope:
Good point, I removed that text and clarified the Microphone block from the Audio I/O block.

MattIPv4:
👀 One other thing, it'd be great to be consistent with either "AXIS" or "Axis" -- we have a mix of both currently?

Echoskope:
The device description on the Axis website has as AXIS, should we keep that same style anytime we use the brand name?

MattIPv4:
Yeah, all the products seem to use "AXIS", so that seems fine to use here

apps/website/src/data/network.ts
```diff
@@ -23,7 +23,7 @@ type NetworkItemCore = {
 };
 
 type NetworkItemSwitch = NetworkItemCore & {
-  type: "switch" | "converter";
+  type: "switch" | "converter"  | "interface";
```
MattIPv4, line 26:
```suggestion
  type: "switch" | "converter" | "interface";
```

MattIPv4 (approved):
LGTM, ty

PR #435, opened 2023-10-14
Enforce prettier in GitHub Actions

MattIPv4:
## Describe your changes

When talking with @Echoskope about preparing #434, it was noticed that the file had a missing `,` that prettier should've enforced. I then realised we don't actually enforce prettier anywhere in CI, only in commit hooks locally if you have them enabled.

This PR adds a GitHub Actions check for PRs to ensure they're following prettier conventions, and fixes a few outstanding violations.

`pnpm run prettier` and `pnpm run prettier --check` work as expected locally. GitHub Actions check works in PR.

### DIFF
--- a/.github/workflows/pull-request.yml
+++ b/.github/workflows/pull-request.yml
@@ -1,7 +1,47 @@
-name: Test
+name: Checks
 on: [pull_request]
+
 jobs:
-  test:
+  prettier:
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+
+      - name: Use Node.js (.nvmrc)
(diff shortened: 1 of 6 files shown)


MattIPv4:
This'll either need an admin merge and then branch protection updated once in main, or we can update branch protection first and then merge this immediately after.

pjeweb (approved):

pjeweb:
LookUp i think gh is confused about the test tasks

pjeweb:
> This'll either need an admin merge and then branch protection updated once in main, or we can update branch protection first and then merge this immediately after.

Oh i missed this comment. Yeah i don't care either way. Just do what needs to be done ;)

PR #440, opened 2023-10-20
Fix homepage performance in Firefox (remove blur placeholder)

MattIPv4:
## Describe your changes

In Firefox specifically, the homepage seems to cause a large lag spike on initial load, preventing the page from painting until it is loaded. In Chrome, this works perfectly with no issues.

After some trial-and-error removal of suspect parts of how the homepage slideshow, it would appear the source of the lag in Firefox was the base64 svg/png placeholder background images being set on the images prior to them loading fully. Removing these seems to completely resolve the lag, though does leave the page with no image until the first slide loads (which seems to be relatively fast anecdotally).

When compared to production, no lag is seen in Firefox when loading the homepage.

### DIFF
--- a/apps/website/src/components/content/Slideshow.tsx
+++ b/apps/website/src/components/content/Slideshow.tsx
@@ -137,7 +137,6 @@ const Slideshow = ({
             quality="50"
             loader={qualityLoader}
             priority={idx === 0}
-            placeholder="blur"
             className="h-full w-full object-cover"
             style={{
               animationDelay: `${idx * animation.duration.offset}ms`,


MattIPv4:
@pjeweb Can you test and see if the latest deployment here exhibits the lag for you? It seems to be gone for me, so the answer would appear that Firefox does not like decoding large base64 images...

MattIPv4:
I've filed https://github.com/webcompat/web-bugs/issues/128656 with the above. Hopefully someone more knowledgeable of Firefox/Gecko internals can create a minimal reproduction and maybe file a Bugzilla report for it.

pjeweb (approved):
Yeah its so much better. Could not reproduce the behaviour anymore!

pjeweb:
> I've filed [webcompat/web-bugs#128656](https://github.com/webcompat/web-bugs/issues/128656) with the above. Hopefully someone more knowledgeable of Firefox/Gecko internals can create a minimal reproduction and maybe file a Bugzilla report for it.

I did not know about that project. Interesting. Looks a lot easier to use than bugzilla.

PR #466, opened 2023-11-28
Add various indices to improve db performance

pjeweb:
## Describe your changes

Went through the request stats and added indices where queries were obviously unoptimized, reading almost the full table etc.

Deploy to your database, check it still runs. With planetscale you can watch the graphs get "better", otherwise you'd have to set up something else to get stats.

### DIFF
--- a/apps/website/prisma/schema.prisma
+++ b/apps/website/prisma/schema.prisma
@@ -30,6 +30,10 @@ model TaskExecutionEvent {
   task       String
   startedAt  DateTime  @default(now())
   finishedAt DateTime?
+
+  @@index([task], name: "TaskExecutionEvent_task")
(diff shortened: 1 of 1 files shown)


pjeweb:
@MattIPv4 can you review this (no rush)? i know its hard with our setup to test with on a local database. but it did bring down rows read on the live db by a lot. this would just bring the schema up to date with the live db.

MattIPv4 (approved):
I'm surprised that indexing on dates works, but these all seem like logic changes in the diff, not that worried about testing locally if this is what's already been tested in prod

pjeweb:
Afaik they work pretty decently with btrees for range queries and sorting and such. But my database knowledge is pretty superficial.

PR #468, opened 2023-11-29
Re-add shop for Christmas apparel

MattIPv4:
## Describe your changes

Adds the shop back to the navigation, re-enables the redirects to the shop, and adds a top-hat promotional message for the new items.

Links all work, wording is correct.

### DIFF
--- a/apps/website/src/data/promo.ts
+++ b/apps/website/src/data/promo.ts
@@ -1,0 +1,20 @@
+interface Promo {
+  title: string;
+  subtitle: string;
+  content: string;
+  link: string;
+  external: boolean;
+  excluded: string[];
+}
+
+const promo: Promo | null = {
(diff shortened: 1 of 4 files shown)


apps/website/src/data/promo.ts
```diff
@@ -0,0 +1,20 @@
...
+  link: string;
+  external: boolean;
+  excluded: string[];
+}
+
+const promo: Promo | null = {
+  title: "Christmas Sweaters & 2024 Calendars",
+  subtitle: "Available from December 1st",
+  content:
+    "Grab yourself an embroidered Christmas sweater and a 2024 calendar featuring a selection of our ambassadors, shipping worldwide and in time for Christmas.",
```
MattIPv4, line 14: I'm not actually using this for now, but could be used to add a section to the homepage or similar down the road

pjeweb (approved):

PR #471, opened 2023-12-09
Add documentation for chat commands

MattIPv4:
## Describe your changes

Adds a page under the about/tech section that contains documentation for all known chat commands available to moderators and trusted chat members.

Commands are correct, documentation is clear and easy to understand.


MattIPv4:
Going to get this in w/ just the commands for now, and do a follow-up PR for the presets once that's ready to be documented.

pjeweb (changes requested):
The non-breaking code elements breaks the layout on mobile. Maybe we can add overflow-auto containers around the command list so you can scroll vertically on mobile?

pjeweb (approved):

PR #473, opened 2023-12-10
Update to Node.js 20

MattIPv4:
## Describe your changes

Updates the whole stack to Node.js 20. Ideally, we'd got to 20.10.0, but it seems Vercel only supports up to 20.5.1, so I've pinned everything to that for consistency.

Everything continues to work.


MattIPv4:
Will update branch protection when ready to merge

pjeweb (approved):
looks to be working well with v20 everywhere

PR #475, opened 2023-12-14
Add documentation for PTZ command presets

MattIPv4:
## Describe your changes

Adds documentation for all the PTZ presets, as part of the command documentation page.

All presets are under the correct cameras, commands are correct, descriptions make sense.


pjeweb:
More ideas: Copy command on click to clipboard, tooltip for detailed explanation (but what would we want to explain?)

pjeweb:
Oh this is the original list by logan https://github.com/loganrx8/alveuspresets/blob/main/descriptions

pjeweb (dismissed):

pjeweb (approved):

PR #481, opened 2023-12-20
Create Grouped component for use with useGrouped hook

MattIPv4:
## Describe your changes

Further to #467, creates a component to be used alongside the hook to handle rendering the items when they're either an array or map, and handling the logic for scrolling to the correct section.

Anchor deep-linking still works as expected on the ambassador page.

### DIFF
--- a/apps/website/src/components/content/Grouped.tsx
+++ b/apps/website/src/components/content/Grouped.tsx
@@ -1,0 +1,78 @@
+import { useCallback, type ComponentType, type RefAttributes } from "react";
+
+import { type Items } from "@/hooks/grouped";
+
+export interface GroupedProps<T> {
+  items: T[];
+  option: string;
+  group: string | null;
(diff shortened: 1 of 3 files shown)


MattIPv4:
Open to other ideas for the interface here, not particularly settled on this.

I had also considered adding the dropdown itself as part of this component, but wasn't sure. What do you think?

pjeweb (approved):
I like the way its split up. Makes sense to me to have the Dropdown and List separate.

Everything seems to work and the code looks fine.

PS: I'm gonna make a separate PR to fix a lint warning in the hook that was introduced by using the router.

pjeweb (approved):
I like the way its split up. Makes sense to me to have the Dropdown and List separate.

Everything seems to work and the code looks fine.

PS: I'm gonna make a separate PR to fix a lint warning in the hook that was introduced by using the router.

PR #483, opened 2023-12-21
Bump Node.js to 20.9.0 to match Vercel

pjeweb:
## Describe your changes

Update Node.js to 20.9.0 to match what is running on Vercel.

Check everything still works and I did not forget an instance of the version number.


pjeweb:
Do we maybe want a script to set the version or is search-and-replace good enough for now?

MattIPv4:
For `@types/node`, it looks like `20.9.5` is the latest version

MattIPv4:
I've generally been fine w/ doing search/replace, and checking the latest types version, but also down to have a tiny script

MattIPv4 (approved):

PR #489, opened 2023-12-22
Add OBS servers to network diagram

MattIPv4:
## Describe your changes

Per discussions with a few tech folks in the Alveus community, we think it makes sense to include the two OBS servers that handle the streaming in the network topology.

Type changes make sense (my head hurts).

### DIFF
--- a/apps/website/src/data/network.ts
+++ b/apps/website/src/data/network.ts
@@ -9,7 +9,7 @@ type NetworkConnectionWired = NetworkConnectionCore & {
 };
 
 type NetworkConnectionWireless = NetworkConnectionCore & {
-  type: "wifi";
+  type: "wifi" | "cloud";
 };
 
 type NetworkConnection = NetworkConnectionWired | NetworkConnectionWireless;
@@ -19,12 +19,11 @@ type NetworkItemCore = {
   name: string;
   model: string;
   url?: string;
-  connection: NetworkConnection;
 };
 
 type NetworkItemSwitch = NetworkItemCore & {
   type: "switch" | "converter" | "interface";
-  links: NetworkItem[];
+  links: NestedNetworkItem[];
 };
 
 type NetworkItemCamera = NetworkItemCore & {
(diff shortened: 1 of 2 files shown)


apps/website/src/data/network.ts
```diff
@@ -33,442 +32,436 @@ type NetworkItemCamera = NetworkItemCore & {
...
+    typeof item.model === "string"
+  );
+};
+
+export const isRootNetworkItem = (item: unknown): item is RootNetworkItem => {
+  return (
+    isNetworkItem(item) &&
+    (!("connection" in item) || typeof item.connection === "undefined")
+  );
+};
```
pjeweb, line 71: this is not used as far as i can see. do you wanna keep it for future purposes?

apps/website/src/data/network.ts
```diff
@@ -33,442 +32,436 @@ type NetworkItemCamera = NetworkItemCore & {
...
+
+export const isNestedNetworkItem = (
+  item: unknown,
+): item is NestedNetworkItem => {
+  return (
+    isNetworkItem(item) &&
+    "connection" in item &&
+    typeof item.connection === "object" &&
+    typeof item.connection.type === "string"
```
pjeweb, line 80: `typeof item.connection.type === "string"` should always be true

apps/website/src/data/network.ts
MattIPv4: Yeah indeed

apps/website/src/data/network.ts
MattIPv4: I'm not sure I follow? The value is unknown, this is a type guard, so we need to check?

pjeweb: Checking `"connection" in item` should be sufficient as a check though. I dont think you need to test the type of the type property. At least my typescript is happy to narrow the type without it.
```typescript
if (!("connection" in targetNode.data.item))
  throw new Error("Invalid target");
const targetConnection = targetNode.data.item.connection;
```
is fine. You could simplify the guard to `export const isNestedNetworkItem = (item: NetworkItem): item is NestedNetworkItem => "connection" in item;` even, since the `targetNode.data.item` is already known to be a `NetworkItem`.

But its fine with the extra checks too.

pjeweb (dismissed):

apps/website/src/data/network.ts
MattIPv4: Ahh okay yeah, so really the `isNetworkItem` check should have a "if connection does exist, check type is string within it". I'm just technically doing the check at the wrong point.

pjeweb: I mean the first check narrows the type down to a nested node and that guarantees the type to be a string already since the type NetworkConnection always has a type (with type string). So unless you add another type of node also with a connection property, checking for the string should not be necessary.

pjeweb (approved):

PR #500, opened 2024-01-14
Add YouTube to homepage watch live button as dropdown

### LINKED ISSUE #497: [Idea]: Link to the YT live stream (and chat)

This probably needs access to the YT API to get the livestreams. Would be nice in the frontend and as short links (`alveus.gg/yt-live` and `alveus.gg/yt-chat`)

Improve discovery of YT stream and chat

MattIPv4:
## Describe your changes

Resolves #497

Closes #503

Adds a new shortlink for the YT livestream (`/live/youtube`), and a matching link for Twitch (`/live/twitch`).

Adds a new dropdown the homepage `Watch Live` button, allowing the user to pick YouTube instead of Twitch if they wish. The button is still a link to Twitch by default, and still only requires a single click if the user does not click on `Twitch` specifically.

Shortlinks both work as expected. Watch Live button is intuitive to interact with, and still allows for single-click navigation.

### DIFF
--- a/apps/website/next.config.mjs
+++ b/apps/website/next.config.mjs
@@ -300,7 +300,17 @@ const config = {
     {
       source: "/live",
       destination: "https://twitch.tv/alveussanctuary",
-      permanent: false,
+      permanent: true,
+    },
+    {
+      source: "/live/twitch",
+      destination: "https://twitch.tv/alveussanctuary",
+      permanent: true,
+    },
+    {
+      source: "/live/youtube",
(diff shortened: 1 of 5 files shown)


pjeweb (commented):
Love the button!

A few issues:
- react complains about the nested <a> from the menu. i think it would make sense to have the link and menu button as siblings
- i'd prefer the menu button to take up the whole height and all the space on the right

See #503

pjeweb (approved):

PR #506, opened 2024-01-17
Bypass consent for crawlers on AQ pages

MattIPv4:
## Describe your changes

This is a bit of an experiment, with the aim hopefully for Google to index the AQ episode pages as video pages.

Google currently sees them as _potential_ video pages, but only due to the OG meta tags, not due to the video content actually on the page (it can't see that as it is behind content).

image

As such, this PR adds a new flag to the consent component, so that we can opt specific usages of it into bypassing the consent screen when we want it to be indexed by known crawlers.

Using a tool to fake your user-agent as Googlebot (https://developers.google.com/search/docs/crawling-indexing/overview-google-crawlers), with consent set to deny all, the Twitch embed on the AQ episode pages should still load.

### DIFF
--- a/apps/website/src/components/Consent.tsx
+++ b/apps/website/src/components/Consent.tsx
@@ -1,4 +1,9 @@
-import { useCallback, type MouseEventHandler, type ReactNode } from "react";
+import {
+  useCallback,
+  type MouseEventHandler,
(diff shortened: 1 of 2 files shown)


pjeweb (approved):
PauseChamp google

Issue #411, opened 2023-09-25
[Idea]: Searchable clips (with user voting)


pjeweb:
### Describe your idea / enhancement

Implement a system for logged in users to submit and vote clips from the alveus twitch channel.

Use cases MVP:

- Everyone:
  - View clips [new, top all time, top 24h, top 7d, top 30d, top year]
  - View a clip (twitch embed, votes, submitter, creator, timestamp/relative time)
- Logged in:
  - Submit a clip
	  - Check its a valid clip and broadcaster_id is alveus or maya?! => https://dev.twitch.tv/docs/api/reference/#get-clips
	  - Only allow submitting a clip once
	  - Save to database
  - Upvote a clip (only upvote?) -> save to database (and maybe kv)
- Admin:
  - Maybe allow deleting submitted clips (without deleting the twitch clip)
- Background tasks:
  - Check if a clip has been deleted (twitch webhook and/or cronjob)

Use cases optional:

- "Hot" clips algorithm
- Discover mode (show random clips) (encourage voting, but do not show vote count before)
- Leaderboard / Awards for top submitters
- Credit clippers/submitters with points (see #24)
  - Problem: Fighting over internet points (submitter vs. clipper) -> clipper should probably get the credit
- Bookmark clips for your own use
- Search clips by title
- Search clips by tags
- Search clips by transcript
- Allow a list of source vods from other twitch channels for clips (for collabs)
- Allow YT clips
- Allow YT shorts
- Allow IG reels
- Allow TikTok videos
- Make clips sharable?! (deep links)

Prior art:

- https://smashscene.com/alveus
- https://github.com/JakeMiki/twitch-clip-queue

User feedback:

- On mobile its hard to get you own clips you've made (http://clips.twitch.tv/my-clips)

### How would this idea / enhancement be beneficial?

Helps discover clips for visitors and media production, encourages clipping, allows crediting clippers with the credit system when implemented.

-----------------------

### Discussed in https://github.com/orgs/alveusgg/discussions/305

<div type='discussions-op-text'>

<sup>Originally posted by **pjeweb** December 14, 2022</sup>
</div>

ExperimentalCyborg:
+ hide votes on a clip for the first X hours after submitting
+ log referrer HTTP header value (if any was set) for every page visit and for every vote (to be able to detect brigading and potentially roll back votes)

pjeweb:
tagging could be used to fill the video/clips section on the ambassador pages

[...]

PR #516, opened 2024-01-29
add user-submitted clips page

trimbe:
## Describe your changes

#411

Adds clips page for users to submit and vote on clips from the alveus or maya channels. Clips can be sorted by top/top 7 days/top 30 days/new. Clips require approval before they can be viewed publicly and an admin panel is added to facilitate that. A daily cron is added to verify validity of submitted clips and remove them if the clip has been deleted. 

Some thoughts: 

I'm not at all attached to the design, I primarily was going for something functional and not ugly. Very open to suggestions to make it look better.

Voting currently requires the user to be logged in, but that might not be great for engagement. Possibly allow logged out voting, but add a unique constraint on clip slug/IP address for the ClipVotes to mitigate potential manipulation?

I like the idea of a "hot" sort, but it seems this would be less useful unless there's some threshold of voting traffic and I'm not sure how much traffic the website can be expected to receive. 

I like many of the other ideas in #411 like ambassador tagging, leaderboard/points of some sort, and deep share links, but wanted to submit this base version first. Happy to add on later if these other features are desirable. 

Deployed here: https://alveusgg-website.vercel.app/clips (don't think Vercel previews can handle the db changes?)


pjeweb:
@trimbe this looks amazing. i will look into the code as soon as i can. still figuring out how to apply the schema changes for the PR to a new db branch, so no preview deployment right now.

apps/website/prisma/schema.prisma
```diff
@@ -454,3 +456,31 @@ model VirtualTicket {
   @@unique([userId, eventId])
   @@index([eventId], name: "VirtualTicket_eventId")
 }
+
+model Clip {
+  id            String @id @default(cuid())
+  clipSlug      String @unique
```
pjeweb, line 462: Is there a specific reason we have use the clip prefix here?

apps/website/prisma/schema.prisma
```diff
@@ -454,3 +456,31 @@ model VirtualTicket {
...
+  userId        String?
+  title         String
+  thumbnail     String
+  createdAt     DateTime @default(now())
+  clipCreator   String?
```
pjeweb, line 467: dito

apps/website/prisma/schema.prisma
```diff
@@ -454,3 +456,31 @@ model VirtualTicket {
...
+  submitted     DateTime @default(now())
```
pjeweb, line 468:
```suggestion
  submittedAt     DateTime @default(now())
```
to be consistent

apps/website/prisma/schema.prisma
```diff
@@ -454,3 +456,31 @@ model VirtualTicket {
   @@unique([userId, eventId])
   @@index([eventId], name: "VirtualTicket_eventId")
 }
+
+model Clip {
+  id            String @id @default(cuid())
+  clipSlug      String @unique
+  userId        String?
+  title         String
+  thumbnail     String
```
pjeweb, line 465: Is this an id or url? Can we call it `thumbnailUrl` or `thumbnailId`?

apps/website/src/server/trpc/router/clips.ts
```diff
@@ -0,0 +1,142 @@
...
+  addClip: publicProcedure
+    .input(clipSubmitInputSchema)
+    .mutation(async ({ ctx, input }) => {
+      // https://clips.twitch.tv/slug?query
+      // https://www.twitch.tv/channel/clip/slug?query
+      // https://www.twitch.tv/clip/slug?query
+      // https://m.twitch.tv/channel/clip/slug?query
+
+      const clipRegex =
+        /^https?:.+(?:clips\.twitch\.tv|(?:m\.)?twitch\.tv\/(?:clip|.+\/clip))\/?([\w\d-]+)/;
```
pjeweb, line 26:
```suggestion
        /^https?:.+(?:clips\.twitch\.tv|(?:m\.)?twitch\.tv\/(?:clip|.+\/clip))\/?([\w-]+)/;
```

`\w` already includes `[0-9]`

apps/website/src/server/trpc/router/clips.ts
```diff
@@ -0,0 +1,142 @@
...
+      const clipDetails = details?.data?.[0];
+
+      if (!clipDetails) {
+        throw new TRPCError({
+          code: "BAD_REQUEST",
+          message: "Clip not found",
+        });
+      }
+
+      const channelAllowlist = ["alveussanctuary", "maya"];
```
pjeweb, line 54: Probably wanna move this to an env variable or config file

apps/website/src/server/trpc/router/clips.ts
```diff
@@ -0,0 +1,142 @@
...
+
+      if (
+        !channelAllowlist.includes(clipDetails.broadcaster_name.toLowerCase())
+      ) {
+        throw new TRPCError({
+          code: "BAD_REQUEST",
+          message: "Clip not from AlveusSanctuary or Maya",
```
pjeweb, line 61: Would keep the message generic or based on the allow list?

pjeweb (commented):
I did not have time to check out the code 100%, but I'm overall pretty impressed! Would appreciate feedback on the code base and how it was to work with too :)

Some initial feedback/ideas:
- Appreciate the MVP. I also think, its best we start small.
- Design overall is pretty good and clean. I only have small things I'd wanna change.
- I did not have that in #414 and I am sorry for that, but I was thinking it might be better to add _all_ created clips to the site. At least the ones from the Alveus channel. I think one problem with the smashscene is, that people usually only submit their own clips and not many do since its an additional step.
- In the same vein, I think it might be better to accept all clips without approval by default and only allow to remove approval.

trimbe:
> I did not have that in https://github.com/alveusgg/alveusgg/issues/411 and I am sorry for that, but I was thinking it might be better to add all created clips to the site. At least the ones from the Alveus channel. I think one problem with the smashscene is, that people usually only submit their own clips and not many do since its an additional step.

Yeah this sounds sensible, doesn't look like there's any events in EventSub so I'll add a cron for this.

> In the same vein, I think it might be better to accept all clips without approval by default and only allow to remove approval.

Alright, sounds good, I just wanted to err on the side of caution for moderation purposes but it's probably fine.

> Would appreciate feedback on the code base and how it was to work with too :)

This is my first time using React/next/prisma in a serious project, but from that perspective, I didn't really have any issues finding my way around or understanding the flow.

trimbe:
@pjeweb So I've written a cronjob to add all clips but now that I'm looking at some of the clips I'm not sure if it's the best idea. It appears that clips created on mobile are saved as soon as the clip button is pressed and given a default title of the stream's title at the time. A pretty significant proportion of clips are these default clips, and I'm not sure how often anything interesting happens in them (misclicks, or people just using them to rewind?). I feel like they just add clutter if you're trying to look through the new clips.

Filtering by views when added may be a good idea. A floor of 10 views reduces the total clips on AlveusSanctuary from 13400 to 7600 and I think that still catches the interesting stuff if people organically share with friends or chat. I'd add some buffer time when adding to allow time for the clips to accumulate views.

Another consideration is overlapping clips, where several users clip the same moment. I'm not sure this is really possible to avoid intelligently, but it also might be mitigated to some extent by the views requirement.

trimbe:
@pjeweb Clip view distribution: 

| Views     | Number |
|-----------|--------|
| 1-5       | 4,247  |
| 6-10      | 1,858  |
| 11-50     | 4,350 |
| 51-100    | 1,330  |
| 101-500   | 1,597  |
| 501-1000  | 223    |
| 1001-5000 | 130    |
| 5000+     | 37     |

I do like the chatbot idea but I worry that will exacerbate the overlapping clip issue. I think with a sane view requirement, we probably accomplish pretty much the same thing anyway. 50+ seems like it may be alright, and I'll stagger it by 10 minutes or so to allow for accumulation of views. And allow non-logged-in submissions.

For the initial import I can add a postinstall script, and increase the threshold to 500+ maybe.

trimbe:
Addressed your reviews and made changes as per my previous comment. Not fully confident on the 30m/50 view numbers, might be a bit high for slow days. None of the clips from today would have been added, for example, but maybe nothing warranted it.

pjeweb:
@trimbe thank you! might take me a few days to review though. sorry about that

pjeweb:
@trimbe would you mind me pushing a merge commit with the updates to the show and tell pages to your branch?

trimbe:
@pjeweb Nope, go ahead.

apps/website/src/pages/clips/submit-clip.tsx
```diff
@@ -0,0 +1,128 @@
+import type { NextPage } from "next";
+import Image from "next/image";
+import type { FormEvent } from "react";
+import { useState } from "react";
+import { useSession } from "next-auth/react";
```
pjeweb, line 5: `useSession` is not being used

apps/website/src/pages/clips/index.tsx
```diff
@@ -0,0 +1,330 @@
...
+    ? "hover:fill-alveus-green-900 fill-green"
+    : "hover:fill-green fill:alveus-green-900";
+
+  return (
+    <div>
+      <div
+        ref={lastItemRef}
+        className="group cursor-pointer"
+        onClick={() => clipClicked(clip)}
+      >
```
pjeweb, line 79: clickable elements should be actually `<button>` elements (or `<a>`)

apps/website/src/pages/clips/index.tsx
```diff
@@ -0,0 +1,330 @@
...
+      </div>
+      <div className="flex justify-between px-4">
+        <p className="text-left">Votes: {clip.voteCount}</p>
+        {session && (
+          <p
+            className={classes("cursor-pointer text-right", thumbsClass)}
+            onClick={() => handleVote(clip.id)}
+          >
+            <IconThumbsUp size={20} />
+          </p>
```
pjeweb, line 101: dito make it a `<button>`

apps/website/src/pages/clips/index.tsx
```diff
@@ -0,0 +1,330 @@
...
+}: {
+  clip: Clip;
+  onClose: (e: React.MouseEvent) => void;
+}) => {
+  const clipUrl = new URL("https://clips.twitch.tv/embed");
+  clipUrl.searchParams.set("clip", clip.slug);
+  clipUrl.searchParams.set("parent", window.location.hostname);
+
+  return (
+    <div
```
pjeweb, line 120: The modal dialog should capture keyboard focus and close when pressing the escape key and have appropriate aria attributes. The easiest would be to use the headless `Dialog` component (see https://github.com/alveusgg/alveusgg/blob/main/apps/website/src/components/shared/ModalDialog.tsx). This will require some tweaks to the event handler but it should be fairly easy to replace the divs with `Dialog`, `Dialog.Panel` and the headline with a `Dialog.Title`. Reusing the existing `ModalDialog` might also work but the button is fixed as a labeled button at the bottom of the dialog ...

apps/website/src/pages/clips/index.tsx
```diff
@@ -0,0 +1,330 @@
...
+          >
+            <path
+              onClick={onClose}
+              strokeLinecap="round"
+              strokeLinejoin="round"
+              strokeWidth="2"
+              d="M6 18L18 6M6 6l12 12"
+            />
+          </svg>
+        </span>
```
pjeweb, line 141: dito use a `<button>`. the events should bubble up, attaching the events to the svg and path should not be necessary

apps/website/src/pages/clips/index.tsx
```diff
@@ -0,0 +1,330 @@
...
+    <>
+      <Meta
+        title="Clips"
+        description="Explore and vote on Twitch clips from our community streams. Discover community-curated highlights and top moments."
+      />
+
+      {/* Nav background */}
+      <div className="-mt-40 hidden h-40 bg-alveus-green-900 lg:block" />
+
+      <div className="relative">
```
pjeweb, line 251: I think this div is not needed. This was for placing leaves i think?!

apps/website/src/pages/show-and-tell/my-posts/[postId]/index.tsx
```diff
@@ -6,16 +6,18 @@ import { trpc } from "@/utils/trpc";
...
 import Meta from "@/components/content/Meta";
 import Section from "@/components/content/Section";
 import Heading from "@/components/content/Heading";
 
-import { ShowAndTellNavigation } from "@/components/show-and-tell/ShowAndTellNavigation";
 import { ShowAndTellEntryForm } from "@/components/show-and-tell/ShowAndTellEntryForm";
 
 import showAndTellHeader from "@/assets/show-and-tell/header.png";
 
+import { showAndTellNavItems } from "../..";
```
Comment (pjeweb) [inline apps/website/src/pages/show-and-tell/my-posts/[postId]/index.tsx:19]: Note sure i like the relative path here. Maybe we can just use the full path or adding back the ShowAndTellNavigation component (basically just wrapping the `PageNavigation`) in its own file and co-locate the items there?

apps/website/src/pages/clips/submit-clip.tsx
```diff
@@ -0,0 +1,128 @@
...
+    {
+      href: "/clips/",
+      label: "Clips",
+      exact: true,
+    },
+    {
+      href: "/clips/submit-clip",
+      label: "Submit Clip",
+    },
+  ];
```
pjeweb, line 40: Instead of repeating the items maybe create a component wrapping the `PageNavigation` in its own file and co-locate the items array there? Exporting the items from one file and reusing them in the other can also work (see comment about the relative import for show and tell).

pjeweb (changes requested):
Overall I'm very happy with your work. I wanna get @MattIPv4's eyes on the UI/UX (esp. design and text) at some point too, but I think this should make it in the site soon :)

MattIPv4:
👋 Apologies @trimbe it looks like one of my recent merges has caused some conflicts here. Would you be able to get them resolved and get this branch up-to-date with main?

trimbe:
@MattIPv4 👍

pjeweb:
Sorry for the delays. Trying to find the time to do the final review.

pjeweb:
Just a small update: I have a bunch of small tweaks in a local branch that I need to finish. If you do not mind, I might push the changes in your PR directly and explain with comments? It will probably still take me some days to do that either way.

trimbe:
Yeah, that's fine, go for it.

pjeweb:
I am gonna close this PR for now. I think we have other people working on a standalone clips archive.

@trimbe Thank you a lot for your work and sorry we did not get it shipped.

PR #517, opened 2024-01-30
Update dependencies

pjeweb:
overall:
- apply new prettier rules, adding trailing commas to tsconfig.json files
- dedupe deps

root:
- update deps (mostly minor updates, except husky 19)
- cleanup husky commit hook

website:
- update deps (mostly minor updates or non-breaking major updates; e.g. next 14.1.0)
- remove date-fns, use luxon instead to cut one dep
- remove node-fetch, use native fetch instead
- replace abandoned next-superjson with next-superjson-plugin and update superjson to v2
- refactor virtual ticket code to not include prisma in edge bundles (throws an exception otherwise)

chatbot:
- update deps (minor)


MattIPv4:
Can we update the root dependencies too? Specifically for Husky with v9, we can clean up the pre-commit file (so it's just the `pnpm` command), and change `husky install` to just `husky`

pjeweb:
the last commit is currently deployed to preview.alveus.gg

MattIPv4 (approved):

PR #523, opened 2024-02-03
Add collab for JaidenAnimations & Alphara

pjeweb:
## Describe your changes

Adds the most recent collaboration, JaidenAnimations & Alpharad

Since each has their own separate channels/social media I did not link any, making the link nullable. Alternatively we could allow multiple links?!

Video is correct, date is correct, names are correct


MattIPv4 (approved):
Seems like a fine solution

PR #535, opened 2024-02-09
Add short links

### LINKED ISSUE #464: [Idea]: Admin/mod tool - Add short Links

Manage short links that redirect to an url thru the admin panel. Could be external sites or pages.

We probably still prefer next redirects if possible, but for quick things having a GUI would be nice.

Maybe also support giving uploaded images a short link (see #287)

Track clicks (counting/aggregating over time).

Better links to share (instead of long or changing external urls)

JonasFranke:
resolves #464


pjeweb:
Thank you for working on this!

One concern with the dynamic route and redirecting in the client is that its unnecessary delay for the redirect. The user should not have to load the full page and JS etc. just to get redirected. Imho it would be cleaner to do it in a [middleware](https://nextjs.org/docs/pages/building-your-application/routing/middleware) and redirect with HTTP (headers) instead. I am not 100% sure its possible to connect to the mysql database from there (it should be since its still a AWS lambda with the old pages router?!). So we maybe should use a key-value store or analytics http endpoint then (see below).

> @pjeweb How would you implement the link usage tracking feature? Just update an int in the database, or do we have better tools for this?

Did you remove the comment? Cant find it anymore.

I think a counter in the mysql database should be good enough™ for our purposes.

Since we are working with the serverless environment we kinda have to use third-party services otherwise. We could use the [Vercel KV](https://vercel.com/docs/storage/vercel-kv) Redis or [Upstash](https://upstash.com/docs/redis/overall/getstarted) Redis (key-value stores).

We also can use [Vercel Analytics Custom Events](https://vercel.com/docs/analytics/custom-events) (only available in payed Pro plan). That would make sense since we track page views there.

Since I never finished setting up the OSS analytics with Plausible, we cant use that _yet_ but maybe soon, if I finish that. Maybe its worth waiting for that?!

@MattIPv4 do you have opinions on this?

JonasFranke:
I really enjoyed working on it and I learned a lot!

After looking into middleware, I also think that it'd be much cleaner, and I'll look deeper into it tomorrow.

I removed the comment 'cause I thought that I was the best (and easiest) solution.
Imho it wouldn't be worth waiting for Plausible.

MattIPv4:
> > @pjeweb How would you implement the link usage tracking feature? Just update an int in the database, or do we have better tools for this?
> 
> Did you remove the comment? Cant find it anymore.
> 
> I think a counter in the mysql database should be good enough™ for our purposes.
> 
> Since we are working with the serverless environment we kinda have to use third-party services otherwise. We could use the [Vercel KV](https://vercel.com/docs/storage/vercel-kv) Redis or [Upstash](https://upstash.com/docs/redis/overall/getstarted) Redis (key-value stores).
> 
> We also can use [Vercel Analytics Custom Events](https://vercel.com/docs/analytics/custom-events) (only available in payed Pro plan). That would make sense since we track page views there.
> 
> Since I never finished setting up the OSS analytics with Plausible, we cant use that _yet_ but maybe soon, if I finish that. Maybe its worth waiting for that?!
> 
> @MattIPv4 do you have opinions on this?

I like the idea of using Vercel Analytics for it, but also not sure how easy that'd be to access outside of Vercel (say if we wanted to show the count in the admin UI). I think a simple click counter in the DB would be fine for now, just to give us a sense of usage? We could always do the DB counter + Vercel analytics?

JonasFranke:
@pjeweb Got it working with middleware!
I also changed the route to `/l/[slug]`

apps/website/src/middleware.ts
pjeweb, line 1: I might be overthinking or missing something, but im not so happy with the additional roundtrip and big db select on every request.

I take it, accessing the db directly in the middleware and selecting the one link (by slug) did not work since its running on edge runtime? Is it worth using the http api by planetscale (`@planetscale/database`) for this, like i did for the bingo and virtual tickets? https://github.com/alveusgg/alveusgg/blob/bd4fe533fb2419a1de866087e6c082700fc9c3d2/apps/website/src/app/api/bingo/%5Bbingo%5D/route.ts#L30-L42

Or maybe even using [prisma with the new adapter](https://github.com/prisma/prisma/discussions/21347)?

The api route with all links does not get cached, right? Next/vercel caching is so confusing with all the options HELP

- [server-side fetch data cache](https://nextjs.org/docs/app/building-your-application/data-fetching/fetching-caching-and-revalidating#fetching-data-on-the-server-with-fetch) does afaik not work in the middleware as its edge only?!
- the [`revalidate` route segment config](https://nextjs.org/docs/app/api-reference/file-conventions/route-segment-config#revalidate) afaik only applies to app router
- The [Vercel CDN](https://vercel.com/docs/edge-network/caching) probably does not cache the api response? Does the middleware go through the CDN?! If it does i think that could work very well. We'd only need to send some headers in the api route if i understand correctly

apps/website/src/pages/api/short-links/add.ts
```diff
@@ -0,0 +1,30 @@
...
+        where: { id: id },
+        data: {
+          clicks: clicks,
+        },
+      });
+      res.status(204).send("");
+    }
+  } else {
+    res.status(403).send("Unauthorized");
+  }
```
pjeweb, line 29: I think you can update the clicks in one atomic update with `increment`. See https://www.prisma.io/docs/orm/reference/prisma-client-reference#atomic-number-operations

apps/website/src/middleware.ts
```diff
@@ -0,0 +1,35 @@
...
+import { env } from "@/env/index.mjs";
+
+export async function middleware(req: NextRequest, event: NextFetchEvent) {
+  const res = await fetch(env.NEXT_PUBLIC_BASE_URL + "/api/short-links", {
+    method: "GET",
+  });
+  const shortLinks = await res.json();
+
+  for (const link of shortLinks) {
+    if (req.nextUrl.pathname.startsWith("/l/" + link.slug)) {
```
pjeweb, line 15: Does it make sense to use `startsWith` here? Should the slug not be a full match. If i understand correctly a slug ' twitch' would always win over `twitch-giveaway`, `twitchsub` etc. if it was added first?

I think next already normalizes trailing slashes, so that _should_ not be a problem.

apps/website/src/middleware.ts
```diff
@@ -0,0 +1,35 @@
...
+      return NextResponse.redirect(link.link);
+    }
+  }
+
+  // If no match is found, redirect the user to the home page
+  return NextResponse.redirect("/");
+}
+
+export const config = {
+  matcher: "/l/:path*",
```
pjeweb, line 34: The star allows slashes in the path right? Should be rather not allow slashes in slugs?

apps/website/src/components/admin/short-links/ShortLinkForm.tsx
```diff
@@ -0,0 +1,114 @@
...
+
+  return (
+    <form className="flex flex-col gap-10" onSubmit={handleSubmit}>
+      {submit.error && (
+        <MessageBox variant="failure">
+          <pre>{submit.error.message}</pre>
+        </MessageBox>
+      )}
+      {submit.isSuccess && (
+        <MessageBox variant="success">Form updated!</MessageBox>
```
pjeweb, line 78:
```suggestion
        <MessageBox variant="success">Short link updated!</MessageBox>
```

apps/website/src/components/admin/short-links/ShortLinkForm.tsx
```diff
@@ -0,0 +1,114 @@
...
+      )}
+
+      <Fieldset legend="Form">
```
pjeweb, line 81:
```suggestion
      <Fieldset legend="Short link">
```

apps/website/src/components/admin/short-links/ShortLinkForm.tsx
```diff
@@ -0,0 +1,114 @@
...
+import { type FormSchema } from "@/server/db/short-links";
+
+import { Button, defaultButtonClasses } from "@/components/shared/Button";
+import { TextField } from "@/components/shared/form/TextField";
+import { Fieldset } from "@/components/shared/form/Fieldset";
+import { MessageBox } from "@/components/shared/MessageBox";
+
+type ShortLinkProps = {
+  action: "create" | "edit";
+  form?: ShortLinks;
```
pjeweb, line 21: Should `form` be renamed `shortLink`?

apps/website/src/components/admin/short-links/ShortLink.tsx
```diff
@@ -0,0 +1,140 @@
...
+  LinkButton,
+} from "@/components/shared/Button";
+import { ModalDialog } from "@/components/shared/ModalDialog";
+import { Headline } from "@/components/admin/Headline";
+import { Panel } from "@/components/admin/Panel";
+import IconPencil from "@/icons/IconPencil";
+import IconTrash from "@/icons/IconTrash";
+import type { AppRouter } from "@/server/trpc/router/_app";
+type RouterOutput = inferRouterOutputs<AppRouter>;
+type FormWithCount = RouterOutput["adminShortLinks"]["getLinks"][number];
```
pjeweb, line 17: Refactor `FormWithCount` -> `ShortLink`

apps/website/src/components/admin/short-links/ShortLink.tsx
```diff
@@ -0,0 +1,140 @@
...
+
+type LinkProps = {
+  form: FormWithCount;
```
pjeweb, line 20: Refactor `form` -> `shortLink`

apps/website/src/pages/admin/short-links/[linkId]/edit.tsx
```diff
@@ -0,0 +1,60 @@
...
+
+  const id = context.params?.linkId;
+  if (!id) {
+    return { notFound: true };
+  }
+
+  return {
+    props: {
+      ...adminProps,
+      formId: String(id),
```
Comment (pjeweb) [inline apps/website/src/pages/admin/short-links/[linkId]/edit.tsx:32]: Refactor `formId` -> `shortLinkId`

apps/website/src/pages/admin/short-links/[linkId]/edit.tsx
```diff
@@ -0,0 +1,60 @@
...
+  InferGetStaticPropsType<typeof getServerSideProps>
+> = ({ menuItems, formId }) => {
+  const link = trpc.adminShortLinks.getLink.useQuery(formId);
+
+  return (
+    <>
+      <Meta title="Edit Form | Admin" />
+
+      <AdminPageLayout title="Edit Form" menuItems={menuItems}>
+        <Headline>Edit Form</Headline>
```
Comment (pjeweb) [inline apps/website/src/pages/admin/short-links/[linkId]/edit.tsx:47]:
```suggestion
      <Meta title="Edit Short Link | Admin" />

      <AdminPageLayout title="Edit Short Link" menuItems={menuItems}>
        <Headline>Edit Short Link</Headline>
```

apps/website/src/pages/admin/short-links/[linkId]/edit.tsx
```diff
@@ -0,0 +1,60 @@
...
+
+  return {
+    props: {
+      ...adminProps,
+      formId: String(id),
+    },
+  };
+}
+
+const AdminEditFormPage: NextPage<
```
Comment (pjeweb) [inline apps/website/src/pages/admin/short-links/[linkId]/edit.tsx:37]:
```suggestion
const AdminEditShortLinkPage: NextPage<
```

apps/website/src/pages/admin/short-links/create.tsx
```diff
@@ -0,0 +1,38 @@
...
+}
+
+const AdminCreateFormPage: NextPage<
+  InferGetStaticPropsType<typeof getServerSideProps>
+> = ({ menuItems }) => {
+  return (
+    <>
+      <Meta title="Create Form | Admin" />
+
+      <AdminPageLayout title="Create Form" menuItems={menuItems}>
```
pjeweb, line 28:
```suggestion
const AdminCreateShortLinkPage: NextPage<
  InferGetStaticPropsType<typeof getServerSideProps>
> = ({ menuItems }) => {
  return (
    <>
      <Meta title="Create Short Link | Admin" />

      <AdminPageLayout title="Create Short Link" menuItems={menuItems}>
```

apps/website/src/pages/admin/short-links/create.tsx
```diff
@@ -0,0 +1,38 @@
...
+import { permissions } from "@/config/permissions";
+
+import { AdminPageLayout } from "@/components/admin/AdminPageLayout";
+import Meta from "@/components/content/Meta";
+import { ShortLinkForm } from "@/components/admin/short-links/ShortLinkForm";
+import { Headline } from "@/components/admin/Headline";
+import { Panel } from "@/components/admin/Panel";
+
+export async function getServerSideProps(context: NextPageContext) {
+  const adminProps = await getAdminSSP(context, permissions.manageForms);
```
pjeweb, line 13:
```suggestion
  const adminProps = await getAdminSSP(context, permissions.manageShortLinks);
```

apps/website/src/pages/admin/short-links/index.tsx
```diff
@@ -0,0 +1,30 @@
+import type { InferGetStaticPropsType, NextPage, NextPageContext } from "next";
+import { getAdminSSP } from "@/server/utils/admin";
+import { permissions } from "@/config/permissions";
+import Meta from "@/components/content/Meta";
+import { AdminPageLayout } from "@/components/admin/AdminPageLayout";
+import { ShortLink } from "@/components/admin/short-links/ShortLink";
+
+export async function getServerSideProps(context: NextPageContext) {
+  const adminProps = await getAdminSSP(context, permissions.manageForms);
```
pjeweb, line 9:
```suggestion
  const adminProps = await getAdminSSP(context, permissions.manageShortLinks);
```

apps/website/src/server/trpc/router/admin/short-links.ts
```diff
@@ -0,0 +1,77 @@
...
+} from "@/server/trpc/trpc";
+import { permissions } from "@/config/permissions";
+import { createForm, editForm, formSchema } from "@/server/db/short-links";
+
+const permittedProcedure = protectedProcedure.use(
+  createCheckPermissionMiddleware(permissions.manageShortLinks),
+);
+
+export const shortLinksRouter = router({
+  createOrEditForm: permittedProcedure
```
pjeweb, line 15:
```suggestion
  createOrEditShortLink: permittedProcedure
```

apps/website/src/server/db/short-links.ts
pjeweb, line 1: Refactor `Form`/`form` -> `ShortLink`/`shortLink` (`formSchema`,`FormSchema`, `createForm` etc.)

pjeweb (changes requested):
Looking good. There are some places where `form` should be renamed. That name was a bad choice ;)

Gonna have another look at the UI/UX later, but I'm pretty sure we can merge this soon.

apps/website/src/middleware.ts
pjeweb: Sorry for the yapping :) gonna try to collect thoughts and write it up better later after work!

JonasFranke: :D Ok I'll let it as is and after a month or so in prod we'll look at it again and decide what to do?

apps/website/src/middleware.ts
JonasFranke: without the star it didn't work (don't know why 🤷 I hate RegEx) I added a strict URL comparison to not react to slashes in slugs

JonasFranke:
> Gonna have another look at the UI/UX later, but I'm pretty sure we can merge this soon.
Did you look into it yet?

apps/website/src/components/admin/short-links/ShortLinkForm.tsx
```diff
@@ -0,0 +1,114 @@
...
+        <TextField
+          label="Slug (Public URL, alphanumeric, dashes allowed)"
+          name="slug"
+          pattern={SLUG_PATTERN}
+          inputMode="url"
+          defaultValue={shortLink?.slug || ""}
+          inputClassName="font-mono"
+          placeholder={convertToSlug(label)}
+          prefix={
+            <div className="cursor-default select-none pl-2 font-mono">{`${env.NEXT_PUBLIC_BASE_URL}/link/`}</div>
```
pjeweb, line 97:
```suggestion
            <div className="cursor-default select-none pl-2 font-mono">{`${env.NEXT_PUBLIC_BASE_URL}/l/`}</div>
```

apps/website/src/server/db/short-links.ts
```diff
@@ -0,0 +1,52 @@
...
+
+export type ShortLinkSchema = z.infer<typeof shortLinkSchema>;
+
+export const existingShortLinkSchema = shortLinkSchema.and(
+  z.object({
+    id: z.string().cuid(),
+  }),
+);
+
+export async function createForm(input: z.infer<typeof shortLinkSchema>) {
```
pjeweb, line 20: This still needs renaming

apps/website/src/server/db/short-links.ts
```diff
@@ -0,0 +1,52 @@
...
+      code: "BAD_REQUEST",
+      message: "Slug already exists",
+    });
+
+  return await prisma.shortLinks.create({
+    data: { ...input, slug },
+  });
+}
+
+export async function editForm(input: z.infer<typeof existingShortLinkSchema>) {
```
pjeweb, line 36: This still needs renaming

apps/website/src/server/db/short-links.ts
```diff
@@ -0,0 +1,52 @@
...
+
+export const existingShortLinkSchema = shortLinkSchema.and(
+  z.object({
+    id: z.string().cuid(),
+  }),
+);
+
+export async function createForm(input: z.infer<typeof shortLinkSchema>) {
+  const slug = convertToSlug(input.slug || input.label);
+  const existingShortLinkWithSlug = await prisma.form.findFirst({
```
pjeweb, line 22:
```suggestion
  const existingShortLinkWithSlug = await prisma.shortLinks.findFirst({
```

apps/website/src/server/db/short-links.ts
```diff
@@ -0,0 +1,52 @@
...
+    });
+
+  return await prisma.shortLinks.create({
+    data: { ...input, slug },
+  });
+}
+
+export async function editForm(input: z.infer<typeof existingShortLinkSchema>) {
+  const slug = convertToSlug(input.slug || input.label);
+  const existingShortLinkWithSlug = await prisma.form.findFirst({
```
pjeweb, line 38:
```suggestion
  const existingShortLinkWithSlug = await prisma.shortLinks.findFirst({
```

apps/website/src/server/trpc/router/admin/short-links.ts
```diff
@@ -0,0 +1,77 @@
...
+        console.log("Hallo Welt");
+        await ctx.prisma.shortLinksTracking.update({
+          where: { id: id },
+          data: { clicks: clicks },
+        });
+      } else {
+        const clicks: number = 1;
+        await ctx.prisma.shortLinksTracking.create({ data: { id, clicks } });
+      }
+    }),
```
pjeweb, line 72: This is not needed anymore i think

apps/website/src/server/utils/admin.ts
pjeweb, line 1: This can be higher in the admin nav, maybe even right after notifications

apps/website/src/pages/api/short-links/track-click.ts
```diff
@@ -0,0 +1,30 @@
...
+    });
+    if (query) {
+      await prisma.shortLinksTracking.update({
+        where: { id: id },
+        data: {
+          clicks: {
+            increment: 1,
+          },
+        },
+      });
```
pjeweb, line 24: Right now the code is missing the creation of the tracking entry. I think this was previously done in the trpc procedure addClick, but did not make it to the api. Maybe you can use an upsert (combined insert and update) for this?

```typescript
prisma.shortLinksTracking.upsert({
  where: { id },
  create: { id, clicks: 1 },
  update: { clicks: { increment: 1 } },
})
```

apps/website/src/pages/api/short-links/track-click.ts
pjeweb, line 1: Can we use `createTokenProtectedApiHandler` from `@/server/utils/api`? That way we do not need a new secret and can save on some duplicated code for the token handling. Also adds a little bit of type hinting and validation. You'd need to create and the Input schema then:

```typescript
export type TrackClickSchema = z.infer<typeof trackClickSchema>;
const trackClickSchema = z.object({ id: z.string().cuid() });
```

 In the middleware you can use `callEndpoint` from `@/server/utils/queue`:

```typescript

callEndpoint<TrackClickSchema>("/api/short-links/track-click", { id: shortLink.id })
```

apps/website/src/pages/api/short-links/track-click.ts
```diff
@@ -0,0 +1,30 @@
...
+export default async function handler(
+  req: NextApiRequest,
+  res: NextApiResponse,
+) {
+  const body = JSON.parse(req.body);
+  const id = body.id;
+  if (body.secret === env.SHORT_LINKS_TRACKING_SECRET) {
+    const query = await prisma.shortLinksTracking.findFirst({
+      where: { id: id },
+    });
```
pjeweb, line 15: (Maybe look at the comment below regarding the creation of tracking entries first :) or this does not make as much sense)

Do we really need to query the entry first? If you use the upsert this should not be necessary. It would be possible that someone called the api with an invalid id, but that would require the correct api token so i do not think we really care. Even if this inserts a click for a non-existent link, it should not matter.

apps/website/src/server/db/short-links.ts
```diff
@@ -0,0 +1,52 @@
+import { z } from "zod";
+import { TRPCError } from "@trpc/server";
+import { prisma } from "@/server/db/client";
+import { convertToSlug, SLUG_REGEX } from "@/utils/slugs";
+
+export const shortLinkSchema = z.object({
+  label: z.string(),
+  slug: z.string().regex(SLUG_REGEX).optional(),
+  link: z.string(),
```
pjeweb, line 9: I think we should require a valid link here. Probably zod's `z.string().url()` should work fine.

apps/website/src/server/db/short-links.ts
```diff
@@ -0,0 +1,52 @@
+import { z } from "zod";
+import { TRPCError } from "@trpc/server";
+import { prisma } from "@/server/db/client";
+import { convertToSlug, SLUG_REGEX } from "@/utils/slugs";
+
+export const shortLinkSchema = z.object({
+  label: z.string(),
+  slug: z.string().regex(SLUG_REGEX).optional(),
```
pjeweb, line 8: With the forms not requiring a label or slug was fine because we have a fallback to the id there. For the short links I think we should check there is a slug (or a label to infer the slug from). Not sure if thats easily doable with zod, so maybe just always require a non empty label.

apps/website/src/server/trpc/router/admin/short-links.ts
```diff
@@ -0,0 +1,77 @@
...
+          await editForm(data);
+          break;
+        }
+      }
+    }),
+
+  deleteLink: permittedProcedure
+    .input(z.string().cuid())
+    .mutation(async ({ ctx, input: id }) =>
+      ctx.prisma.shortLinks.delete({ where: { id } }),
```
pjeweb, line 42: We should probably also remove the tracking entry if it exists here.

pjeweb (changes requested):
Working pretty well in terms of UI/UX. Sorry for the slow review. Biggest problem right now is that click aren't tracked because the tracking entry is never created (see comments).

apps/website/src/pages/api/short-links/track-click.ts
JonasFranke: Didn't know we had such a cool function :)

MattIPv4:
👋 @JonasFranke would you be able to resolve the conflicts that're on this branch and get it up-to-date with main?

JonasFranke:
> 👋 @JonasFranke would you be able to resolve the conflicts that're on this branch and get it up-to-date with main?

Got it up-to-date with main! Don't ask what happened to my main branch while updating

apps/website/src/env/index.mjs
pjeweb, line 1: this file should be reverted. we do not need the env var anymore

apps/website/src/pages/admin/short-links/[linkId]/edit.tsx
```diff
@@ -0,0 +1,60 @@
+import type {
+  InferGetStaticPropsType,
+  NextPage,
+  GetServerSidePropsContext,
+} from "next";
+
+import { getAdminSSP } from "@/server/utils/admin";
+import { permissions } from "@/config/permissions";
```
Comment (pjeweb) [inline apps/website/src/pages/admin/short-links/[linkId]/edit.tsx:8]:
```suggestion
import { permissions } from "@/data/permissions";
```

apps/website/src/pages/admin/short-links/create.tsx
```diff
@@ -0,0 +1,38 @@
+import type { InferGetStaticPropsType, NextPageContext, NextPage } from "next";
+
+import { getAdminSSP } from "@/server/utils/admin";
+import { permissions } from "@/config/permissions";
```
pjeweb, line 4:
```suggestion
import { permissions } from "@/data/permissions";
```

apps/website/src/pages/admin/short-links/index.tsx
```diff
@@ -0,0 +1,30 @@
+import type { InferGetStaticPropsType, NextPage, NextPageContext } from "next";
+import { getAdminSSP } from "@/server/utils/admin";
+import { permissions } from "@/config/permissions";
```
pjeweb, line 3:
```suggestion
import { permissions } from "@/data/permissions";
```

apps/website/src/server/trpc/router/admin/short-links.ts
```diff
@@ -0,0 +1,63 @@
+import { z } from "zod";
+import {
+  createCheckPermissionMiddleware,
+  protectedProcedure,
+  router,
+} from "@/server/trpc/trpc";
+import { permissions } from "@/config/permissions";
```
pjeweb, line 7:
```suggestion
import { permissions } from "@/data/permissions";
```

apps/website/src/components/admin/short-links/ShortLinkForm.tsx
```diff
@@ -0,0 +1,114 @@
...
+
+      if (action === "edit") {
+        if (!shortLink) return;
+        submit.mutate({ action: "edit", id: shortLink.id, ...mutationData });
+      } else {
+        submit.mutate(
+          { action: "create", ...mutationData },
+          {
+            onSuccess: async () => {
+              await router.push(`/admin/short-links`);
```
pjeweb, line 61: super minor nit but this can be normal double quotes

apps/website/src/components/admin/short-links/ShortLinkForm.tsx
```diff
@@ -0,0 +1,114 @@
...
+          pattern={SLUG_PATTERN}
+          inputMode="url"
+          defaultValue={shortLink?.slug || ""}
+          inputClassName="font-mono"
+          placeholder={convertToSlug(label)}
+          prefix={
+            <div className="cursor-default select-none pl-2 font-mono">{`${env.NEXT_PUBLIC_BASE_URL}/l/`}</div>
+          }
+        />
+        <TextField
```
pjeweb, line 100: Missing the default value when editing

```suggestion
        <TextField
          defaultValue={shortLink?.link || ""}
```

apps/website/src/components/admin/short-links/ShortLinkForm.tsx
```diff
@@ -0,0 +1,114 @@
...
+          label="Name"
+          name="label"
+          value={label}
+          onChange={setLabel}
+        />
+        <TextField
+          label="Slug (Public URL, alphanumeric, dashes allowed)"
+          name="slug"
+          pattern={SLUG_PATTERN}
+          inputMode="url"
```
pjeweb, line 92:
```suggestion
          inputMode="url"
          isRequired
          type="url"
          pattern="https?://.+"
```

I'd probably throw on some more validation even though we do not really show validation error in the client.

apps/website/src/components/admin/short-links/ShortLinkForm.tsx
```diff
@@ -0,0 +1,114 @@
...
+      {submit.isSuccess && (
+        <MessageBox variant="success">Short link updated!</MessageBox>
+      )}
+
+      <Fieldset legend="Short link">
+        <TextField
+          label="Name"
+          name="label"
+          value={label}
+          onChange={setLabel}
```
pjeweb, line 86:
```suggestion
          onChange={setLabel}
          isRequired
```

I'd add the required attribute

pjeweb (changes requested):
Working great! Had a few rebase/merge problems and some small fixes.

apps/website/src/components/admin/short-links/ShortLinkForm.tsx
pjeweb: oh this is my mistake. this does not make sense. i wanted to suggest type=url etc for the url textfield not the slug...

JonasFranke:
@pjeweb I think it's ready to merge 🎉

pjeweb (approved):
Thank you! I think this should be great now :) I merged it to preview too for testing.

apps/website/.env.example
```diff
@@ -78,3 +78,4 @@ SUPER_USER_IDS=
 # Encryption passphrase for data at rest
 #   You can generate the secret via 'openssl rand -base64 24' on Linux
 DATA_ENCRYPTION_PASSPHRASE=
+
```
MattIPv4, line 81:
```suggestion
```
Unrelated diff feels unrelated?

apps/website/prisma/schema.prisma
```diff
@@ -282,6 +282,24 @@ model FormEntry {
   @@index([userId], name: "FormEntry_userId")
 }
 
+model ShortLinks {
+  id    String @id @default(cuid())
+  label String
+  slug  String
```
MattIPv4, line 288: Should we be telling the schema that slug is unique?

apps/website/src/components/admin/short-links/ShortLink.tsx
```diff
@@ -0,0 +1,142 @@
...
+            {"✅"}
+          </Button>
+        </td>
+        <td className="w-1/2 p-1">
+          <div className="flex flex-col gap-0.5">
+            <div className="text-xl">{shortLink.label}</div>
+            <div className="flex flex-col gap-1">
+              <Link
+                className="underline"
+                href={`/l/${shortLink.slug || shortLink.id}`}
```
MattIPv4, line 46: Looking at the schema, slug is required, so why do we need fallback to id here?

apps/website/src/components/admin/short-links/ShortLink.tsx
```diff
@@ -0,0 +1,142 @@
...
+                target="_blank"
+              >
+                Public Link: {shortLink.slug || shortLink.id}
```
MattIPv4, line 49: Ditto (also, I think it'd be smart to have the full link here with the short URL method/env/whatever-it-is-I-forgot?)

apps/website/src/components/admin/short-links/ShortLink.tsx
```diff
@@ -0,0 +1,142 @@
...
+        </td>
+        <td className="">
+          <Link className="underline" href={shortLink.link} target="_blank">
+            {shortLink.link}
+          </Link>
+        </td>
+        <td>
+          {clicks.data?.map((c) =>
+            shortLink.id === c.id ? String(c.clicks) : "",
+          )}
```
MattIPv4, line 62: This feels rather inefficient -- can we do a reduce over `clicks.data` once at the start to construct a `Map`, so we can do an O(1) lookup here, rather than a map for every row?

apps/website/src/middleware.ts
```diff
@@ -0,0 +1,34 @@
...
+  });
+  const shortLink = await res.json();
+
+  if (shortLink) {
+    event.waitUntil(
+      callEndpoint<TrackClickSchema>("/api/short-links/track-click", {
+        id: shortLink.id,
+      }),
+    );
+    return NextResponse.redirect(shortLink.link);
```
JonasFranke, line 25: Should we add the status code 308 or is 307 fine?

apps/website/src/pages/api/short-links/index.ts
```diff
@@ -0,0 +1,16 @@
...
+export default async function handler(
+  req: NextApiRequest,
+  res: NextApiResponse,
+) {
+  const body = JSON.parse(req.body);
+  const slug = body.slug;
+
+  const shortLink = await prisma.shortLinks.findFirst({
+    where: { slug: slug },
+  });
```
MattIPv4, line 14:
```suggestion
  const body = JSON.parse(req.body);

  const shortLink = await prisma.shortLinks.findFirst({
    where: { slug: body.slug },
  });
```
The slug const feels redundant here (body in theory is too, but I can see logically it makes sense to keep it in a const)

apps/website/src/middleware.ts
pjeweb: I'll leave that up to @MattIPv4. Im happy with the temporary redirect as the target can change. I think we discussed this in regards to the next config redirects.

apps/website/src/server/db/short-links.ts
```diff
@@ -0,0 +1,54 @@
...
+export type ShortLinkSchema = z.infer<typeof shortLinkSchema>;
+
+export const existingShortLinkSchema = shortLinkSchema.and(
+  z.object({
+    id: z.string().cuid(),
+  }),
+);
+
+export async function createShortLink(input: z.infer<typeof shortLinkSchema>) {
+  const slug = convertToSlug(input.slug || input.label);
```
MattIPv4, line 21: I'm tempted to say this logic should live in the client, so the logic here is just worrying about writing to the DB, and the client handles this fallback?

apps/website/src/server/db/short-links.ts
```diff
@@ -0,0 +1,54 @@
...
+
+  return await prisma.shortLinks.create({
+    data: { ...input, slug },
+  });
+}
+
+export async function editShortLink(
+  input: z.infer<typeof existingShortLinkSchema>,
+) {
+  const slug = convertToSlug(input.slug || input.label);
```
MattIPv4, line 39: Ditto

apps/website/src/middleware.ts
```diff
@@ -0,0 +1,34 @@
...
+  }
+
+  // If no match is found, redirect the user to the home page
+  return NextResponse.redirect(env.NEXT_PUBLIC_BASE_URL + "/");
```
MattIPv4, line 29: Is there a good way to just return the normal 404 page?

apps/website/src/middleware.ts
MattIPv4: Temporary is preferred here I think 👍 Anything permanent should probably be in the next config anyway rather than using this system

apps/website/src/components/admin/short-links/ShortLink.tsx
```diff
@@ -0,0 +1,143 @@
...
+            <div className="text-xl">{shortLink.label}</div>
+            <div className="flex flex-col gap-1">
+              <Link
+                href={`/l/${shortLink.slug}`}
+                target="_blank"
+                className="flex"
+              >
+                <div className="mr-1.5">Public Link:</div>
+                <div className="underline">
+                  {env.NEXT_PUBLIC_BASE_URL + "/l/" + shortLink.slug}
```
MattIPv4, line 53: Can we use `getShortBaseUrl` instead of the regular base URL here?

apps/website/src/components/admin/short-links/ShortLinkForm.tsx
```diff
@@ -0,0 +1,122 @@
...
+          ? String(formData.get("url"))
+          : "https://" + String(formData.get("url"));
+
+      const mutationData: ShortLinkSchema = {
+        label: convertToSlug(String(formData.get("label"))),
+        link: link,
+      };
+
+      const slug = formData.has("slug") && String(formData.get("slug"));
+      if (slug && slug !== "") {
```
MattIPv4, line 49:
```suggestion
      if (slug) {
```
An empty string is falsey so the second condition here is redundant

apps/website/src/components/admin/short-links/ShortLinkForm.tsx
```diff
@@ -0,0 +1,122 @@
...
+        <TextField
+          label="Slug (Public URL, alphanumeric, dashes allowed)"
+          name="slug"
+          inputMode="url"
+          pattern={SLUG_PATTERN}
+          defaultValue={shortLink?.slug || ""}
+          inputClassName="font-mono"
+          placeholder={convertToSlug(label)}
+          prefix={
+            <div className="cursor-default select-none pl-2 font-mono">{`${env.NEXT_PUBLIC_BASE_URL}/l/`}</div>
```
MattIPv4, line 102: Would probably be good to use the short URL base here too

apps/website/src/middleware.ts
```diff
@@ -0,0 +1,34 @@
...
+    event.waitUntil(
+      callEndpoint<TrackClickSchema>("/api/short-links/track-click", {
+        id: shortLink.id,
+      }),
+    );
+    return NextResponse.redirect(shortLink.link);
+  }
+
+  // If no match is found, redirect the user to the 404 page
+  return NextResponse.redirect(env.NEXT_PUBLIC_BASE_URL + "/404");
```
MattIPv4, line 29: Out of curiosity, what happens if we return nothing here -- does Next.js serve a 404 without a redirect happening?

apps/website/src/server/db/short-links.ts
```diff
@@ -0,0 +1,54 @@
...
+export type ShortLinkSchema = z.infer<typeof shortLinkSchema>;
+
+export const existingShortLinkSchema = shortLinkSchema.and(
+  z.object({
+    id: z.string().cuid(),
+  }),
+);
+
+export async function createShortLink(input: z.infer<typeof shortLinkSchema>) {
+  const slug = input.slug || input.label;
```
MattIPv4, line 21: Is the fallback still needed here? It should be guaranteed that the body will have a slug now?

apps/website/src/server/db/short-links.ts
```diff
@@ -0,0 +1,54 @@
...
+
+  return await prisma.shortLinks.create({
+    data: { ...input, slug },
+  });
+}
+
+export async function editShortLink(
+  input: z.infer<typeof existingShortLinkSchema>,
+) {
+  const slug = input.slug || input.label;
```
MattIPv4, line 39: Ditto

apps/website/src/middleware.ts
JonasFranke: It serves the Next.js standart 404 page
image

JonasFranke: Maybe my local env is bugged. If I open a normal page with an invalid url I also get the next.js 404 page

JonasFranke: But `.next` should work

apps/website/src/components/admin/short-links/ShortLink.tsx
```diff
@@ -0,0 +1,143 @@
...
+            <div className="text-xl">{shortLink.label}</div>
+            <div className="flex flex-col gap-1">
+              <Link
+                href={`/l/${shortLink.slug}`}
+                target="_blank"
+                className="flex"
+              >
+                <div className="mr-1.5">Public Link:</div>
+                <div className="underline">
+                  {env.NEXT_PUBLIC_SHORT_BASE_URL + "/l/" + shortLink.slug}
```
MattIPv4, line 53: Please use the [`getShortBaseUrl`](https://github.com/alveusgg/alveusgg/blob/main/apps/website/src/utils/short-url.ts) helper here please?

apps/website/src/components/admin/short-links/ShortLinkForm.tsx
```diff
@@ -0,0 +1,123 @@
...
+        <TextField
+          label="Slug (Public URL, alphanumeric, dashes allowed)"
+          name="slug"
+          inputMode="url"
+          pattern={SLUG_PATTERN}
+          defaultValue={shortLink?.slug || ""}
+          inputClassName="font-mono"
+          placeholder={convertToSlug(label)}
+          prefix={
+            <div className="cursor-default select-none pl-2 font-mono">{`${env.NEXT_PUBLIC_SHORT_BASE_URL}/l/`}</div>
```
MattIPv4, line 103: Ditto, use `getShortBaseUrl` please

apps/website/src/server/db/short-links.ts
```diff
@@ -0,0 +1,54 @@
...
+  });
+  if (existingShortLinkWithSlug)
+    throw new TRPCError({
+      code: "BAD_REQUEST",
+      message: "Slug already exists",
+    });
+
+  return await prisma.shortLinks.create({
+    data: { ...input, slug },
+  });
```
MattIPv4, line 33:
```suggestion
  const existingShortLinkWithSlug = await prisma.shortLinks.findFirst({
    where: { slug: input.slug },
  });
  if (existingShortLinkWithSlug)
    throw new TRPCError({
      code: "BAD_REQUEST",
      message: "Slug already exists",
    });

  return await prisma.shortLinks.create({
    data: input,
  });
```
We can clean this up a bit now we don't have the fallback

apps/website/src/server/db/short-links.ts
```diff
@@ -0,0 +1,54 @@
...
+    throw new TRPCError({
+      code: "BAD_REQUEST",
+      message: "Slug already exists",
+    });
+
+  const { id, ...data } = input;
+  return await prisma.shortLinks.update({
+    where: { id: id },
+    data: { ...data, slug },
+  });
```
MattIPv4, line 53:
```suggestion
  const existingShortLinkWithSlug = await prisma.shortLinks.findFirst({
    where: { slug: input.slug, id: { not: input.id } },
  });
  if (existingShortLinkWithSlug)
    throw new TRPCError({
      code: "BAD_REQUEST",
      message: "Slug already exists",
    });

  const { id, ...data } = input;
  return await prisma.shortLinks.update({
    where: { id },
    data,
  });
```
Ditto

MattIPv4:
Apologies for the back and forth, that should be the last set of nitpicks 💚

apps/website/src/components/admin/short-links/ShortLink.tsx
JonasFranke: didn't know we had it

PR #538, opened 2024-02-12
Fix OAuth Client Access Tokens not refreshing

### LINKED ISSUE #536: [Bug]: OAuth client access token does not get updated when expired

`getClientCredentialsAccessToken` currently fails to update an existing OAuth client access token because it `create` while an existing token with the unique key already exists.

Update the token for the client + service if it already exists, otherwise create it.

### Steps to reproduce

- Create client access token by e.g. going to `/admin/twitch`
- Set token expire_at date to the past in the database (or wait ;) )
- Open `/admin/twitch` again and see error in dev tools or dev server stdout

### Browser specs

n/a

### Anything else?

_No response_

pjeweb:
## Describe your changes

Resolves #536 

...

...

### DIFF
--- a/apps/website/src/server/utils/oauth2.ts
+++ b/apps/website/src/server/utils/oauth2.ts
@@ -60,13 +60,13 @@ export async function getClientCredentialsAccessToken(
     oAuth2ServiceUrls[service].accessTokenPath,
     undefined,
   );
-  const cache = await prisma.clientAccessToken.findFirst({
+  const existing = await prisma.clientAccessToken.findFirst({
     where: { client_id: clientId, service },
   });
 
   let accessToken;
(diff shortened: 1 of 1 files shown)


pjeweb:
This is deployed to preview.alveus.gg

MattIPv4 (approved):
LGTM

Maybe I'm missing something in this logic flow, or maybe its elsewhere, but is there not a refresh token stored/used?

pjeweb:
> Maybe I'm missing something in this logic flow, or maybe its elsewhere, but is there not a refresh token stored/used?

This auth flow does not provide a refresh token. It is just using the client secret. https://dev.twitch.tv/docs/authentication/getting-tokens-oauth/#client-credentials-grant-flow

PR #539, opened 2024-02-12
Update GitHub Actions to latest major versions

JonasFranke:
## Describe your changes

Update jobs to their latest version to fix the appearing warning.

> Node.js 16 actions are deprecated. Please update the following actions to use Node.js 20: actions/checkout@v3, actions/setup-node@v3, pnpm/action-setup@v2, actions/cache@v3. For more information see: https://github.blog/changelog/2023-09-22-github-actions-transitioning-from-node-16-to-node-20/.

### DIFF
--- a/.github/workflows/pull-request.yml
+++ b/.github/workflows/pull-request.yml
@@ -7,15 +7,15 @@ jobs:
 
     steps:
       - name: Checkout
-        uses: actions/checkout@v3
+        uses: actions/checkout@v4.1.1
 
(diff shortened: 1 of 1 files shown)


MattIPv4 (changes requested):
👀 I'm generally not a fan of pinning to exact versions in GitHub Actions -- can we stick to just pinning to the major tag?

All the major updates are just jumps to Node.js 20, so in principle this change is fine.

MattIPv4 (approved):

PR #543, opened 2024-02-14
Add collab for Emily Wang (emilyywang)

pjeweb:
## Describe your changes

Adds the most recent collaboration, Emily Wang (emilyywang)

Video is correct, date is correct, name/link are correct

### DIFF
--- a/apps/website/src/pages/collaborations.tsx
+++ b/apps/website/src/pages/collaborations.tsx
@@ -28,6 +28,12 @@ type Collaboration = {
 type Collaborations = Record<string, Collaboration>;
 
 const collaborations = {
+  emilyywang: {
+    name: "Emily Wang",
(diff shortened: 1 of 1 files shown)


apps/website/src/pages/collaborations.tsx
```diff
@@ -28,6 +28,12 @@ type Collaboration = {
 type Collaborations = Record<string, Collaboration>;
 
 const collaborations = {
+  emilyywang: {
```
MattIPv4, line 31:
```suggestion
  emilyWang: {
```
Curious on your opinion here, I know we don't really have a set way of doing these?

pjeweb: i dont really mind. your suggestion seems more consistent though

MattIPv4 (approved):

PR #551, opened 2024-02-17
Add hour tracking for Give an Hour for Earth

### LINKED ISSUE #509: [Idea]: WWF 'Give an Hour' tracking for Show and Tell

On Show and Tell submissions, from March 1st through to April 22nd, enable a checkbox for users to indicate their submission was a 'Give an Hour' activity, and if checked show an input for how many hours were spent on the activity.

On the Show and Tell homepage, from March 1st through to April 22nd, show a progress bar at the top for total hours spent as part of 'Give an Hour', ideally as a progress bar that automatically picks a target based on the current total.

Aim to keep the internal naming generic, as it is likely that after April 22nd we may want to switch this over to a more generic wording that allows folks to mark their submissions as general volunteering, so we can track activity from the community on an on-going basis.

Supporting WWF + Give an Hour, encouraging volunteering and other related activities.

pjeweb:
## Describe your changes

Resolves #509 

Deployed to https://preview.alveus.gg

- Test submitting a show and tell entry
- Test editing an existing show and tell entry
- Test the progress bar updates when the entry is approved (not before) on Show and Tell page and Give an Hour page

### DIFF
--- a/apps/website/prisma/schema.prisma
+++ b/apps/website/prisma/schema.prisma
@@ -357,22 +357,24 @@ model LinkAttachment {
 }
 
 model ShowAndTellEntry {
-  id             String    @id @default(cuid())
-  userId         String?
-  title          String
-  text           String    @db.Text
-  displayName    String?
-  createdAt      DateTime  @default(now())
-  updatedAt      DateTime  @default(now())
-  approvedAt     DateTime?
-  seenOnStreamAt DateTime?
-  seenOnStream   Boolean   @default(false)
+  id                  String    @id @default(cuid())
+  userId              String?
+  title               String
+  text                String    @db.Text
+  displayName         String?
+  createdAt           DateTime  @default(now())
+  updatedAt           DateTime  @default(now())
+  approvedAt          DateTime?
+  seenOnStreamAt      DateTime?
+  seenOnStream        Boolean   @default(false)
+  volunteeringMinutes Int?
 
   attachments ShowAndTellEntryAttachment[]
   user        User?                        @relation(fields: [userId], references: [id])
 
   @@index([userId], name: "ShowAndTellEntry_userId")
   @@index([seenOnStream, approvedAt, updatedAt], name: "ShowAndTellEntry_public")
+  @@index([volunteeringMinutes, approvedAt, createdAt], name: "ShowAndTellEntry_volunteeringMinutes")
 }
 
 model ShowAndTellEntryAttachment {
--- a/apps/website/prisma/schema.prisma
+++ b/apps/website/prisma/schema.prisma
@@ -373,8 +373,8 @@ model ShowAndTellEntry {
   user        User?                        @relation(fields: [userId], references: [id])
 
   @@index([userId], name: "ShowAndTellEntry_userId")
+  @@index([approvedAt, createdAt], name: "ShowAndTellEntry_approved")
   @@index([seenOnStream, approvedAt, updatedAt], name: "ShowAndTellEntry_public")
-  @@index([approvedAt, createdAt], name: "ShowAndTellEntry_volunteeringMinutes")
 }
 
 model ShowAndTellEntryAttachment {
--- a/apps/website/prisma/schema.prisma
+++ b/apps/website/prisma/schema.prisma
@@ -374,7 +374,7 @@ model ShowAndTellEntry {
 
   @@index([userId], name: "ShowAndTellEntry_userId")
   @@index([seenOnStream, approvedAt, updatedAt], name: "ShowAndTellEntry_public")
-  @@index([volunteeringMinutes, approvedAt, createdAt], name: "ShowAndTellEntry_volunteeringMinutes")
+  @@index([approvedAt, createdAt], name: "ShowAndTellEntry_volunteeringMinutes")
 }
 
 model ShowAndTellEntryAttachment {
--- a/apps/website/src/components/shared/form/NumberField.tsx
+++ b/apps/website/src/components/shared/form/NumberField.tsx
@@ -36,6 +36,7 @@ export type NumberFieldProps = AriaNumberFieldProps & {
   name: string;
   className?: string;
   inputClassName?: string;
+  groupClassName?: string;
   labelClassName?: string;
   list?: string;
   prefix?: ReactNode;
@@ -87,15 +88,15 @@ export function NumberField(props: NumberFieldProps) {
       <label className={props.labelClassName} {...labelProps}>
         {props.label}
       </label>
-      <div className="flex" {...groupProps}>
+      <div className={classes("flex", props.groupClassName)} {...groupProps}>
         {props.showButtons && (
           <NumberButton {...decrementButtonProps}>-</NumberButton>
         )}
-        <div className="flex w-full items-center gap-1 rounded-sm border border-gray-700 bg-white text-gray-500">
+        <div className="flex w-full items-center justify-center rounded-sm border border-gray-700 bg-white text-gray-500 focus-within:outline">
           {props.prefix}
           <input
             className={classes(
-              "w-full flex-1 bg-white p-1 px-2 text-black",
+              "w-full flex-1 bg-white p-1 px-2 text-black focus:outline-0",
               props.inputClassName,
             )}
             {...inputProps}
--- a/apps/website/src/components/shared/form/NumberField.tsx
+++ b/apps/website/src/components/shared/form/NumberField.tsx
@@ -83,6 +83,10 @@ export function NumberField(props: NumberFieldProps) {
     decrementButtonProps,
   } = useNumberField(customizedProps, state, ref);
 
+  const focusInput = useCallback(() => {
+    ref.current?.focus();
+  }, []);
+
   return (
     <div className={props.className || "flex-1"}>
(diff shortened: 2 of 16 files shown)


apps/website/prisma/schema.prisma
```diff
@@ -357,22 +357,24 @@ model LinkAttachment {
...
+  seenOnStreamAt      DateTime?
+  seenOnStream        Boolean   @default(false)
+  volunteeringMinutes Int?
 
   attachments ShowAndTellEntryAttachment[]
   user        User?                        @relation(fields: [userId], references: [id])
 
   @@index([userId], name: "ShowAndTellEntry_userId")
   @@index([seenOnStream, approvedAt, updatedAt], name: "ShowAndTellEntry_public")
+  @@index([volunteeringMinutes, approvedAt, createdAt], name: "ShowAndTellEntry_volunteeringMinutes")
```
MattIPv4, line 377: 🤔 If we're trying to optimise for the frequent query to get the total minutes, we don't want the minutes themselves actually in the index -- we just want to be able to lookup all the rows that're approved and in the time range?

apps/website/src/components/show-and-tell/GiveAnHourProgress.tsx
MattIPv4, line 1: I wonder if we could show a message like `Loading hours given...` in place of `0 hours given` on first load?

apps/website/src/components/shared/form/NumberField.tsx
```diff
@@ -1,21 +1,49 @@
...
 
+function NumberButton(
+  props: AriaButtonOptions<"button"> & { children: ReactNode },
+) {
+  const ref = useRef(null);
+  const { buttonProps } = useButton(props, ref);
+  return (
+    <button
+      className={classes(
+        "m-1 block w-12 rounded-full bg-gray-700 text-lg text-white",
```
MattIPv4, line 23: You might want to set a fixed height on these as well? They're not quite circles currently (and it was visually upsetting to me lol):

image

I'm not sure w-12 is doing anything here currently as the parent container this is used in has a flex w/ width set to min-content and this doesn't have a flex-shrink 0, so it's being shrunk.

I got this looking nice by setting w-8 h-8 flex-shrink-0 (potentially also flex items-center justify-center)

```suggestion
        "m-1 block w-8 h-8 flex-shrink-0 flex items-center justify-center rounded-full bg-gray-700 text-lg text-white",
```

apps/website/src/components/shared/form/NumberField.tsx
```diff
@@ -47,34 +75,46 @@ export function NumberField(props: NumberFieldProps) {
...
-            "w-full flex-1 bg-white p-1 px-2 text-black",
-            props.inputClassName,
+      <label className={props.labelClassName} {...labelProps}>
+        {props.label}
+      </label>
+      <div className={classes("flex", props.groupClassName)} {...groupProps}>
+        {props.showButtons && (
+          <NumberButton {...decrementButtonProps}>-</NumberButton>
+        )}
+        <div className="flex w-full items-center justify-center rounded-sm border border-gray-700 bg-white text-gray-500 focus-within:outline">
```
MattIPv4, line 95: Can we make it so a click anywhere in here focuses the actual input? Visually it looks like you could click on "hours" in the form, but it doesn't do anything currently. (I guess with the exception of click on the reset x)

apps/website/src/components/show-and-tell/ShowAndTellEntryForm.tsx
```diff
@@ -321,38 +381,35 @@ export function ShowAndTellEntryForm({
...
           </div>
 
-          <p className="text-sm italic opacity-75">
-            <strong>Give an Hour</strong> tracking will be available from March
-            1st.
-          </p>
+          {!canTrackGiveAnHour && (
+            <p className="text-sm italic opacity-75">
+              <strong>Give an Hour</strong> tracking will be available from
+              March 1st.
```
MattIPv4, line 410: Silly nitpick, should the date here come from the config? And maybe handling for the time being after the date as well with a 'tracking was available from x to y'?

apps/website/src/pages/show-and-tell/give-an-hour.tsx
```diff
@@ -77,13 +77,6 @@ const Card = ({
...
 
 const GiveAnHourPage: NextPage = () => {
-  // Demo logic for the progress bar
-  const router = useRouter();
-  const [progressHours, setProgressHours] = useState(0);
-  const onClickProgress = useCallback(() => {
-    if ("demo" in router.query) setProgressHours((prev) => prev + 4);
-  }, [router.query.demo]);
-
   return (
```
MattIPv4, line 80: This can be an implicit return now?

apps/website/src/server/db/show-and-tell.ts
```diff
@@ -91,6 +91,7 @@ const showAndTellSharedInputSchema = z.object({
   text: z.string().max(1_000),
   imageAttachments: imageAttachmentsSchema,
   videoLinks: videoLinksSchema.max(MAX_VIDEOS),
+  volunteeringMinutes: z.number().int().nonnegative().nullable(),
```
MattIPv4, line 94: Should this be positive? Allowing someone to submit 0 hours seems silly -- they should just not check the box?

MattIPv4 (commented):
I'm liking the blue/green gradient as the color scheme for this!

apps/website/prisma/schema.prisma
pjeweb: oh right i think i was selecting on volunteeringMinute NOT NULL at some point

apps/website/src/components/show-and-tell/ShowAndTellEntryForm.tsx
pjeweb: I added start and end time and kept the info visible always (before, during, and after), using the default mm/dd/yyyy date formatting. I hope thats okay?

apps/website/prisma/schema.prisma
```diff
@@ -357,22 +357,24 @@ model LinkAttachment {
...
+  seenOnStreamAt      DateTime?
+  seenOnStream        Boolean   @default(false)
+  volunteeringMinutes Int?
 
   attachments ShowAndTellEntryAttachment[]
   user        User?                        @relation(fields: [userId], references: [id])
 
   @@index([userId], name: "ShowAndTellEntry_userId")
   @@index([seenOnStream, approvedAt, updatedAt], name: "ShowAndTellEntry_public")
+  @@index([approvedAt, createdAt], name: "ShowAndTellEntry_volunteeringMinutes")
```
MattIPv4, line 377:
```suggestion
  @@index([approvedAt, createdAt], name: "ShowAndTellEntry_approved")
```
Nitpick, but this index isnt really for the minutes now, its just a generic index for approval

apps/website/src/components/shared/form/NumberField.tsx
```diff
@@ -47,34 +75,52 @@ export function NumberField(props: NumberFieldProps) {
...
+          <div className="contents" onClick={focusInput}>
+            {props.prefix}
+            <input
+              className={classes(
+                "w-full flex-1 bg-white p-1 px-2 text-black focus:outline-0",
```
MattIPv4, line 104: I'm still seeing an outline here -- it seems to be `focus-within` that's causing it to show?

```suggestion
                "w-full flex-1 bg-white p-1 px-2 text-black outline-none",
```

Maybe this will ensure it never shows? I tried focus-withing:outline-0 but it doesn't seem to win.

apps/website/src/components/show-and-tell/ShowAndTellEntryForm.tsx
```diff
@@ -61,6 +104,39 @@ export function ShowAndTellEntryForm({
...
+    } else if (now > end) {
+      active = false;
+      verb = "was";
+    }
+
+    setTrackingStatus({
+      active,
+      text: `tracking ${verb} available from ${trackingStatusFrom} to ${trackingStatusTo}`,
+    });
+  }, []);
```
MattIPv4, line 133: Nitpick, this could be a useMemo rather than a state + effect

pjeweb: oh true. i was under the impression the form was SSR'ed but it isnt

apps/website/src/data/show-and-tell.ts
```diff
@@ -1,2 +1,5 @@
 export const MAX_IMAGES = 16;
 export const MAX_VIDEOS = 6;
+
+export const giveAnHourStart = "2024-02-01T00:00:00";
```
MattIPv4, line 4:
```suggestion
export const giveAnHourStart = "2024-03-01T00:00:00";
```

MattIPv4 (approved):
Otherwise, LGTM

apps/website/src/data/show-and-tell.ts
```diff
@@ -1,2 +1,5 @@
...
+export const giveAnHourEnd = "2024-03-22T23:59:59";
```
MattIPv4, line 5:
```suggestion
export const giveAnHourEnd = "2024-04-22T23:59:59";
```

MattIPv4 (approved):

PR #552, opened 2024-02-17
Fix Vite test deprecation warning (switch to ESM)

JonasFranke:
## Describe your changes

Vitetest shows this deprecation warning in the test workflow
> The CJS build of Vite's Node API is deprecated. See https://vitejs.dev/guide/troubleshooting.html#vite-cjs-node-api-deprecated for more details.

This PR fixes that


apps/website/build-scripts/create-data.ts
```diff
@@ -8,10 +8,17 @@ import { isActiveAmbassadorKey } from "@alveusgg/data/src/ambassadors/filters";
 import { camelToKebab, sentenceToKebab } from "../src/utils/string-case";
 import { typeSafeObjectKeys } from "../src/utils/helpers";
 
-mkdirSync(join(__dirname, "../src/data/generated"), { recursive: true });
+const dirname = new URL(".", import.meta.url).pathname;
+
+mkdirSync(join(dirname, "../src/data/generated").replace("\\", ""), {
```
MattIPv4, line 13:
```suggestion
mkdirSync(new URL("../src/data/generated", import.meta.url), {
```
Instead of using a weird hack to get a dirname equivalent, can we just use the proper way to do it with URL directly (please apply the same to the other paths below)

MattIPv4 (changes requested):

MattIPv4 (commented):
I think `next.config.mjs` can just become `next.config.js` now? As can `prettier.config.mjs`? Can any of the other CJS files be swapped to ESM?

pjeweb (changes requested):

pjeweb:
> I think `next.config.mjs` can just become `next.config.js` now? As can `prettier.config.mjs`? Can any of the other CJS files be swapped to ESM?

I think the lint-staged config can be a .js with esm too.

MattIPv4:
👋 @JonasFranke wondering how things are going here, and if you need any help to get this across the line?

JonasFranke:
> 👋 @JonasFranke wondering how things are going here, and if you need any help to get this across the line?

👋 I may need some help with the lint-staged files. I'm not able to convert them to es modules.

pjeweb:
I made a follow-up PR (#561) moving more configs to esm/ts (incl. lint-staged) and cleaning up the files in general. If you want to take some of those commits in here do that, or we make it a separate PR.

apps/website/next-sitemap.config.js
```diff
@@ -1,10 +1,18 @@
...
   // If there is a NEXT_PUBLIC_VERCEL_URL set, use that like NextAuth.js does
-  siteUrl: process.env.NEXT_PUBLIC_VERCEL_URL
+  const url = process.env.NEXT_PUBLIC_VERCEL_URL
     ? `https://${process.env.NEXT_PUBLIC_VERCEL_URL}`
-    : process.env.NEXT_PUBLIC_BASE_URL,
+    : process.env.NEXT_PUBLIC_BASE_URL;
+
+  //Fallback to localhost if no public url is present
+  return url ? url : "http://localhost";
+};
```
MattIPv4, line 11: GitHub won't let me apply a suggestion here, but I'm not sure I follow why this needs to be a method that gets called?

```js
// If there is a NEXT_PUBLIC_VERCEL_URL set, use that like NextAuth.js does
// Fallback to localhost if no public url is present
const siteUrl =  (process.env.NEXT_PUBLIC_VERCEL_URL
  ? `https://${process.env.NEXT_PUBLIC_VERCEL_URL}`
  : process.env.NEXT_PUBLIC_BASE_URL) || "http://localhost";
```

MattIPv4:
I'm fine w/ merging this once that final comment is resolved, and then getting your follow-up to move more things over to ESM in after.

apps/website/next-sitemap.config.js
```diff
@@ -1,10 +1,13 @@
...
+
+const siteUrl =
+  (process.env.NEXT_PUBLIC_VERCEL_URLnpm
     ? `https://${process.env.NEXT_PUBLIC_VERCEL_URL}`
-    : process.env.NEXT_PUBLIC_BASE_URL,
+    : process.env.NEXT_PUBLIC_BASE_URL) || "http://localhost";
+
+/** @type {import('next-sitemap').IConfig} */
+const config = {
+  siteUrl: siteUrl,
```
MattIPv4, line 10:
```suggestion
  // If there is a NEXT_PUBLIC_VERCEL_URL set, use that like NextAuth.js does
  // Fallback to localhost if we don't have a URL from the env for any reason
  siteUrl: (process.env.NEXT_PUBLIC_VERCEL_URL
    ? `https://${process.env.NEXT_PUBLIC_VERCEL_URL}`
    : process.env.NEXT_PUBLIC_BASE_URL) || "http://localhost",
```
We can just do this inline as before rather than having a constant

apps/website/next-sitemap.config.js
```diff
@@ -1,10 +1,17 @@
 // @ts-check
+
+const siteUrl =
+  (process.env.NEXT_PUBLIC_VERCEL_URLnpm
+    ? `https://${process.env.NEXT_PUBLIC_VERCEL_URL}`
+    : process.env.NEXT_PUBLIC_BASE_URL) || "http://localhost";
+
```
MattIPv4, line 7:
```suggestion
```

MattIPv4 (approved):

pjeweb (approved):

Issue #427, opened 2023-10-08
[Idea]: Make collabs sortable


pjeweb:
### Describe your idea / enhancement

Similar to ambassadors it would be great to be able to sort the collabs by name, maybe add a text search (suggested by josh_raiden in twitch chat)

### How would this idea / enhancement be beneficial?

Helps find collabs

MattIPv4:
Text search seems a bit redundant to me, almost all browsers have that functionality built-in. Allowing sorting by oldest instead of newest seems sensible to me though. Anything else you think would make sense to sort by?

pjeweb:
> Text search seems a bit redundant to me, almost all browsers have that functionality built-in.

Thats fair, although i think a dedicated search makes it easier for less advanced users and on mobile devices, but def. not as important.

> Anything else you think would make sense to sort by?

Can't really think of much else other than date and name. Sorting by length or "vod vs. video" does not really make much sense imo.

Not sure if we think thats good, but we could sort them by "relevance" as bad as that sounds. Having Ludwig up at the top would probably be good 🤷 


PS: I might open another issue, but it might be nice to have a small overview of all collabs without having to scroll down (maybe just the collaborators' profile pics?). The two column layout also looks great, but makes it a bit harder to scan fast.  But thats a different story.

MattIPv4:
> Not sure if we think thats good, but we could sort them by "relevance". As bad as that sounds, having Ludwig up at the top would probably be good 🤷

We could probably do a sort based on following, so its an actual measurable metric. Might be something to run past Alveus folks before we do though, 'cause I'm not sure its the greatest look to have that.



> PS: I might open another issue, but it might be nice to have a small overview of all collabs without having to scroll down (maybe just the collaborators' profile pics?). The two column layout also looks great, but makes it a bit harder to scan fast. But thats a different story.

Sounds like a good idea to me, to improve navigation as the page gets longer and longer.

PR #558, opened 2024-02-21
Use Grouped component/hook for events + collaborations

MattIPv4:
## Describe your changes

Use the Grouped component/hook to handle the grouping and rendering of content on the events + collaborations pages. This provides some further consistency across all our pages that use grouping for content presentation, and sets the groundwork to enable #427.

Note that the dropdown has not been added to either page, as there is only one sort option (chronological) at present.

Events page now has year headings, with the current year being visually hidden.

Collaborations page visually remains the same (note that some anchors will have changed due to the logic changes).

### DIFF
--- a/apps/website/src/components/content/Grouped.tsx
+++ b/apps/website/src/components/content/Grouped.tsx
@@ -7,6 +7,8 @@ export interface GroupedProps<T> {
   option: string;
   group: string | null;
   name: string | null;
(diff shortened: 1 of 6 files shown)


pjeweb (approved):
~~this will change the url fragments slightly but i guess we dont really use the links anywhere and its not super important even it someone has an old link~~ you said that already

Issue #136, opened 2023-03-31
Add JSON-LD data to pages


MattIPv4:
Investigate what JSON-LD schema/data makes sense to expose on pages, and expose it.

MattIPv4:
Closing out as JSON-LD data was added for Animal Quest in #560, and I don't think any other pages will benefit from markup.

PR #560, opened 2024-02-23
Add JSON-LD data to AQ episodes

MattIPv4:
## Describe your changes

Adds JSON-LD (#136) to the Animal Quest episode pages, in yet another attempt to get Google to recognise them as videos.

https://search.google.com/test/rich-results detects the video

https://validator.schema.org/ reports no schema issues

### DIFF
--- a/apps/website/src/pages/animal-quest/[episodeName].tsx
+++ b/apps/website/src/pages/animal-quest/[episodeName].tsx
@@ -2,6 +2,7 @@ import type { GetStaticPaths, GetStaticProps, NextPage } from "next";
 import Image from "next/image";
 import { useRouter } from "next/router";
 import { Fragment, useEffect, useMemo, useState } from "react";
+import type { Episode } from "schema-dts";
 
 import animalQuest, {
   hosts,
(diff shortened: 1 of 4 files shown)


apps/website/src/pages/animal-quest/[episodeName].tsx
```diff
@@ -502,6 +515,37 @@ const AnimalQuestEpisodePage: NextPage<AnimalQuestEpisodePageProps> = ({
           </Consent>
         </Section>
       </div>
+
+      <JsonLD<Episode>
```
Comment (MattIPv4) [inline apps/website/src/pages/animal-quest/[episodeName].tsx:519]: @pjeweb maybe you have an idea here -- this would not work for me inside `<Meta>`, React/Next just didn't render it, which felt very odd?

Comment (MattIPv4) [reply apps/website/src/pages/animal-quest/[episodeName].tsx:519]: Well that's even weirder

Comment (pjeweb) [reply apps/website/src/pages/animal-quest/[episodeName].tsx:519]: yeah no clue. it just get nuked when the head gets rendered. and you cannot use the `<Script strategy="beforeInteractive" />` tag since that is only allowed in the layout/document (https://nextjs.org/docs/messages/no-before-interactive-script-outside-document)

Comment (pjeweb) [reply apps/website/src/pages/animal-quest/[episodeName].tsx:519]: did you try to deploy the version with it in the header and run the tests?

pjeweb (approved):
pogg

apps/website/src/pages/animal-quest/[episodeName].tsx
Comment (MattIPv4) [reply apps/website/src/pages/animal-quest/[episodeName].tsx:519]: Negative, just building locally. It always seemed to disappear as soon as JS runs 🤷

PR #561, opened 2024-02-24
Cleanup configs (use esm/ts)

pjeweb:
## Describe your changes

Follow up to #552 

This PR cleans up a bunch of things:
- use esm by default (`"type": "module"` in `package.json`), renaming `.mjs` files to `.js` and simplifying imports for those
- enable typescript checking by default, adding opt-out comments for the remaining files where types are incomplete
- convert `lint-staged.config.js`s to ESM
- convert `tailwind.config.js` to Typescript/ESM
- clean up `sw.tsup.config.ts` and `vitest.config.ts` (export a const instead of an anonymous config)
- move PostCSS and prettier config to the `package.json` (not 100% sure about this)
- cleanup service worker code, deleting the unused root service worker and enabling type checking and linting for the remaining
- fix a few eslint warnings for unused imports

- Check everything still builds and runs on dev server
- Check type checking and linting works
- Check tests still run


apps/website/sw.tsup.config.ts
```diff
@@ -1,9 +1,9 @@
...
-import { env } from "@/env/index.mjs";
 
-export default defineConfig((options) => ({
+import "./build-scripts/env";
+import { env } from "@/env";
+
+const config = defineConfig((options) => ({
   entry: {
     "push/alveus/PushServiceWorker": "src/sw/PushServiceWorker.ts",
     //RootServiceWorker: "src/sw/RootServiceWorker.ts",
```
MattIPv4, line 9:
```suggestion
```
Can we remove this line given you've deleted the file?

apps/website/vitest.config.ts
```diff
@@ -1,11 +1,14 @@
...
+
+const config = defineConfig({
   resolve: {
     alias: {
       "@": fileURLToPath(new URL("./src", import.meta.url)),
     },
   },
 });
+
+export default config;
```
MattIPv4, line 14: Out of curiosity, why declare a const and then export it, rather than export default like we do in a few other places?

pjeweb: So the eslint rule [import/no-anonymous-default-export](https://github.com/import-js/eslint-plugin-import/blob/main/docs/rules/no-anonymous-default-export.md) enabled through the next eslint config does not allow exporting an object literal without a name as default. So for config files where there is not a `defineConfig()` call, we have to assign it to a var first. I think this also works better for declaring the type with jsdocs. I felt like its cleaner if we do the same for the configs where the config goes through `defineConfig()` too, but i dont have strong feelings.

For the anonymous exports I kinda agree and understand the rule, cause it makes it easier to (visually) grep import and export. I do like named exports for that reason too. But again no strong feelings.

apps/website/.eslintrc.json
```diff
@@ -14,7 +17,8 @@
   "rules": {
     "@typescript-eslint/consistent-type-imports": "warn",
     "@typescript-eslint/no-unused-vars": [
-      "warn", // or "error"
+      "warn",
+      // or "error"
```
pjeweb, line 21: why does it do this?

apps/website/vitest.config.ts
MattIPv4: Could we update the tailwind + lint-staged configs to follow the same pattern for consistency?

pjeweb: Interesting that the rule does not complain about those. Yeah we can do that for sure

apps/website/build-scripts/video-loader.cjs
```diff
@@ -1,3 +1,4 @@
+// @ts-nocheck
```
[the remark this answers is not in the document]
MattIPv4: Ah okay, no worries for now 👍

MattIPv4 (approved):

PR #568, opened 2024-02-27
Give an Hour: Fix date display (using local timezone instead of UTC)

pjeweb:
## Describe your changes

Since we use the local timezone to check the tracking status using local time, we parse the start and end time without an explicit timezone (= local time). This PR will make sure, the date is displayed in the local time to prevent a different date from being shown. By default UTC is used to format the date so we need to set it to null (= use user time zone).

- set local timezone to something > UTC and check the correct start date is shown (03/01)
- set local timezone to something < UTC and check the correct end date is shown (04/22)


MattIPv4 (approved):
Approved from EST

PR #562, opened 2024-02-26
Hide video in footer from known crawlers

MattIPv4:
## Describe your changes

As part of my continued effort to get Google to like the AQ pages more, I have observed in search console that Google seems to be seeing the footer video on _all_ our pages and (correctly) noting that it isn't the main content and so won't be indexed...

However, on the AQ pages where we also have a bunch of meta data pointing to the Twitch player as a legitimate video on the page, Google is reporting the footer video there and seems to be implying it thinks it is all the same, and so is not indexing video content on the page because the video it is detecting on the page is the footer video which is not the main content.

Hopefully, this change will completely remove the footer video from the DOM for Google, so it can't index it and get confused.

## Notes for testing your change

Override UA to be a Googlebot UA. Observe Twitch player load by default on AQ page, observe no video in footer.

PR #570, opened 2024-02-29
Move video to top of AQ pages as the primary content

MattIPv4:
## Describe your changes

It's me, back again with yet more changes to help convince Google to index the video content on the AQ pages (if this doesn't work, I'll probably give up).

Google now seems to be detecting the video content correctly after #562 was merged, but it reporting that it is considered to be "supplementary" content on the page, rather than primary content.

As such, this moves the video to the top of the page to make it clear that it is the primary content and that this is a "video page".

Layout is responsive across all breakpoints, and generally looks good.


pjeweb (approved):
i actually quite like it up there :)

MattIPv4:
Yeah, same, I think this is nicer.

The only change I might look at doing separately is removing the "meet the ambassadors" section, and instead have a vertical carousel on the right of the video with them. I didn't do it in this though as we don't have a vertical carousel built :P

PR #572, opened 2024-03-01
Set ambassador favicons on ambassador pages

MattIPv4:
## Describe your changes

cc https://github.com/alveusgg/data/pull/58

Some notes on the changes in Layout:

- We don't need a meta tag for `favicon.ico`, legacy browsers will look for that file anyway, and if we do have it, Chrome prefers it over the PNG favicon for reasons:tm:
- Setting multiple PNG favicon sizes is redundant, all modern browsers are fine with "any" size, and it makes setting it from Meta easier
- browserconfig + mstile icon were removed as I don't think many folks are using Windows versions with tiles anymore, or adding the Alveus site to bookmarks in the OS

On most pages on the site, the regular Alveus favicon shows.

On ambassadors pages where the ambassador has an icon in alveusgg/data, the ambassador icon shows as the favicon. On ambassador pages w/o an icon, the regular Alveus favicon shows.

Favicon switching works as you navigate across pages on the site.

Favicon switching works across Firefox + Chrome + Safari.


pjeweb (approved):
very cool idea

Issue #327, opened 2023-07-16
Add month schedule to updates page


pjeweb:
I think it would be great to have the monthly stream calendar on the updates page (pretty much !ceoschedule but only alveus stream + maybe regular weekly stream).

Easiest solution would be allowing to upload an image (or multiple) of the calendar. Even better would be an actual html calendar with all the dates. That would need some more backend work.

I dont think using the twitch stream schedule API would help much, also I do not really want to embed or link a google calendar.

prSquirrel:
> I dont think using the twitch stream schedule API would help much, also I do not really want to embed or link a google calendar.

Why so? Maybe i'm missing something, but it seems like the path of least resistance without having Alveus maintain extra things.

Going deeper into the html calendar idea. Is there any data source for these calendar images, or are they made from scratch every time? If Alveus has some internal calendar then we could query it in the backend (e.g. google calendar API, which is free tier) and render it. 

![image](https://github.com/alveusgg/alveusgg/assets/16546566/4148f604-f0ff-469f-933f-048f0a661701)

<br/>

Some further ideas which just popped into my mind:
- Export as image
- Export rendered grid / text only
  - Possible use case - use as a base for next calendar design, as an input to Photoshop for example
- Link to subscribe to the calendar feed
  - Probably covered by push notifications, but i would personally find it handy

pjeweb:
Welcome and thank you for looking into this issue!

There is no source calendar we could use as of yet.

The current workflow afaik is:

- maya plans streams on paper calendar for a month
- sends mik a photo of the calendar through discord
- mik creates the calendar in some graphics program (not sure what exactly)
- upload to imgur and update chat command
- when changes happen (probably announced on stream), mik updates the graphic and command

> Why so? Maybe i'm missing something, but it seems like the path of least resistance without having Alveus maintain extra things.
I don't think the twitch schedule is being used. But it might work as a backend. I am not sure if that gives us the flexibility needed though. Google calendar could maybe work too. I just do not know who would/could reliably maintain the schedule.

I can ask the team and mods if we can some shared calendar with api set up (probably google calendar). Also I think we still want the designed calendars in some sort going forward. Not sure how that can be combined. I like the idea of generating the grid/text and using that in the gfx program. I can ask mik if he an work with that. Maybe it could also work the other way around? Making the background and using it behind the html calendar?

[...]

PR #576, opened 2024-03-03
Embed schedule image/link on updates page

MattIPv4:
## Describe your changes

As an interim solution until #327 is implemented properly with the schedule data living in the DB, this PR adds an embedded image of the schedule short-link that was enabled with the work in #535.

If the schedule short-link does not exist, a no schedule available message is shown.

If the schedule short-link does exist, the imgur image is embedded via next/image.


MattIPv4:
I don't have access to shortlinks on preview.alveus.gg to add a link to that DB for testing.

pjeweb:
> I don't have access to shortlinks on preview.alveus.gg to add a link to that DB for testing.

You should have super user perms now

apps/website/src/components/notifications/Schedule.tsx
```diff
@@ -0,0 +1,31 @@
...
+
+  return (
+    <Link
+      href={`${getShortBaseUrl()}/l/${schedule.data.slug}`}
+      external
+      custom
+      className="group"
+    >
+      <Image
+        src={schedule.data.link}
```
pjeweb, line 23: Should we do a very basic check if its an imgur image link?

pjeweb (approved):

apps/website/src/components/notifications/Schedule.tsx
MattIPv4: I thought about doing that, but decided I trust Mik enough to stick to using imgur, and if not, I kinda want to see it error so we know to fix it.

PR #582, opened 2024-03-05
Show and Tell: Use proper post filter for volunteering minutes

pjeweb:
## Describe your changes

This fixes the volunteering minutes not counting previously approved posts that are currently pending review due to updates by the author.

Check the show and tell page shows the correct number of volunteering hours.


MattIPv4:
👀 What does getPostFilter generate, and how does that play with the indexes we have on the table for keeping this lookup efficient (not that we're really at a scale where that'll matter either way)

pjeweb:
> 👀 What does getPostFilter generate, and how does that play with the indexes we have on the table for keeping this lookup efficient (not that we're really at a scale where that'll matter either way)

Hmm. Its checking `approvedAt >= updatedAt` and we also check on `createdAt` so we do in fact not have a good index. Any suggestions for the index name? You suggested not calling it `_volunteeringMintues` but the only name I can come up with `_approvedAndCreatedAt` (or `_times`)?

MattIPv4 (approved):

PR #587, opened 2024-03-07
Add calendar to updates page + admin UI

### LINKED ISSUE #327: Add month schedule to updates page

### LINKED ISSUE #585: [Idea]: Remove "active" column from short links admin page

Since we do not have a way to disable short-links, we can drop the column.

short-links

Improve admin UI clutter and consistency, only showing the active column with buttons if they do something.

MattIPv4:
## Describe your changes

Resolves #327.

Also resolves #585.

Events can be added, edited, removed from the admin UI.

Users can clearly see events for the current month on the updates page, on mobile + desktop.

### DIFF
--- a/apps/website/prisma/schema.prisma
+++ b/apps/website/prisma/schema.prisma
@@ -300,6 +300,19 @@ model ShortLinksTracking {
   updatedAt DateTime @updatedAt
 }
 
+model CalendarEvent {
+  id          String @id @default(cuid())
+  title       String
+  description String
+  link        String
+  startAt     DateTime
+
+  createdAt DateTime @default(now())
+  updatedAt DateTime @updatedAt
+
+  @@index([startAt], name: "CalendarEvent_start")
+}
+
 model Bingo {
   id                 String    @id @default(cuid())
   type               String
--- a/apps/website/prisma/schema.prisma
+++ b/apps/website/prisma/schema.prisma
@@ -303,7 +303,7 @@ model ShortLinksTracking {
 model CalendarEvent {
   id          String @id @default(cuid())
   title       String
-  description String
+  description String?
   link        String
   startAt     DateTime
 
--- a/apps/website/package.json
+++ b/apps/website/package.json
@@ -79,7 +79,6 @@
     "slugify": "^1.6.6",
     "smooth-scroll-into-view-if-needed": "^2.0.2",
     "superjson": "^2.2.1",
-    "tailwindcss-text-fill-stroke": "2.0.0-beta.1",
     "zod": "^3.22.4"
   },
   "devDependencies": {
--- a/apps/website/package.json
+++ b/apps/website/package.json
@@ -79,6 +79,7 @@
     "slugify": "^1.6.6",
     "smooth-scroll-into-view-if-needed": "^2.0.2",
     "superjson": "^2.2.1",
+    "tailwindcss-text-fill-stroke": "2.0.0-beta.1",
     "zod": "^3.22.4"
   },
   "devDependencies": {
--- a/apps/website/src/components/admin/calendar-events/CalendarEventForm.tsx
+++ b/apps/website/src/components/admin/calendar-events/CalendarEventForm.tsx
@@ -1,0 +1,129 @@
+import type { FormEvent } from "react";
+import { useCallback } from "react";
+
+import type { CalendarEvent } from "@prisma/client";
+import { useRouter } from "next/router";
+
+import { trpc } from "@/utils/trpc";
+
+import type { CalendarEventSchema } from "@/server/db/calendar-events";
+
+import { Button, defaultButtonClasses } from "@/components/shared/Button";
+import { TextField } from "@/components/shared/form/TextField";
+import { Fieldset } from "@/components/shared/form/Fieldset";
+import { MessageBox } from "@/components/shared/MessageBox";
+import {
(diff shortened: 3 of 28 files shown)


pjeweb (commented):
Love it! A few small things I noticed on my first review:
- paint order does not quite yet work on chrome/edge so the text looks like a big black blob https://caniuse.com/mdn-css_properties_paint-order
- would it make sense to actually use a <table/> with proper semantics? you should still be able to change the css layout to a grid etc. just needs some nested loops for the <tr/>s. otherwise we could consider using aria roles? https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/table_role
- but it might actually be good to actually use the table layout for cases where a lot of events / long descriptions are stacked on one day?
grid
- can we drop the horizontal padding around the calendar, aligning it with the headlines and announcements?
- on mobile I think the monday headings are a bit confusing. at least it was for me at first. what do you think about having the week day in each cell?
cal-alt

I also have a few ideas but we ca think about them later:

-  can we allow to changing to the next month (if there events entered for that month)
- show the current date / fade out past days
- for the admin ui maybe we can offer suggestions for the common links? it does not show as nice on FF but it might help entering them quicker https://github.com/alveusgg/alveusgg/blob/f4e137cd98d53d1b136c698b2a2bf3cf0785efdd/apps/website/src/components/admin/notifications/SendNotificationForm.tsx#L148-L166
suggest

MattIPv4:
> paint order does not quite yet work on chrome/edge so the text looks like a big black blob https://caniuse.com/mdn-css_properties_paint-order

Not sure what a good solution is to this, thoughts? I suppose we could use the text-shadow hack to achieve a stroke instead?

> would it make sense to actually use a with proper semantics? you should still be able to change the css layout to a grid etc. just needs some nested loops for the s. otherwise we could consider using aria roles? https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/table_role
> but it might actually be good to actually use the table layout for cases where a lot of events / long descriptions are stacked on one day?

The issue of all the rows matching the height of the tallest is just due to the auto-rows, I can remove that. I will play around locally with switching it to an actual table, just need to figure out grouping the days by week.

> can we drop the horizontal padding around the calendar, aligning it with the headlines and announcements?
> on mobile I think the monday headings are a bit confusing. at least it was for me at first. what do you think about having the week day in each cell?

Sure, and yeah, I think that works better. Will get those changes in.

> can we allow to changing to the next month (if there events entered for that month)

The API allows for this (well, querying any date range at all). I was intending to create a follow-up issue to actually add this functionality though, once this was in.

> show the current date / fade out past days

Yah, I can look at doing a subtle indication of this with the date number.

> for the admin ui maybe we can offer suggestions for the common links? it does not show as nice on FF but it might help entering them quicker

Might be something to create a follow-up item for, though once you've submitted a link once the browser should provide auto-complete suggestions anyway?

apps/website/prisma/schema.prisma
```diff
@@ -300,6 +300,19 @@ model ShortLinksTracking {
   updatedAt DateTime @updatedAt
 }
 
+model CalendarEvent {
+  id          String @id @default(cuid())
+  title       String
+  description String
```
pjeweb, line 306: I do not think we show this anywhere, do we?

pjeweb (commented):
- Should we add a legend?
- We should add a disclaimer that "times/dates are subject to change"
- We probably should allow events without time (only date) for scheduled YT videos etc.
- I do not think we use the description anywhere. Can we make it optional and/or actually render it?
- There are some problems with time zones. I'll try to figure them out, but my brain is not capable right now. I also tried to add a toggle to show in Alveus time (CT) _or_ local time.

I also tried a few things to emulate the text stroke in Chrome/Edge. Im not really happy with any of them. Maybe can we use a background color instead? 

bg

bg-mobile

MattIPv4:
> Should we add a legend?

I thought about that, but figured with the links it was less of an issue compared to the image. The colors here just feel more like grouping?

> We should add a disclaimer that "times/dates are subject to change"

Will add.

> We probably should allow events without time (only date) for scheduled YT videos etc.

I thought about this as well, but decided not to, as I think it would make it rather complicated given we're showing the calendar in the user's local time. With the disclaimer, I think it'd be fine for such items to just have an estimated time added, like midday or w/e.

> I do not think we use the description anywhere. Can we make it optional and/or actually render it?

Making it optional sounds good. I didn't have a good usage for it right now, but figured we might want to build a richer experience down the road, maybe with a popup for each event that shows the link + description etc.

> There are some problems with time zones. I'll try to figure them out, but my brain is not capable right now. I also tried to add a toggle to show in Alveus time (CT) or local time.

From Discord, it sounds like we figured this out, I just need to change how I'm grouping the events by day? As for a local/Alveus time toggle, that sounds like a good follow-up issue to create?

> I also tried a few things to emulate the text stroke in Chrome/Edge. Im not really happy with any of them. Maybe can we use a background color instead?

Looks clean enough, will switch.

pjeweb (approved):
great work

PR #593, opened 2024-03-10
Fix user roles being cached across requests

pjeweb:
## Describe your changes

Instead of caching the user role lookup in a global cache object (map), we look it up on every request when the user session object is created and access the roles from there.

This fixes an issue where permissions/roles would not update for users because of the cache.

Check permission checks still work (positive and negative) and updates are applied immediately (on navigation).

### DIFF
--- a/apps/website/src/server/db/users.ts
+++ b/apps/website/src/server/db/users.ts
@@ -1,20 +1,10 @@
 import { prisma } from "@/server/db/client";
 
-const cache = new Map<string, string[]>();
-
 export async function getRolesForUser(userId: string) {
-  const cachedRoles = cache.get(userId);
-  if (cachedRoles) {
-    return cachedRoles;
-  }
-
-  const roles = (
+  return (
     await prisma.userRole.findMany({
+      select: { role: true },
       where: { user: { id: userId } },
     })
-  ).map((role) => role.role);
-
-  cache.set(userId, roles);
-
-  return roles;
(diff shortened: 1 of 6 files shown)


pjeweb:
Also deployed to preview.alveus.gg now

apps/website/src/server/db/users.ts
```diff
@@ -1,20 +1,10 @@
...
 export async function getRolesForUser(userId: string) {
-  const cachedRoles = cache.get(userId);
-  if (cachedRoles) {
-    return cachedRoles;
-  }
-
-  const roles = (
+  return (
     await prisma.userRole.findMany({
+      select: { role: true },
```
MattIPv4, line 6: What does this do? Just returning the role column only?

apps/website/src/server/utils/auth.ts
```diff
@@ -23,21 +22,19 @@ export function checkIsSuperUserSession(session: Session | null) {
 
 export async function checkPermissions(
```
MattIPv4, line 23: This doesn't need to be async now?

apps/website/src/server/db/users.ts
pjeweb: yeah. does not really make any difference, but its `select role` instead of `select *` since we only need the role

MattIPv4 (approved):

PR #595, opened 2024-03-10
Fix schedule missing some events for clients with time zones > UTC

pjeweb:
## Describe your changes

Instead of calculating the end of the month on the server in UTC, calculate it on the client in the display time zone.

This PR also changes the time display in the client to use local format (`20:00 CET` instead of `8:00 PM GMT+1`) and prevents double fetches when date calculations are incomplete (one effect-loop).

Test events show properly with different client and server timezones. Good luck.

### DIFF
--- a/apps/website/src/components/notifications/Schedule.tsx
+++ b/apps/website/src/components/notifications/Schedule.tsx
@@ -78,10 +78,10 @@ export function Schedule() {
     [startOfMonth],
   );
 
-  const events = trpc.calendarEvents.getCalendarEvents.useQuery({
-    start: startOfMonth,
-    end: endOfMonth,
-  });
+  const events = trpc.calendarEvents.getCalendarEvents.useQuery(
+    { start: startOfMonth, end: endOfMonth },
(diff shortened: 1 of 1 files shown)


apps/website/src/components/notifications/Schedule.tsx
```diff
@@ -64,20 +64,24 @@ export function Schedule() {
...
     () =>
       startOfMonth &&
       new Date(
         startOfMonth.getFullYear(),
         startOfMonth.getMonth() + 1,
         0,
-      ).getDate(),
+        23,
+        59,
+        59,
```
MattIPv4, line 76: The logic in the API does a `<`, so do we want this to be `1, 0, 0, 0` rather than `0, 23, 59, 59`? (The `0,0,0` wouldn't be needed)

pjeweb: ah yeah

MattIPv4:
With that last commit, I think places that're doing endOfMonth.getDate() will need a -1 now?

MattIPv4 (approved):

PR #596, opened 2024-03-11
Allow schedule month to be changed + use calendar view in admin UI

### LINKED ISSUE #590: [Idea]: Allow updates calendar to change month
can we allow to changing to the next month (if there events entered for that month)

_Originally posted by @pjeweb in https://github.com/alveusgg/alveusgg/pull/587#pullrequestreview-1923579954_

### LINKED ISSUE #594: [Idea]: Abstract calendar rendering logic, use on admin UI

Abstract the calendar rendering logic currently in the schedule, so it can be reused for the admin UI. Instead of the events linking out to their respective links, they'd link to the edit page. The edit page would then also contain the delete button.

Make the admin UI for events much easier to work with.

This is dependent on #590 being completed, so previous/future events in the admin UI can be controlled.

MattIPv4:
## Describe your changes

Resolves #590 

Resolves #594 

Swapping months works as expected, no matter what timezone you're in. Admin UI behaves similarly to the public UI.

### DIFF
--- a/apps/website/src/components/admin/calendar-events/CalendarEventForm.tsx
+++ b/apps/website/src/components/admin/calendar-events/CalendarEventForm.tsx
@@ -5,6 +5,11 @@ import type { CalendarEvent } from "@prisma/client";
 import { useRouter } from "next/router";
 
 import { trpc } from "@/utils/trpc";
+import { classes } from "@/utils/classes";
+import {
+  inputValueDatetimeLocalToUtc,
+  utcToInputValueDatetimeLocal,
+} from "@/utils/local-datetime";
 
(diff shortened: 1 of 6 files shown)


pjeweb (approved):
pogg

not necessary, but two minor things that can again be their own PR if you like:
- can we render the header and maybe a calendar skeleton while loading events? that would make the navigation a bit smoother
- would it be possible to move the create form for the events in to below the preview calendar? would make it easier to add multiple entries while seeing the calendar

pjeweb:
Oh and is it on purpose that you left out the colors in the backend? i think having the colors there too would probably be better?

pjeweb (approved):
very cool

PR #605, opened 2024-03-17
Add color themes to Calendar for each month

MattIPv4:
## Describe your changes

Following the themes that Mik had in the old image-based schedules, this adds themes to each month in the Calendar.

Days + events are visible on all desktop/mobile across all the theme backgrounds.


MattIPv4:
My only question was should I put this behind a `themed` prop on `Calendar`, so the admin UI uses the default theme always? Or are we fine w/ the admin UI using the theme too?

pjeweb:
> My only question was should I put this behind a `themed` prop on `Calendar`, so the admin UI uses the default theme always? Or are we fine w/ the admin UI using the theme too?

I think its fine or even better having the same design there.

pjeweb (approved):
very nice

PR #615, opened 2024-03-29
Add categories to calendar events to determine color

### LINKED ISSUE #611: [Idea]: Replace Calendar event color logic w/ category selection

Instead of driving the event colors based on their URL, allow for a category to be set for each event which will dictate the color.

Categories that should be supported:

- Maya streams
- Maya YT videos
- Alveus regular streams
- Alveus special streams
- Alveus YT videos
- Collaborations

Resolves the current issue where collaboration events with streamers on YT are showing up in the same color as regular YT content.

MattIPv4:
## Describe your changes

Resolves #611.

An admin can provide any category name they desire but are provided with default recommendations which map to known colors.

This adds a new column to the DB that is required, but I have not set a default as I do not wish for there to be a default going forward. When this schema change is pushed to prod, please add a default of "Alveus Regular Stream" for the initial push, and then do a secondary push without the default.

Suggested categories work as expected, resulting in a color.


pjeweb (dismissed):
Only thing I'd say is, having a fixed list of categories and a "normal" select box would probably be easier to use (small typos might make it hard to see why a color does not work, resetting is cumbersome etc.).

I assume you wanted to future proof if we need new categories? Since we don't show it besides the color, does that really matter though? We'd need to change code for new categories to do something anyways then, right?

It could still be a string in the schema (instead of an enum) to keep that future proof.

PS: Would it be the better UX to reset the form on submit?

pjeweb (approved):
LGTM
Database is prepared!

Issue #81, opened 2023-03-04
Provide local dev stack


pjeweb:
- [ ] provide a docker compose or similar with required services 
  - [x] database => mysql
  - [x] s3 compatible file storage => minio/localstack
  - [ ] cronjob => ?

MattIPv4:
@pjeweb do we need a local replacement for crons? You can always just hit `localhost:3000/api/cron` to trigger it manually I believe. If we do, we could use https://www.npmjs.com/package/vercel-cron, but it seems excessive?

pjeweb:
Agreed. Not necessary.

PR #616, opened 2024-03-29
Provide local dev DB + S3 bucket with Docker Compose

### LINKED ISSUE #586: [Idea]: Provide alternative to Planetscale for (free) dev setup

Since Planetscale will drop the free plan in a month, I suggest we find another (free) suggestion for contributors. Could be a local db (#81) or maybe we try to migrate to Vercel's Postgres aka neon which is (still) free.

Do not make contributing and local development harder.

MattIPv4:
## Describe your changes

Resolves #586 (and partially #81)

Adds a local MySQL DB that can be spun up with docker-compose, as well as a minio instance to provide a local S3 bucket.

MinIO requires the S3 logic use path-style rather than hostname-style, ensure that this does not break S3 usage with other providers like DigitalOcean Spaces that is used in production.


README.md
```diff
@@ -70,14 +71,12 @@ Hey there! Welcome to Alveus.gg! There's a few ways that you can help contribute
 
 1. Install Node.js (see `engines` in `package.json` for the required versions), or use `fnm`/`nvm` to install the correct version of Node.js, and use `corepack enable` to use PNPM.
 2. Install dependencies: `pnpm install --frozen-lockfile`
-3. Create a [PlanetScale](https://planetscale.com/) account (free) or provide your own MySQL server, that should give you two DSN for the main and shadow database (something like `mysql://user:pass@us-east.connect.psdb.cloud/alveusgg?sslaccept=strict` and `mysql://user:pass@us-east.connect.psdb.cloud/alveusgg/shadow?sslaccept=strict`)
+3. Run `docker-compose up -d` from within `apps/website` to start a local MySQL database, and an S3 bucket with MinIO.
```
pjeweb, line 74: Any reason we recommend the older `docker-compose` over `docker compose`?

apps/website/src/server/utils/file-storage.ts
```diff
@@ -12,9 +12,14 @@ import { env } from "@/env";
 import { prisma } from "@/server/db/client";
 import { probeImageMeta } from "@/server/utils/probe-image-meta";
 
-function getBucketUrl() {
+const pathStyle = true; // MinIO only works with path style
```
pjeweb, line 15: I mean this could easily be an env variable `FILE_STORAGE_PATH_STYLE=true` even if DO works with path style?

pjeweb (commented):
Tested locally and on preview with DO. Working great. So much better!

README.md
MattIPv4: Habit lol, will swap

pjeweb (approved):
very nice

PR #620, opened 2024-03-30
Abstract button styling into dedicated component

MattIPv4:
## Describe your changes

Abstracts the button styling we use throughout the site into a dedicated component.

I realised we also had a "shared" button component that is used throughout our forms, so I moved that into the forms subdirectory for clarity.

All buttons using the standard styling are now using the component.


pjeweb (approved):
very nice

PR #625, opened 2024-03-30
Fix Animal Quest CTA responsiveness on ambassador pages

MattIPv4:
## Describe your changes

Using multiple inline-block spans should encourage the text to break where we want it to (as was the aim with the nowrap), but still allowing it to wrap further when it needs to (unlike before).

All Animal Quest CTAs wrap correctly across mobile/tablet/desktop viewports.

| Before | After |
|-|-|
| image | image |


pjeweb (approved):
lgtm

PR #632, opened 2024-04-03
Redirect admin pages to login for unauth'ed users

### LINKED ISSUE #567: [Idea]: Redirect 404's in /admin/ to the login

I purposely made admin links show the default 404 page for "security" reasons. I do now think its way better UX if we can use deep links to admin features and admins being asked to log in and since its all open source its not really a big security feature anyways.

It could also be a "not authorized" page with a login button.

Allow deep linking to admin features without showing an error page if the users are not logged in.

JonasFranke:
## Describe your changes

Resolves #567 

Since I fucked the first try of this PR with git rebase is here a cleaned up version.


apps/website/src/pages/admin/show-and-tell/index.tsx
```diff
@@ -8,8 +8,18 @@ import { permissions } from "@/data/permissions";
 import Meta from "@/components/content/Meta";
 
 export async function getServerSideProps(context: NextPageContext) {
+  const session = await getSession(context);
```
pjeweb, line 11: This file is missing the import.

pjeweb (changes requested):

apps/website/src/pages/admin/unauthorized.tsx
```diff
@@ -0,0 +1,28 @@
+import type { NextPage } from "next";
+
+import Meta from "@/components/content/Meta";
+import Section from "@/components/content/Section";
+import Heading from "@/components/content/Heading";
+
+const UserUnauthorizedPage: NextPage<SigninPageProps> = () => {
```
pjeweb, line 7: The props type is not needed.

```suggestion
const UserUnauthorizedPage: NextPage = () => {
```

apps/website/src/pages/admin/unauthorized.tsx
```diff
@@ -0,0 +1,28 @@
...
+  return (
+    <>
+      <Meta
+        title="Unauthorized"
+        description="You are not authorized to view this page."
+      />
```
pjeweb, line 13: I'd say we should noindex this page and then we also don't need a description:

```suggestion
      <Meta title="Unauthorized" noindex />
```

apps/website/src/pages/admin/show-and-tell/index.tsx
```diff
@@ -6,10 +6,21 @@ import { AdminPageLayout } from "@/components/admin/AdminPageLayout";
 import { AdminShowAndTellEntriesPanel } from "@/components/admin/show-and-tell/AdminShowAndTellEntriesPanel";
 import { permissions } from "@/data/permissions";
 import Meta from "@/components/content/Meta";
+import { getSession } from "next-auth/react";
```
pjeweb, line 9: Could you move this up below the next import?

apps/website/src/pages/admin/twitch/index.tsx
```diff
@@ -14,8 +15,18 @@ import { ProvideAuth } from "@/components/admin/twitch/ProvideAuth";
 import { ChannelConfig } from "@/components/admin/twitch/ChannelConfig";
 
 export async function getServerSideProps(context: NextPageContext) {
+  const session = await getSession(context);
   const adminProps = await getAdminSSP(context, permissions.manageTwitchApi);
-  return adminProps ? { props: adminProps } : { notFound: true };
+  if (!adminProps) {
+    return {
+      redirect: {
+        destination: session?.user?.id ? "unauthorized" : "/auth/signin",
```
pjeweb, line 23: Tiny nit, but can we use the same absolute path here too?

```suggestion
        destination: session?.user?.id ? "/admin/unauthorized" : "/auth/signin",
```

apps/website/src/pages/admin/unauthorized.tsx
pjeweb, line 1: I'm not 100% sure, but I think this should either use the admin layout (dark gray background, admin menu) or moved to pages/unauthorized since it does not have the admin layout!?

pjeweb (changes requested):
Thank you for continuing the work!

Biggest thing i'd like is that you get redirected backl to the initial page after logging in. I think it should work to pass the origin url as a query param to the signin page (eg. `/auth/sigin?callbackUrl=/admin/activity-feed`). The rest should already be set up.
https://next-auth.js.org/getting-started/client#specifying-a-callbackurl

Also you can try to `pnpm lint` and `pnpm build` locally to catch some of the errors before we approve the deployment.

apps/website/src/pages/admin/unauthorized.tsx
JonasFranke: > I'm not 100% sure, but I think this should either use the admin layout (dark gray background, admin menu) or moved to pages/unauthorized since it does not have the admin layout!?

I think the layout should stay the same (and the page should be moved to pages/unauthorized) to separate the admin and the regular pages

apps/website/src/pages/admin/bingo/[bingoId]/live.tsx
```diff
@@ -35,9 +36,18 @@ import { BingoCardGrid } from "@/components/bingo/BingoCardGrid";
...
+  const session = await getSession(context);
   const adminProps = await getAdminSSP(context, permissions.manageBingos);
   if (!adminProps) {
-    return { notFound: true };
+    const id = context.params?.bingoId;
+    return {
+      redirect: {
+        destination: session?.user?.id
+          ? "/unauthorized"
+          : `/auth/signin?callbackUrl=/admin/bingo${id ? "/" + id + "/live" : ""}`,
```
Comment (pjeweb) [inline apps/website/src/pages/admin/bingo/[bingoId]/live.tsx:47]: I think it would be cleaner and more maintainable to use the context object to get the path (same for all routes probably). We could also try to refactor further so we do not have the same code in each route, but im happy with being explicit here.

```suggestion
          : `/auth/signin?callbackUrl=${encodeURIComponent(context.resolvedUrl)}`,
```

pjeweb (changes requested):
Very cool. Now it really feels a lot better in terms of UX!

pjeweb (approved):
Thank you!

PR #649, opened 2024-04-07
Sort creator selector by popularity

### LINKED ISSUE #427: [Idea]: Make collabs sortable

MattIPv4:
## Describe your changes

Adds a popularity value for each creator (using their most popular platform for the most part), and sorts the creator selector based on that. This should help increase visibility to the more popular creators for ease-of-use on the page, while still preserving the chronological order of the collabs themselves.

I think at this point, this resolves #427 in lieu of using the regular sort-by/group-by functionality.

Values all seem correct.


apps/website/src/pages/collaborations.tsx
```diff
@@ -235,13 +236,25 @@ const CollaborationsPage: NextPage = () => {
...
               We work with other content creators to educate our combined
               audiences, introducing them to the educational ambassadors at
               Alveus and their conservation missions.
             </p>
+
+            <p className="mt-2 text-balance text-lg">
+              We&apos;ve hosted{" "}
+              <abbr
+                title={`Across ${collaborations.length.toLocaleString()} collaboration streams`}
+              >
```
MattIPv4, line 251: This feels like a misuse of abbr, but it does exactly what I want here?

pjeweb (approved):

apps/website/src/pages/collaborations.tsx
pjeweb: Yeah i think this is fine-ish. If we wanna be even more accessible, relying on the title attribute isn't gonna be the best, but i dont think its a big deal here.

pjeweb (approved):

PR #651, opened 2024-04-07
Clarify Show and Tell alt text input

### LINKED ISSUE #566: [Idea]: Show and Tell: Make image alt text more clear

I tried to make it clear the description is only show to screen readers, but we still get submissions that put text there that should be accessible to all users, not just as an alternative text describing the image.

sat-alt

I propose finding a better description/UI to make that clear. Maybe use "alternative text" or add "e.g. for blind users"?

Prevent incorrect entries, improve UX, make reviews easier.

MattIPv4:
## Describe your changes

Resolves #566.

Wording is clear.


pjeweb (approved):
thank you

PR #652, opened 2024-04-11
Add collab for Eret & Ava

MattIPv4:
## Describe your changes

Adds the most recent collaboration, Eret + Ava

Video is correct, date is correct, names/link are correct


MattIPv4:
Will get #649 rebased once this is in.

PR #653, opened 2024-04-12
Fix image attachment schema not allowing enough chars

### LINKED ISSUE #606: [Bug]: Errors Submitting to Show and Tell

After filling out all the fields and uploading 9 photos with captions, I kept getting this error message: "Invalid `prisma.imageAttachment.create()` invocation: The provided value for the column is too long for the column's type. Column: for".

After refreshing the page after a timeout/expiration, I checked for any special characters and looked other posts for any formatting or special character issues I might have had, and edited down the captions and got another error: "Unique constraint failed on the (not available)" (I forgot to copy the first part of the error message, I think it was similar to the above).

Then I tried with 7 photos instead and got the first error again (pictured below).

Alveus Sanctuary - Show and Tell - Submit Post Error

After cropping a wider photo and editing the captions again, I got this error: "Error uploading file: Invalid image".

When I click "submit for review", my form should be uploaded or give me a clear error message to fix my submission for upload/approval.

### Steps to reproduce

- Reload the page
- Click "Continue anonymously"
- Reattach the same photos and refill all fields with the exact same content, or reattach fewer/smaller photos and refill all fields with shorter content
- Click "submit for review"

### Browser specs

whatsmybrowser.org/b/WEX26

### Anything else?

Maybe there could be a simple one-click bug report button after a failed submission so more users can report errors without needing to use Github. However, these errors (or at least several of them) don't seem that common judging by the number of successful and approved submissions on the Show and Tell page, including ones from the last few days.

pjeweb:
## Describe your changes

There currently is a mismatch between the database schema (strings of 191 chars - the default for VARCHAR = String in MySQL/Prisma) and the max length of input fields (caption and alt text). This commit alters the table to allow the same number of characters as the input fields to prevent errors.

Resolves #606

Test show and tell post submission. Attach image(s) and fill the fields with lots of chars :)

### DIFF
--- a/apps/website/prisma/schema.prisma
+++ b/apps/website/prisma/schema.prisma
@@ -301,7 +301,7 @@ model ShortLinksTracking {
 }
 
 model CalendarEvent {
-  id          String @id @default(cuid())
+  id          String   @id @default(cuid())
   title       String
   description String?
   category    String
@@ -453,12 +453,12 @@ model ImageMetadata {
 
 model ImageAttachment {
   id                  String @id @default(cuid())
-  name                String
-  title               String
-  alternativeText     String
-  caption             String
-  description         String
-  url                 String
+  name                String @db.VarChar(200)
+  title               String @db.VarChar(200)
+  alternativeText     String @db.VarChar(300)
+  caption             String @db.VarChar(200)
+  description         String @db.VarChar(200)
+  url                 String @db.VarChar(200)
   fileStorageObjectId String @unique
 
   fileStorageObject          FileStorageObject           @relation(fields: [fileStorageObjectId], references: [id], onDelete: Cascade)


pjeweb:
This is already deployed to the production database. It is still broke on preview. I messed that up, but it works :)

apps/website/prisma/schema.prisma
```diff
@@ -301,7 +301,7 @@ model ShortLinksTracking {
 }
 
 model CalendarEvent {
-  id          String @id @default(cuid())
+  id          String   @id @default(cuid())
```
pjeweb, line 304: I guess the IDE wanted this indented. Can we keep it like this?

MattIPv4: Yeah, seems right, its adding the extra space because of the `?` on one of the types further down in this model

MattIPv4 (approved):

apps/website/prisma/schema.prisma
pjeweb: Its aligned with `DateTime` even further down. Is it worth running `pnpm prisma format` in lint staged?

MattIPv4: Yeah sounds worth adding to lint-staged and the package scripts, in a follow-up pr

Issue #656, opened 2024-04-16
[Idea]: Add prisma format to lint-staged and package scripts


pjeweb:
### Describe your idea / enhancement

We should add `(pnpm) prisma format` to lint-staged and package scripts to make sure the prisma schema file is formatted correctly.

### How would this idea / enhancement be beneficial?

Prevent style discussions and messy diffs in changes.

PR #661, opened 2024-04-20
Add prisma format to lint-staged

bl4nkfce:
## Describe your changes

- Added `prisma format` to lint-staged
- Updated `pnpm format` to also run `prisma format`

Resolves https://github.com/alveusgg/alveusgg/issues/656

### DIFF
--- a/package.json
+++ b/package.json
@@ -28,7 +28,7 @@
     "prepare": "husky",
     "lint-staged": "lint-staged --relative",
     "lint": "prettier . --check",
-    "format": "prettier . --write"
+    "format": "prettier . --write && pnpm dlx prisma format --schema apps/website/prisma/schema.prisma"
   },
   "pnpm": {
     "patchedDependencies": {
--- a/package.json
+++ b/package.json
@@ -28,7 +28,7 @@
     "prepare": "husky",
     "lint-staged": "lint-staged --relative",
     "lint": "prettier . --check",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -28,7 +28,7 @@
     "prepare": "husky",
     "lint-staged": "lint-staged --relative",
     "lint": "prettier . --check",
-    "format": "prettier . --write"
+    "format": "prettier . --write && pnpm dlx prisma format --schema apps/website/prisma/schema.prisma"
```
MattIPv4, line 31: I'm not sure I like the idea of calling a prisma command from the root package, rather than from the package where the file + dependency live. Can we move this into apps/website/package.json?

bl4nkfce: Yeah I can move it to the website package.

Should It still be possible to call it from root though?

I think that having it as a script in the website package and running it in the pre-commit hook should be enough.

MattIPv4: pnpm has a recursive flag, so running the root format can run the format command in each workspace too

kkolasinski: I've moved it to the website package like you suggested

bl4nkfce: I've moved it like you suggested

MattIPv4 (approved):
This seems fine to me. @pjeweb is this what you had in mind?

pjeweb (approved):

pjeweb:
> This seems fine to me. @pjeweb is this what you had in mind?

Yeah this works great! Thank you @bl4nkfce !

PR #665, opened 2024-04-24
Show post/user count on Show and Tell

### LINKED ISSUE #659: [Idea]: Show some stats about Show & Tell posts

Show how many posts have been submitted by how many users/names.

Both in the front end for visitors and maybe also in the admin area (maybe even more info there?).

Show community engagement.

sougataghar477:
## Describe your changes

Shows the total approved post count, and the total user count (accounting for auth'ed users and anon users), for Show and Tell.

Resolves #659

Counts are correct.


apps/website/src/pages/show-and-tell/index.tsx
```diff
@@ -391,6 +401,9 @@ const ShowAndTellIndexPage: NextPage<ShowAndTellPageProps> = ({
           onKeyDown={handleArrowKeys}
           tabIndex={-1}
         >
+          <h1 className="text-center text-3xl italic">
+            {totalPostsCount} posts by {usersCount} users
+          </h1>
```
MattIPv4, line 406:
```suggestion
          <p className="text-center text-3xl italic">
            {totalPostsCount} posts by {usersCount} users
          </p>
```
Semantically, I don't think this should be an h1, it is not the primary heading for the entire page. I don't think it should be a semantic heading at all.

apps/website/src/pages/show-and-tell/index.tsx
```diff
@@ -391,6 +401,9 @@ const ShowAndTellIndexPage: NextPage<ShowAndTellPageProps> = ({
           onKeyDown={handleArrowKeys}
           tabIndex={-1}
         >
+          <h1 className="text-center text-3xl italic">
+            {totalPostsCount} posts by {usersCount} users
```
MattIPv4, line 405:
```suggestion
            {totalPostsCount.toLocaleString()} posts by {usersCount.toLocaleString()} users
```
We should locale format both the numbers here, so they get thousands separators etc.

MattIPv4, line 405:
```suggestion
            {totalPostsCount} posts by {usersCount} community members
```
I think it'd be nicer to refer to users as community members here

MattIPv4 (commented):
Some initial thoughts

apps/website/src/server/db/show-and-tell.ts
```diff
@@ -243,6 +243,29 @@ export async function getPosts({
...
+export async function getUsersCount() {
+  const userCountwithId = await prisma.showAndTellEntry.findMany({
+    select: { id: true },
+    where: {
+      userId: {
+        not: null,
+      },
+    },
+    distinct: ["userId"],
+  });
```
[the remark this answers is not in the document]
pjeweb: Weirdly Prisma does not support aggregates with distinct: https://github.com/prisma/prisma/issues/4228

I do not think we can prevent creating objects for each row if we wanna use the ORM. What I suggested was using `select: {}` to save some bytes.

apps/website/src/pages/show-and-tell/index.tsx
pjeweb: I think that would break SSR for different locale.

apps/website/src/server/db/show-and-tell.ts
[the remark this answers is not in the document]
MattIPv4: Oh, that sucks. `select: {}` sounds fine then. I can't imagine this table will ever get big enough that this is truly a problem (and if it is, we can look at doing a raw query at that point)

apps/website/src/pages/show-and-tell/index.tsx
MattIPv4: What's the recommended way to do that, that doesn't break SSR? useMemo them?

apps/website/src/server/db/show-and-tell.ts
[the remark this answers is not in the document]
pjeweb: Nope that errors too `The `select` statement for type ShowAndTellEntry must not be empty.`

apps/website/src/pages/show-and-tell/index.tsx
pjeweb: For the pages router and SSR I'd say `useState` + `useEffect`, yeah. I don't think you can use client components with `"use client"` in the pages router. It's kinda a special case since we do not have the language/locale in the router or server-side at all.

Something like:

```typescript
function useLocale() {
  const [locale, setLocale] = useState('en-US'); // default for initial render
  useEffect(() => {
    setLocale(undefined); // use browser default locale
  }, []);
  return locale;
}

const locale = useLocale();
value.toLocaleString(locale);
```

pjeweb: or honestly just use US formatting

MattIPv4:
@sougataghar477 it looks like you've got a conflict that needs resolving, and one outstanding comment.

MattIPv4 (approved):

PR #667, opened 2024-04-26
Show and Tell: Permanently Enable Tracking Volunteering Hours

pjeweb:
## Describe your changes

As decided by the executive director, Alveus will continue the Give an Hour initiative beyond the official WWF campaign. This PR removes the restrictions on volunteering hours and adapts the wording.

This will also reduce the number of characters allowed for new posts as requested by the CEO to keep the posts short for stream :)

I am very unsure on the wording and if we keep the WWF logo, but I think this makes it clear its not part of the official campaign anymore without needing to replace all the original branding.

- Check existing posts are still fully readable and editable
- Check new posts can be submitted, reviewed and seen incl. volunteering hours (or no hours)
- Check wording is fine

### DIFF
--- a/apps/website/src/components/show-and-tell/ShowAndTellEntryForm.tsx
+++ b/apps/website/src/components/show-and-tell/ShowAndTellEntryForm.tsx
@@ -353,14 +353,15 @@ export function ShowAndTellEntryForm({
       </div>
 
       <div className="space-y-10">
-        <Fieldset legend="Give an Hour">
+        <Fieldset legend="Give an hour for earth">
           <p>
             Do you want to track hours you spent on this activity as part of the
-            Alveus community total for WWF&apos;s{" "}
+            Alveus communities effort to give an hour for earth? This started as
+            part of WWF&apos;s{" "}
             <Link href="/show-and-tell/give-an-hour" external>
               Give an Hour
             </Link>{" "}
-            initiative?
+            initiative.
           </p>
 
           <div className="flex items-center gap-8">
--- a/apps/website/src/components/show-and-tell/ShowAndTellEntryForm.tsx
+++ b/apps/website/src/components/show-and-tell/ShowAndTellEntryForm.tsx
@@ -362,7 +362,7 @@ export function ShowAndTellEntryForm({
         <Fieldset legend="Give an hour for Earth">
           <p>
             Do you want to track hours you spent on this activity as part of the
-            Alveus communities effort to give an hour for Earth? This started as
+            Alveus community&apos;s effort to give an hour for Earth? Originally
             part of WWF&apos;s{" "}
             <Link href="/show-and-tell/give-an-hour" external>
               Give an Hour
--- a/apps/website/src/components/show-and-tell/ShowAndTellEntryForm.tsx
+++ b/apps/website/src/components/show-and-tell/ShowAndTellEntryForm.tsx
@@ -34,6 +34,8 @@ import {
 } from "../shared/form/VideoLinksField";
 import Link from "../content/Link";
 
+const dateWhenTextShortened = new Date(2024, 3, 26);
+
 export const allowedFileTypes = [
   "image/png",
   "image/jpeg",
@@ -287,7 +289,11 @@ export function ShowAndTellEntryForm({
               label="Content"
               name="text"
               defaultValue={entry?.text}
-              maxLength={700}
+              maxLength={
+                entry?.createdAt && entry.createdAt < dateWhenTextShortened
+                  ? 700
+                  : 300
+              }
             />
           </Fieldset>
         </div>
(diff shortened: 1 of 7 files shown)


apps/website/src/pages/show-and-tell/give-an-hour.tsx
```diff
@@ -78,7 +78,7 @@ const GiveAnHourPage: NextPage = () => (
   <>
     <Meta
       title="Give an Hour | Show and Tell"
-      description="Join the Alveus community and WWF to Give an Hour for Earth! Discover what actions you can take to help the environment and wildlife."
+      description="Join the Alveus community and give an hour for earth! Discover what actions you can take to help the environment and wildlife."
```
Stolers, line 81: Not sure, but maybe "Earth" (with capital e) is better here and at the other changed places? We talk here about Earth about the named place and not "dirt". (Please let me know if this comment is ok... I am used to do pull requests / code reviews at work and never done it in an open source setting)

pjeweb: Of course its okay and appreciated! :)

Stolers: Thanks, looks nice now. Do I have to close this conversation or accept that somehow? (Gitlab user here with :D)

apps/website/src/components/show-and-tell/ShowAndTellEntryForm.tsx
```diff
@@ -40,6 +34,8 @@ import {
 } from "../shared/form/VideoLinksField";
 import Link from "../content/Link";
 
+const dateWhenTextShortened = new Date(2024, 3, 26);
```
MattIPv4, line 37: Can we place this in the show and tell data file?

apps/website/src/components/show-and-tell/ShowAndTellEntryForm.tsx
```diff
@@ -386,47 +359,38 @@ export function ShowAndTellEntryForm({
       </div>
 
       <div className="space-y-10">
-        <Fieldset legend="Give an Hour">
+        <Fieldset legend="Give an hour for Earth">
           <p>
             Do you want to track hours you spent on this activity as part of the
-            Alveus community total for WWF&apos;s{" "}
+            Alveus communities effort to give an hour for Earth? This started as
```
MattIPv4, line 365:
```suggestion
            Alveus community's effort to give an hour for Earth? Originally
```

apps/website/src/pages/show-and-tell/give-an-hour.tsx
```diff
@@ -94,9 +94,10 @@ const GiveAnHourPage: NextPage = () => (
         <Heading>Give an Hour for Earth</Heading>
 
         <p className="text-lg">
-          Join the Alveus community and WWF in participating in Give an Hour for
-          Earth from March 1st to April 22nd. Discover what actions you can take
-          to help the environment and wildlife.
+          Join the Alveus community in giving hours for Earth. Discover what
+          actions you can take to help the environment and wildlife. This
+          initiative started March 1st to April 22nd as part of WWF&apos;s Give
+          an Hour for Earth campaign in 2024 but will continue beyond that date.
```
MattIPv4, line 100:
```suggestion
          actions you can take to help the environment and wildlife.
          Originally part of WWF&apos;s Give an Hour for Earth campaign
          in 2024, we're now tracking everything we do for Earth!
```

apps/website/src/components/show-and-tell/ShowAndTellEntryForm.tsx
```diff
@@ -320,7 +289,11 @@ export function ShowAndTellEntryForm({
               label="Content"
               name="text"
               defaultValue={entry?.text}
-              maxLength={700}
+              maxLength={
+                entry?.createdAt && entry.createdAt < dateWhenTextShortened
+                  ? 700
+                  : 300
+              }
```
MattIPv4, line 296: Is this enforced at the DB at all (either length), or entirely client-side?

pjeweb: So yes and no. We do a simple length check on the HTML (1k) on the server and check the text characters in client (300/700, ignoring html tags). 

I think this is the most simple way to do this. In theory this means that the text could be rejected server-side cause the tags (and attributes) even though its <= 300/700 chars in the client. Also people could "hack" the client and send up to 1k chars. But I think this is not actually a problem since we review the posts anyways.

I made a new PR to improve the formatting of server error messages #672

pjeweb: Problem with checking text chars on the server is, we'd probably need to strip the tags and count chars, ideally with JSDOM, but we already use dompurify, meaning a lot of work for very little benefit.

Checking HTML length in the client on the other hand seems like bad UX...

pjeweb: I moved the limit in the data file and added a comment. I hope that helps a bit :)

MattIPv4 (approved):

PR #672, opened 2024-04-27
Improve TRPC error messages

pjeweb:
## Describe your changes

Flatten Zod errors in TRPC mutations and queries and format them properly.

Check all the error messages look good.


apps/website/src/components/show-and-tell/ShowAndTellEntryForm.tsx
```diff
@@ -197,8 +198,7 @@ export function ShowAndTellEntryForm({
             router.push(`/show-and-tell/my-posts/`);
           }
         },
-        onError: (err) => {
-          setError(err.message);
+        onError: () => {
```
pjeweb, line 201: Using the mutation error directly instead of writing the message to state.

PR #677, opened 2024-04-29
Fix Show and Tell text shortened date to match deployment

pjeweb:
## Describe your changes

FeelsDankMan I kinda did not think about the time between commit and deployment where people could still write >300 char posts. This should fix it, by moving the check to the time of deployment.

Check posts are editable with the appropriate text length limits (esp. on 2024-04-26/2024-04-27).


MattIPv4:
🤔 Did this affect any actual posts in production?

MattIPv4 (approved):

pjeweb:
> 🤔 Did this affect any actual posts in production?

No, you just have to be careful not to touch (and save) the text field when reviewing.

PR #680, opened 2024-05-05
Fix footer hydration error due to phone number detection on iOS

bl4nkfce:
## Describe your changes

Fixes footer hydration issues when running the website on iOS, solution based on [common iOS issues](https://nextjs.org/docs/messages/react-hydration-error#common-ios-issues)

image

I tested it by running IOS Simulator (iPhone 14 running iOS 17.2)

### DIFF
--- a/apps/website/src/components/content/Meta.tsx
+++ b/apps/website/src/components/content/Meta.tsx
@@ -81,6 +81,7 @@ const Meta = ({
       {(env.NEXT_PUBLIC_NOINDEX === "true" || noindex) && (
         <meta key="robots" name="robots" content="noindex" />
       )}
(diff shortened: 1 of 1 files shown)


bl4nkfce:
We could also wrap the EIN number in `<a href=""></a>` to prevent iOS phone number detection.

pjeweb:
> We could also wrap the EIN number in `<a href=""></a>` to prevent iOS phone number detection instead of using the meta tag.

I think your solution is great. If we ever add a phone number and want it linked, we can explicitly link it. But i dont think we'd ever want that anyways!

MattIPv4 (approved):

Issue #637, opened 2024-04-05
[Idea]: Add Missing Step for iOS to Notification Popup 


pjeweb:
### Describe your idea / enhancement

The notification bell popup should show a third step explaining you need to open the PWA when added to Home Screen and press the bell again to get push notifications.

### How would this idea / enhancement be beneficial?

Improve UX for iOS users

bl4nkfce:
There's also some small CSS issues with this popup - the content overflows the popup.
Take a look at the screenshot below.

The content container has a dynamic max-width based on vw (80vw - 50px), and the content has a min-width set to 300px, which causes the overflow on some screens (tested on iphone 14 and iphone 15 simulators)

![safari_ios_iphone_14_simulator](https://github.com/alveusgg/alveusgg/assets/26897688/f546258d-2e2e-4e6b-8f2a-ab4507623ac4)

bl4nkfce:
@pjeweb
I can take care of this - I'll add the missing step to the guide and adjust the popup styles slightly on smaller screens.

pjeweb:
> @pjeweb I can take care of this - I'll add the missing step to the guide and adjust the popup styles slightly on smaller screens.

that would be great! thank you!

PR #682, opened 2024-05-05
iOS Notification Settings missing step

bl4nkfce:
## Describe your changes
Resolves https://github.com/alveusgg/alveusgg/issues/637
- Fixed overflowing image in notification settings on iOS
- Added missing step in push notifications setup for iOS

### Before
notifications_settings_before

### After
notifications_settings_after


bl4nkfce:
I noticed now that the consent management button is slightly covering the Notification Settings popover - should we also adjust that?

MattIPv4:
> I noticed now that the consent management button is slightly covering the Notification Settings popover - should we also adjust that?

I would think that the popover should probably stack above the button?

bl4nkfce:
> > I noticed now that the consent management button is slightly covering the Notification Settings popover - should we also adjust that?
> 
> I would think that the popover should probably stack above the button?

Pushed a small change to the consent button as well - removed z-index on small and medium screens.

On small screens it was covering the Notification Settings popover and the expanded navigation menu.
On medium screens (iPad) it was covering the navigation menu as well.
On large and up it's needed cause I noticed that the "leaves" on Ambassadors page were covering it and it does not interfere with anything else, so i left the previous value of 50.

pjeweb (approved):

MattIPv4 (approved):

PR #689, opened 2024-05-10
Show and Tell: Fix volunteering minutes always being set

pjeweb:
## Describe your changes

Fix volunteering minutes always being set even if checkbox not checked

Check that you can submit the posts with and without volunteering minutes **and** its showing the correct thing after saving/reloading!


pjeweb:
Deployed to https://preview.alveus.gg/show-and-tell/submit-post for testing :)

pjeweb:
Probably yes it should but no the input would still "submit" even if disabled. Will check later.

MattIPv4 (approved):

PR #691, opened 2024-05-10
Add copy-to-clipboard button to commands

### LINKED ISSUE #487: [Idea]: Improved commands list
To expand on #475, we could add:
- copy command to clipboard on click
- tooltip for detailed explanation (but what would we want to explain?)

_Originally posted by @pjeweb in https://github.com/alveusgg/alveusgg/issues/475#issuecomment-1858888848_

sougataghar477:
## Describe your changes

Adds buttons to the commands on the command page to copy them to the users clipboard.

Resolves #487

Check the commands are copyable. Layout still looks good on all devices.


apps/website/src/pages/about/tech/commands.tsx
```diff
@@ -70,6 +71,8 @@ const signature = (command: NamedCommand) => {
 };
 
 const AboutTechPage: NextPage = () => {
+  const { copyToClipboard } = useClipboard();
+  const [index, setIndex] = useState<string>("");
```
pjeweb, line 75: Can we give this a name that more clearly represents what this state is? Like `selectedCommand / setSelectedCommand`?

apps/website/src/pages/about/tech/commands.tsx
```diff
@@ -220,13 +223,28 @@ const AboutTechPage: NextPage = () => {
...
                               {signature(command)}
+                              <IconCopy
+                                onClick={() => {
+                                  copyToClipboard(signature(command));
+                                  setIndex(command.description);
+                                  setTimeout(() => {
+                                    setIndex("");
+                                  }, 2000);
+                                }}
+                              />
```
pjeweb, line 234: This should be a proper `<button>`. Also does this work well if you click multiple commands fast? I think we would need to keep the timeout id somewhere and reset it when clicking to prevent timeout handlers overriding each other.

Since we have pretty much the same button again below, it's probably worth to put the button in a reusable component - can be in the same file.

apps/website/src/pages/about/tech/commands.tsx
```diff
@@ -311,14 +329,33 @@ const AboutTechPage: NextPage = () => {
...
+                                  onClick={() => {
+                                    copyToClipboard(
+                                      `!ptzload ${camera.toLowerCase()} ${name}`,
+                                    );
+                                    setIndex(`${camera.toLowerCase()} ${name}`);
+                                    setTimeout(() => {
+                                      setIndex("");
+                                    }, 2000);
+                                  }}
+                                />
```
pjeweb, line 343: dito

apps/website/src/pages/show-and-tell/index.tsx
pjeweb, line 1: This file probably needs to be reset to `origin/main`

pjeweb (changes requested):

apps/website/src/pages/show-and-tell/index.tsx
```diff
@@ -415,6 +415,11 @@ const ShowAndTellIndexPage: NextPage<ShowAndTellPageProps> = ({
           onKeyDown={handleArrowKeys}
           tabIndex={-1}
         >
+          {/* displays posts by users */}
+          <p className="text-center text-3xl italic">
+            {totalPostsCount.toLocaleString()} posts by{" "}
+            {usersCount.toLocaleString()} community members
+          </p>
```
MattIPv4, line 422:
```suggestion
```
This change seems rather unrelated

MattIPv4 (changes requested):

PR #697, opened 2024-05-19
Secure cron API w/ Vercel's CRON_SECRET + timing-safe token comparison for protected APIs

pjeweb:
## Describe your changes

This adds a check for the cron endpoint, checking the Authorization header againts the `CRON_SECRET` env, provided by the Vercel environment (or locally through the `.env` file). This will also use a timing safe string comparison for the tokens of this and existing protected APIs using the `ACTION_API_SECRET`.

Check the APIs are still callable with and only with the correct tokens. Check the cron task still gets run on Vercel.

### DIFF
--- a/apps/website/src/pages/api/cron.ts
+++ b/apps/website/src/pages/api/cron.ts
@@ -1,10 +1,23 @@
 import type { NextApiRequest, NextApiResponse } from "next";
+
+import { env } from "@/env";
+import timingSafeCompareString from "@/server/utils/timing-safe-compare-string";
 import { runScheduledTasks } from "@/server/scheduler";
 
 export default async function handler(
   req: NextApiRequest,
   res: NextApiResponse,
 ) {
+  // Check for secret to confirm this is a valid request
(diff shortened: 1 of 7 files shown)


apps/website/src/pages/api/cron.ts
```diff
@@ -1,10 +1,23 @@
...
 export default async function handler(
   req: NextApiRequest,
   res: NextApiResponse,
 ) {
+  // Check for secret to confirm this is a valid request
+  const authHeader = req.headers["authorization"];
+  if (
+    env.CRON_SECRET &&
+    (authHeader === undefined ||
+      !timingSafeCompareString(authHeader, `Bearer ${env.CRON_SECRET}`))
```
MattIPv4, line 16:
```suggestion
  const { authorization } = req.headers;
  if (
    env.CRON_SECRET &&
    (authorization === undefined ||
      !timingSafeCompareString(authorization, `Bearer ${env.CRON_SECRET}`))
```
nit for consistency?

MattIPv4 (approved):

PR #700, opened 2024-05-22
Remove "exotic" mentions in content

MattIPv4:
## Describe your changes

Request from the folks at Alveus, Alveus should be referred to as a wildlife sanctuary, not an exotic animal sanctuary.

All mentions of exotic removed, content still reads fine.

### DIFF
--- a/apps/website/src/pages/index.tsx
+++ b/apps/website/src/pages/index.tsx
@@ -172,7 +172,7 @@ const Home: NextPage = () => {
 
             <p className="mt-8 text-lg">
               Alveus is a virtual education center following the journeys of our
-              non-releasable exotic ambassadors, aiming to educate and spark an
+              non-releasable animal ambassadors, aiming to educate and spark an
               appreciation for them and their wild counterparts.
             </p>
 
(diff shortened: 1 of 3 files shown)


apps/website/src/pages/index.tsx
```diff
@@ -239,11 +239,11 @@ const Home: NextPage = () => {
...
-                functions as an exotic animal sanctuary and as a virtual
-                education center facility to provide permanent homes to
-                non-releasable exotic animals. These animals function as
-                ambassadors, so viewers can watch their journeys, get to know
-                the animals, and gain an appreciation for their species.
+                functions as a wildlife sanctuary and as a virtual education
+                center facility to provide permanent homes to non-releasable
+                animal ambassadors. These animals function as ambassadors, so
+                viewers can watch their journeys, get to know the animals, and
+                gain an appreciation for their species.
```
MattIPv4, line 246: The sentences here feel a bit duplicative now, open to suggestions (or can leave it as-is, not a big deal)

pjeweb (approved):
Good enough, but I agree there is some redundancy. If I find the time before merging, I can try to rephrase, otherwise I'd merge as-is.

apps/website/src/pages/index.tsx
JonasFranke: smth like Viewers are able to watch ambassador's journeys, get to know them and gain appreciation for their species?

MattIPv4: I'm going to leave this as-is for now so we can get this in, but feel free to open a PR if you do want to take a stab at rewording :)

PR #704, opened 2024-05-27
Add initial stream overlay

MattIPv4:
## Describe your changes

Replicates the current stream overlay more-or-less, with the date/time + temperature, and branding for the stream.

Rest of the site is unaffected by the layout/app changes, overlay looks as expected.

### DIFF
--- a/apps/website/src/components/layout/Layout.tsx
+++ b/apps/website/src/components/layout/Layout.tsx
@@ -1,5 +1,4 @@
-import { useEffect, type ReactNode, useMemo } from "react";
-import { PT_Sans, PT_Serif } from "next/font/google";
+import { type ReactNode, useMemo } from "react";
 import Head from "next/head";
 import { useRouter } from "next/router";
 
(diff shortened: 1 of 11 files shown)


apps/website/src/components/layout/Layout.tsx
pjeweb, line 1: Not sure I like moving the font classes in the effect only. This will produce a flash of un-styled text at least sometimes. Maybe move the fonts into its own module and import them into both app components?

pjeweb (changes requested):
Nice. Can't wait for more things being added.

apps/website/src/components/layout/Layout.tsx
pjeweb: A custom document might work too. https://nextjs.org/docs/pages/building-your-application/routing/custom-document

MattIPv4: We have a custom document, but I don't think most next hooks can be used there. Let me try using the fonts there though.

MattIPv4: Yeah, next/font doesn't work in document.

pjeweb (approved):

PR #707, opened 2024-05-30
Fix overlay upcoming event refreshing + relative formatting

MattIPv4:
## Describe your changes

While the old code was fetching events every 60s, the time range it was using was fixed 🤦 The logic has now been updated so that the time range is refreshed every 60s, which then triggers events to be prefetched.

Additionally, added some relative date formatting for the events, so we'll show "Today"/"Tomorrow" instead of a date (falling back to a date for anything else).

Events are fetched every 60s, with events previously shown that are in the past no longer being shown.

### DIFF
--- a/apps/website/src/pages/stream/overlay.tsx
+++ b/apps/website/src/pages/stream/overlay.tsx
@@ -78,20 +78,34 @@ const OverlayPage: NextPage = () => {
     return () => clearInterval(weatherInterval.current);
   }, []);
 
-  // Get the events for the next three days
+  // Set the range for upcoming events
   // Refresh every 60s
   const [upcomingRange, setUpcomingRange] = useState<[Date, Date]>();
+  const upcomingRangeInterval = useRef<NodeJS.Timeout>();
   useEffect(() => {
-    const now = new Date();
-    const next = new Date(now);
-    next.setDate(next.getDate() + 3);
-    setUpcomingRange([now, next]);
+    const updateRange = () => {
+      const now = new Date();
+      const next = new Date(now);
+      next.setDate(next.getDate() + 3);
+      setUpcomingRange([now, next]);
+    };
+
+    updateRange();
+    upcomingRangeInterval.current = setInterval(updateRange, 60 * 1000);
+    return () => clearInterval(upcomingRangeInterval.current);
   }, []);
-  const events = trpc.calendarEvents.getCalendarEvents.useQuery(
-    { start: upcomingRange?.[0], end: upcomingRange?.[1] },
-    { enabled: upcomingRange !== undefined, refetchInterval: 60 * 1000 },
-  );
-  const firstEventId = events.data?.[0]?.id;
+
+  // Get the upcoming events
+  // Refresh when the range changes
+  const { data: events, refetch } =
+    trpc.calendarEvents.getCalendarEvents.useQuery(
+      { start: upcomingRange?.[0], end: upcomingRange?.[1] },
+      { enabled: false },
+    );
+  useEffect(() => {
+    if (upcomingRange) refetch();
+  }, [upcomingRange, refetch]);
+  const firstEventId = events?.[0]?.id;
 
   // If we have an upcoming event, swap socials with it
   // Swap every 60s
@@ -110,8 +124,8 @@ const OverlayPage: NextPage = () => {
     return () => clearInterval(eventInterval.current);
   }, [firstEventId]);
   const event = useMemo(
-    () => eventId && events.data?.find((event) => event.id === eventId),
-    [eventId, events.data],
+    () => eventId && events?.find((event) => event.id === eventId),
+    [eventId, events],
   );
 
   // This can be a client-side only page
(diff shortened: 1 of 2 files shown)


apps/website/src/pages/stream/overlay.tsx
```diff
@@ -78,20 +78,34 @@ const OverlayPage: NextPage = () => {
...
+  // Get the upcoming events
+  // Refresh when the range changes
+  const { data: events, refetch } =
+    trpc.calendarEvents.getCalendarEvents.useQuery(
+      { start: upcomingRange?.[0], end: upcomingRange?.[1] },
+      { enabled: false },
+    );
+  useEffect(() => {
+    if (upcomingRange) refetch();
+  }, [upcomingRange, refetch]);
```
pjeweb, line 107: Any reason we need the manual refetch? The state change should be enough to trigger the query.

```suggestion
  const { data: events } = trpc.calendarEvents.getCalendarEvents.useQuery(
    { start: upcomingRange?.[0], end: upcomingRange?.[1] },
    { enabled: upcomingRange !== undefined },
  );
```

apps/website/src/utils/datetime.ts
```diff
@@ -71,6 +71,48 @@ export const formatDateTimeParts = (
...
+    dateTime,
+    { style, time, timezone },
+    { locale, zone },
+  );
+
+  // Determine how many days away the date is
+  const dateToday = DateTime.now().setZone(zone ?? undefined);
+  const dateGiven = DateTime.fromJSDate(dateTime).setZone(zone ?? undefined);
+  const daysToday = dateToday.startOf("day").toUnixInteger() / (60 * 60 * 24);
+  const daysGiven = dateGiven.startOf("day").toUnixInteger() / (60 * 60 * 24);
```
pjeweb, line 94: I have the feeling unix timestamps of the start of the day might not always be divisible by 86400 which would break the `=== 0`/`=== 1`? But I might be wrong. I hate time ;)

pjeweb (approved):

apps/website/src/utils/datetime.ts
MattIPv4: The start of the day is always going to be 0 seconds into the day, and the epoch started at 0 seconds into a day, so it should always be divisible by 86400 (number of seconds in a day).

apps/website/src/pages/stream/overlay.tsx
MattIPv4: Doing it like this was doing all sorts of random refetches in the background, not just when the timestamps changed. I could not understand why.

pjeweb: I see yeah it will refetch on window focus, which would happen a lot in dev ;) Updated the suggestion above, but doing it in the effect works too :)

MattIPv4: This still doesn't work though, `events` seems to become empty for a second as it gets the new data, which doesn't seem to happen with a manual refetch 🤔

MattIPv4:
```ts
  const { data: events, isSuccess } =
    trpc.calendarEvents.getCalendarEvents.useQuery(
      { start: upcomingRange?.[0], end: upcomingRange?.[1] },
      { enabled: upcomingRange !== undefined },
    );
  const firstEventId = useMemo((prev?: string) => isSuccess
    ? events?.[0]?.id
    : prev, [isSuccess, events]);
```
I can do this instead if you think this is clearer (I'm thinking it might be)?

apps/website/src/utils/datetime.ts
pjeweb: So it probably will only happen on the days where daylight savings changes (since leap seconds seem to not a thing anymore).

```typescript
const dateToday = DateTime.fromJSDate(new Date(2024, 2, 11)).setZone("America/Chicago");
const dateGiven = DateTime.fromJSDate(new Date(2024, 2, 12)).setZone("America/Chicago");
const daysToday = dateToday.startOf("day").toUnixInteger() / (60 * 60 * 24);
const daysGiven = dateGiven.startOf("day").toUnixInteger() / (60 * 60 * 24);

console.log({ daysGiven, daysToday, days: daysGiven - daysToday });
  
`{daysGiven: 19793.208333333332, daysToday: 19792.25, days: 0.9583333333321207}`
```

MattIPv4: Oh ew I forgot about daylight savings -- I think it'd be safe to just floor these once the division is done?

pjeweb: Yeah! Maybe a plain round to get both ways (rounding up and down)?

MattIPv4: Where would rounding up ever be valid? We always want to get the days up to that day, not including that day?

pjeweb: Yeah i was thinking about rounding the difference, not the terms. This is great :)

pjeweb (approved):

PR #725, opened 2024-07-11
Intelligently truncate event title on stream overlay

MattIPv4:
## Describe your changes

If an event has a rather long title, such as `Connor Stream - Lowe's trip, finishing wolf enclosure and building catapults!`, it looks visually messy on the overlay as it will intersect multiple cams on the stream.

image

To solve for this, truncate the string to a maximum of 30 chars, but attempt to intelligently split the title on `-`, `|`, or `:` before resorting to chopping the string with an ellipsis.

Events with titles under 30 chars are rendered as is.

Events with titles over 30 chars with no `-`/`|`/`:` are truncated with ellipsis.

Events with titles over 30 chars containing `-`/`|`/`:`, with the first section being under 30 chars, are truncated to show just the first section.

Events with titles over 30 chars containing `-`/`|`/`:`, with the first section being over 30 chars, are truncated to show just the first section which is also truncated with ellipsis.

### DIFF
--- a/apps/website/src/pages/stream/overlay.tsx
+++ b/apps/website/src/pages/stream/overlay.tsx
@@ -14,6 +14,17 @@ import logoImage from "@/assets/logo.png";
 
 import { type WeatherResponse } from "../api/stream/weather";
 
+const truncate = (value: string, max: number) => {
+  if (value.length <= max) return value;
+
+  for (const pattern of [" - ", " | ", ": "]) {
+    const index = value.indexOf(pattern);
+    if (index !== -1 && index < max) return value.slice(0, index);
+  }
+
+  return value.slice(0, max - 3) + "...";
+};
+
 const OverlayPage: NextPage = () => {
   // Get the current time and date
   // Refresh every 250ms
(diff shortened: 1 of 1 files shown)


apps/website/src/pages/stream/overlay.tsx
```diff
@@ -14,6 +14,17 @@ import logoImage from "@/assets/logo.png";
...
 
+const truncate = (value: string, max: number) => {
+  if (value.length <= max) return value;
+
+  for (const pattern of [" - ", " | ", ": "]) {
+    const index = value.indexOf(pattern);
+    if (index !== -1 && index < max) return value.slice(0, index);
+  }
+
+  return value.slice(0, max - 3) + "...";
```
pjeweb, line 25: not really important but maybe use an actual ellipsis char here?
```suggestion
  return value.slice(0, max - 3) + "…";
```

pjeweb (approved):

apps/website/src/pages/stream/overlay.tsx
pjeweb: okay its like pretty much no difference in how its rendered. so dont bother

pjeweb (approved):

apps/website/src/pages/stream/overlay.tsx
MattIPv4: I imagine the font we're using has a ligature for `...` to be rendered as an ellipsis anyway yeah.

PR #734, opened 2024-08-03
Show & Tell: Add posts remaining counter

### LINKED ISSUE #669: [Idea]: Show & Tell: Show number of post x of x (not seen on stream) in fullscreen mode

For stream it would be nice if the presenter could see at what post of how many they are (the ones not seen on stream). This could be where the camera lives in the bottom right.

Help estimate time left for stream etc.

ACA30:
## Describe your changes

Resolves #669 - adds a counter that shows the number of remaining posts when in presenter view.

image

Implemented as (number of posts AFTER this post) / (total posts to show).

Could be implemented as `Post X of Y (Z remaining)` -  let me know if this is preferable.

I figured showing the remaining count instead of X of Y as #669 suggests allows the actually desired information to be conveyed in an easier way

N/A

## Potential Concerns

Weird sorting stuff - admin edits can make this behave strangely (i.e. editing after approved date/after shown on stream/etc), because of the way the query filters are implemented. I would consider removing the 
```ts
approvedAt: {
  gte: latestSeenEntry?.seenOnStreamAt ?? new Date(0),
},
```
filter from `postsToShow` if all posts will reliably be marked as seen on stream. This should avoid that weird behavior.

### DIFF
--- a/apps/website/next.config.js
+++ b/apps/website/next.config.js
@@ -70,11 +70,6 @@ const config = {
         protocol: "https",
         hostname: "alveus.nyc3.cdn.digitaloceanspaces.com",
       },
-      // Local Dev
-      {
-        protocol: "http",
-        hostname: "localhost",
-      },
     ],
   },
   redirects: async () => [
--- a/apps/website/next.config.js
+++ b/apps/website/next.config.js
@@ -70,6 +70,11 @@ const config = {
         protocol: "https",
         hostname: "alveus.nyc3.cdn.digitaloceanspaces.com",
       },
+      // Local Dev
+      {
+        protocol: "http",
+        hostname: "localhost",
+      },
     ],
   },
   redirects: async () => [
--- a/apps/website/src/pages/show-and-tell/index.tsx
+++ b/apps/website/src/pages/show-and-tell/index.tsx
@@ -1,50 +1,50 @@
-import type { InferGetStaticPropsType, NextPage } from "next";
-import Image from "next/image";
 import {
   useCallback,
   useEffect,
   useMemo,
   useRef,
   useState,
-  type KeyboardEventHandler,
   type WheelEvent,
+  type KeyboardEventHandler,
 } from "react";
+import type { InferGetStaticPropsType, NextPage } from "next";
+import Image from "next/image";
 import scrollIntoView from "smooth-scroll-into-view-if-needed";
 
+import { delay } from "@/utils/delay";
+import { trpc } from "@/utils/trpc";
 import {
   getPosts,
   getPostsCount,
   getPostsToShow,
   getUsersCount,
 } from "@/server/db/show-and-tell";
-import { delay } from "@/utils/delay";
-import { trpc } from "@/utils/trpc";
 
 import useOnToggleNativeFullscreen from "@/hooks/fullscreen";
 import useIntersectionObserver from "@/hooks/intersection";
 import useLocaleString from "@/hooks/locale";
 
+import IconLoading from "@/icons/IconLoading";
+import IconArrowUp from "@/icons/IconArrowUp";
 import IconArrowDown from "@/icons/IconArrowDown";
 import IconArrowsIn from "@/icons/IconArrowsIn";
 import IconArrowsOut from "@/icons/IconArrowsOut";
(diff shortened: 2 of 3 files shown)


ACA30:
Actually, I think I will implement that change to `postsToShow` - let me know if this is not ideal, and we can figure out a new solution, but I think avoiding the potential weird behavior is worth it.

apps/website/src/server/db/show-and-tell.ts
```diff
@@ -488,3 +488,22 @@ export async function deletePost(id: string, authorUserId?: string) {
...
+
+export async function getPostsToShow() {
+  // Find the latest entry marked as seenOnStream=true
+  const latestSeenEntry = await prisma.showAndTellEntry.findFirst({
+    where: {
+      seenOnStream: true,
+    },
+    orderBy: [...postOrderBy],
+  });
+
```
pjeweb, line 500: This is unused now.

apps/website/src/server/db/show-and-tell.ts
```diff
@@ -488,3 +488,22 @@ export async function deletePost(id: string, authorUserId?: string) {
...
+  // Calculate the number of entries since the latest seen entry
+  const postsToShow = await prisma.showAndTellEntry.count({
+    where: {
+      seenOnStream: false, // Only consider entries that have not been seen on stream
```
pjeweb, line 504: I think this should check that the posts counted are approved (`= approvedAt > updatedAt` which is available as `whereApproved` in that file already).

apps/website/src/pages/show-and-tell/index.tsx
pjeweb, line 1: Is the different sorting coming from prettier or is that something on your end? It looks like its sorting alphabetically? I do not think we usually do that. 

Just to keep the diffs small.

pjeweb (changes requested):
Thank you for working on this. Already looks great. Remaining / Total is the best I think!

I'll need to test the maffs a bit more when the filter works.

apps/website/src/server/db/show-and-tell.ts
ACA30: Ah this is true

apps/website/src/pages/show-and-tell/index.tsx
ACA30: Might be prettier on my end - it should obey `.prettierrc` in the repo though I thought?

ACA30: ah - the fact that there _isn't_ a `.prettierrc` would explain that lol - I will update all my PRs

pjeweb: Yeah but we dont have one afaik. We just run with prettier defaults. I think i was confusing it with eslint. Thats what would be sorting imports but it should not do it alphabetically. I'm pretty sure its your editor of some sorts.

ACA30: It's definitely prettier :p

apps/website/next.config.js
```diff
@@ -70,6 +70,11 @@ const config = {
         protocol: "https",
         hostname: "alveus.nyc3.cdn.digitaloceanspaces.com",
       },
+      // Local Dev
+      {
+        protocol: "http",
+        hostname: "localhost",
+      },
```
pjeweb, line 77: Should this be its own PR to document the change?

pjeweb (dismissed):

apps/website/next.config.js
ACA30: I can do that if you want.

MattIPv4 (approved):
LGTM (this makes an assumption that all the outstanding posts are at the start, but that seems like a fine assumption).

@pjeweb any thoughts?

ACA30:
> this makes an assumption that all the outstanding posts are at the start, but that seems like a fine assumption).

Yep, the assumption is based off of this sorting, so should work fine!
https://github.com/alveusgg/alveusgg/blob/4d55a06e757a2760a9085a8386eb2f020373a088/apps/website/src/server/db/show-and-tell.ts#L50C1-L54C12

PR #737, opened 2024-08-03
Replace Animated Bell Peepo on Updates Page

### LINKED ISSUE #698: [Idea]: Replace Animated Webp Bell Peepo on Updates Page With Video or Still Image

ACA30:
## Describe your changes

Resolves #698. Replaces the animated webp peepo with a webm version (reducing filesize from ~8MB to 1.3MB).

Should support `prefers-reduced-motion`/`prefers-reduced-data`, swapping it for a static webp image.

NOTE: The webm file is in the `public/assets` folder instead of `src/assets` because the video loader was not having a fun time, and despite a while of trying to get it to work, I could not figure it out 😛


ACA30:
I will say, the fps in this version does look a little weird (it seems really fast) - maybe as a result of the conversion process. If we have the original animation somewhere and can have the animator re-export as a webm, that would be ideal.

apps/website/src/pages/updates.tsx
```diff
@@ -59,12 +59,31 @@ const UpdatesPage: NextPage = () => {
...
+          <video
+            className="video-reduced reduced-motion:hidden reduced-data:hidden"
+            width="100%"
+            height="100%"
+            autoPlay
+            loop
+            muted
+            playsInline
+          >
+            <source src="/assets/bell-peepo.webm" type="video/webm" />
```
MattIPv4, line 72: Why are we placing this in the public assets, rather than it being an asset import?

apps/website/src/pages/updates.tsx
```diff
@@ -12,7 +12,7 @@ import { RecentNotifications } from "@/components/notifications/RecentNotificati
 import { Schedule } from "@/components/notifications/Schedule";
 import updateChannels from "@/components/shared/data/updateChannels";
 
-import bellPeepo from "@/assets/bell-peepo.webp";
```
MattIPv4, line 15: Has the old peepo asset been removed as we're no longer using it?

MattIPv4:
> I will note that the speed in this version does look a little weird (it seems really fast) - maybe as a result of the conversion process. If we have the original animation somewhere and can have the animator re-export as a webm, that would be ideal.

Have asked 👍

apps/website/src/pages/updates.tsx
ACA30: I spent a bit trying to get the video loader to accept it but was having a bunch of trouble getting it working.

I can take another stab at it if that is preferable.

MattIPv4:
[bellAlert.webm](https://github.com/user-attachments/assets/d204cbf6-6edb-4cda-bde1-23779aa81a88)

Apologies for the delay in getting this to you

apps/website/src/pages/updates.tsx
MattIPv4: Completely missed this sorry, don't run it through the video loader, just mark the file extension as a regular asset in the webpack config if it isn't already.

MattIPv4:
👋 @ACA30 any update on this?

MattIPv4:
👋 Given the lack of movement on this, I'm going to close this for now. Please feel free to open a new PR with updates if/when you have the time.

PR #739, opened 2024-08-03
Show & Tell: Add Accreditation Badge

### LINKED ISSUE #738: [Idea]: Add special zoo field to Show & Tell

With more and more people submitting Show & Tells from zoos - we should add a field where people can input a zoo name (autosuggest AZA zoos) and automatically list a "Verified AZA Zoo" badge or something on the frontend if it's an AZA zoo.

Allow mods to manually assign the badge for zoos that meet CAZA/EAZA/etc standards (or find some solution to also include those zoos in the autofill).

Reduce the amount of work for mods, searching up if zoos are AZA compliant.

ACA30:
## Describe your changes

Resolves #738. With more and more people submitting Show & Tells from zoos - this adds the option for people to mark their submission as an accredited zoo to display a badge.

For now, the badge links to the AZA about us page, but we could consider building our own page that's more generic, listing some of the larger accrediting bodies for zoos & aquariums, and why it's an important accreditation.

image
image

N/A

### DIFF
--- a/apps/website/src/components/show-and-tell/ShowAndTellEntry.tsx
+++ b/apps/website/src/components/show-and-tell/ShowAndTellEntry.tsx
@@ -1,3 +1,19 @@
+import type {
+  FileStorageObject,
+  ImageAttachment,
+  ImageMetadata,
+  LinkAttachment,
+  ShowAndTellEntryAttachment,
+  ShowAndTellEntry as ShowAndTellEntryModel,
+} from "@prisma/client";
+import parse, {
+  domToReact,
+  Element,
+  Text,
+  type DOMNode,
+  type HTMLReactParserOptions,
+} from "html-react-parser";
+import Image from "next/image";
 import {
   forwardRef,
   Fragment,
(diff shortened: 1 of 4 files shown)


apps/website/src/components/show-and-tell/ShowAndTellEntry.tsx
```diff
@@ -176,7 +178,40 @@ const Header = ({ entry, isPresentationView }: ShowAndTellEntryProps) => {
...
+                isPresentationView
+                  ? "group text-nowrap rounded-full bg-alveus-tan-900/95 p-1 text-3xl shadow-lg transition-all hover:scale-102 hover:bg-alveus-tan-900 focus:bg-alveus-tan-900"
+                  : "hover:underline focus:underline",
+              )}
+              target="_blank"
+              custom
+            >
+              <strong
+                className={classes(
+                  "bg-[#e0684b] bg-clip-text font-bold text-transparent",
```
ACA30, line 198: I'm doing it this way instead of using `text-[#e0684b]` because the font width looks different next to the volunteeringMinutes badge.

ACA30: the font color is simply the AZA logo color

MattIPv4:
🤔 Is this something that Maya/Alveus staff asked for on a stream?

I'm trying to understand the want for this? I understand folks are posting more posts about Zoos, but I'm not sure why we need a dedicated badge for it? It seems like this adds a _lot_ more moderation overhead to make sure that any post using this is actually from an accredited zoo now?

pjeweb:
> 🤔 Is this something that Maya/Alveus staff asked for on a stream?
> 
> I'm trying to understand the want for this? I understand folks are posting more posts about Zoos, but I'm not sure why we need a dedicated badge for it? It seems like this adds a _lot_ more moderation overhead to make sure that any post using this is actually from an accredited zoo now? It also doesn't really seem to cater to the international audience?

We did try to review all zoo posts in the last show and tells and add a "mod comment" if the text did not say anything about AZA.

My concern is more on the posters side, since they get the burden to check before posting, increasing the barrier to post, while we would ideally still need to check them. 

Regarding the international accreditations, I think we could just accept the local XAZA (WAZA etc.), but we would need to vet with Maya if she is comfortable with that. We do not really know if standards are comparable, even if the are "sister" organizations.

ACA30:
> We did try to review all zoo posts in the last show and tells and add a "mod comment" if the text did not say anything about AZA.
Yeah, this is specifically what I noticed is there were a significant number of mod comments clarifying that zoos were AZA accredited, so I figured a mechanism like this could be a slightly less burdensome way to do that.
> My concern is more on the posters side, since they get the burden to check before posting, increasing the barrier to post, while we would ideally still need to check them.
This is true, but it could also make sure people are more mindful about ensuring the zoos they are visiting are actually accredited.
> Regarding the international accreditations, I think we could just accept the local XAZA (WAZA etc.), but we would need to vet with Maya if she is comfortable with that. We do not really know if standards are comparable, even if the are "sister" organizations.
This was basically the intention, but I 100% agree it should be checked w/ Maya and perhaps the way it's implemented could be reconsidered to cater more to the international audience.

pjeweb:
@ACA30 Thank you for the PR again, but I think we will just use mod comments for now. As discussed above we do not see the benefits outweigh the problems.

PR #741, opened 2024-08-03
Add copy-to-clipboard button to commands

pjeweb:
## Describe your changes

> Adds buttons to the commands on the command page to copy them to the users clipboard.
> 
> Resolves https://github.com/alveusgg/alveusgg/issues/487

Remade #691 by @sougataghar477 since I scuffed up the PR branch. 

Check the commands are copyable. Layout still looks good on all devices.

Commands page preview: https://alveusgg-website-git-sougataghar477-main-alveus.vercel.app/


apps/website/src/pages/about/tech/commands.tsx
```diff
@@ -220,13 +225,28 @@ const AboutTechPage: NextPage = () => {
                           <pre>
                             <code className="text-sm">
                               {signature(command)}
+                              <CopyButton
+                                onClick={() => {
+                                  copyToClipboard(signature(command));
+                                  setSelectedCommand(command.name);
+                                  setTimeout(() => {
+                                    setSelectedCommand("");
+                                  }, 2000);
```
pjeweb, line 234: The timeouts *should* be cleared, so we'd probably need a reference for the running timeout somewhere. Not sure if its worth. Its just a bit scuffed if you click a bunch of copy buttons.

apps/website/src/pages/about/tech/commands.tsx
```diff
@@ -70,6 +71,10 @@ const signature = (command: NamedCommand) => {
 };
 
 const AboutTechPage: NextPage = () => {
+  const copyToClipboard = (command: string) => {
+    navigator.clipboard.writeText(command);
+  };
```
pjeweb, line 76: Does not need to be in component body.

apps/website/src/pages/about/tech/commands.tsx
MattIPv4: If we don't want to fix this properly now, can we leave a FIXME in the code for it?

apps/website/src/pages/about/tech/commands.tsx
MattIPv4: We should abstract this copy method into a util -- the share button also has some copy to clipboard logic, though it uses older DOM methods for doing it. We can probably do both for fallback?

apps/website/src/pages/about/tech/commands.tsx
MattIPv4: I think a better way to do this actually would be to store a Map of command name -> timeout, so that if you click a bunch in quick succession they each timeout on their own, and if you click the same one again we can clear the old timeout for it. So, not a shared state/timeout.

apps/website/src/pages/about/tech/commands.tsx
pjeweb: Yeah not sure why i did it so complicated back then. I extracted it into a hook.

apps/website/src/pages/about/tech/commands.tsx
pjeweb: Did it by moving the logic into a hook/component. So each is tracking its own state.

MattIPv4 (approved):

PR #742, opened 2024-08-03
Fix images on localhost

ACA30:
## Describe your changes

Fix images hosted in the local S3 instance (for dev environment). Split off from #734.

N/A

### DIFF
--- a/apps/website/.env.example
+++ b/apps/website/.env.example
@@ -76,7 +76,7 @@ FILE_STORAGE_ENDPOINT=http://localhost:9000
 FILE_STORAGE_SECRET=alveusgg_secret
 FILE_STORAGE_REGION=us-east-1
 FILE_STORAGE_KEY=alveusgg_key
-FILE_STORAGE_CDN_URL=
+FILE_STORAGE_CDN_URL=http://localhost
 FILE_STORAGE_BUCKET=alveusgg
 FILE_STORAGE_PATH_STYLE=true # MinIO requires path style
 
--- a/apps/website/.env.example
+++ b/apps/website/.env.example
@@ -76,7 +76,7 @@ FILE_STORAGE_ENDPOINT=http://localhost:9000
 FILE_STORAGE_SECRET=alveusgg_secret
 FILE_STORAGE_REGION=us-east-1
 FILE_STORAGE_KEY=alveusgg_key
-FILE_STORAGE_CDN_URL=http://localhost
+FILE_STORAGE_CDN_URL=
 FILE_STORAGE_BUCKET=alveusgg
 FILE_STORAGE_PATH_STYLE=true # MinIO requires path style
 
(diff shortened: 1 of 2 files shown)


MattIPv4:
@pjeweb Do we need any of the hard-coded S3 hosts in this config still? I'm wondering if we can just rely on the CDN URL env?

And then another related thought, can we try setting the CDN URL to localhost for local dev, so we don't need to hard-code localhost?

pjeweb:
> @pjeweb Do we need any of the hard-coded S3 hosts in this config still? I'm wondering if we can just rely on the CDN URL env?
> 
> And then another related thought, can we try setting the CDN URL to localhost for local dev, so we don't need to hard-code localhost?

Should be fine to remove. I just checked, the dev/preview CDN env was not set right on vercel, otherwise it looks like that should work.

ACA30:
> > Do we need any of the hard-coded S3 hosts in this config still? I'm wondering if we can just rely on the CDN URL env?
> 
> Should be fine to remove. I just checked, the dev/preview CDN env was not set right on vercel, otherwise it looks like that should work.

Removed.

> And then another related thought, can we try setting the CDN URL to localhost for local dev, so we don't need to hard-code localhost?

Done. @MattIPv4

pjeweb:
Does this actually work? Does the CDN for localhost not need the bucket path in there for `getViewUrlForFileStorageObjectKey()` in `apps/website/src/server/utils/file-storage.ts`. 

I think the safer version is to not use the CDN env var on local dev and falling back to the endpoint url for the next config remote patterns.

Was working on this too and ended up with this in the `next.config.js`:

```js

/**
 * @param {string} url
 * @returns {import("next/dist/shared/lib/image-config").RemotePattern}
 */
function urlOriginAsRemotePattern(url) {
  const parsed = new URL(url);
  return {
    protocol: parsed.protocol === "http:" ? "http" : "https",
    hostname: parsed.hostname,
    port: parsed.port,
  };
}

// … omitted

/** @type {import("next").NextConfig} */
const config = {
  // … omitted
  images: {
    remotePatterns: [
      // … omitted
      // S3 - File Storage
      urlOriginAsRemotePattern(
        process.env.FILE_STORAGE_CDN_URL ||
          process.env.FILE_STORAGE_ENDPOINT ||
          "https://localhost",
      ),

// … omitted
…

**Comment by MattIPv4** [comment]
:this: Was just testing the same thing locally - you could set the CDN URL to include the bucket locally and parse out just the origin for the config, or do what you're doing which I think is probably the cleaner way.

**Comment by ACA30** [comment]
Implemented @pjeweb's solution!

File: apps/website/.env.example [hunk 1]
```diff
@@ -76,7 +76,7 @@ FILE_STORAGE_ENDPOINT=http://localhost:9000
 FILE_STORAGE_SECRET=alveusgg_secret
 FILE_STORAGE_REGION=us-east-1
 FILE_STORAGE_KEY=alveusgg_key
-FILE_STORAGE_CDN_URL=
+FILE_STORAGE_CDN_URL=http://localhost
```
pjeweb, line 79: This should be reverted then too

pjeweb (approved):
Tested it, should be fine.

Issue #592, opened 2024-03-10
[Idea]: Add toggle for local/Alveus time on updates calendar


MattIPv4:
> I also tried to add a toggle to show in Alveus time (CT) _or_ local time.

_Originally posted by @pjeweb in https://github.com/alveusgg/alveusgg/pull/587#pullrequestreview-1926321755_

MattIPv4:
Relevant bits of code:

The admin panel month selection + event fetching:

https://github.com/alveusgg/alveusgg/blob/5ca12a4afaa74d5d7abe052109deb7b934d57e1d/apps/website/src/components/admin/calendar-events/CalendarEvents.tsx#L14-L62

The public calendar month selection + event fetching:

https://github.com/alveusgg/alveusgg/blob/5ca12a4afaa74d5d7abe052109deb7b934d57e1d/apps/website/src/components/notifications/Schedule.tsx#L10-L60

The calendar rendering itself:

https://github.com/alveusgg/alveusgg/blob/5ca12a4afaa74d5d7abe052109deb7b934d57e1d/apps/website/src/components/content/Calendar.tsx#L71-L225

ExperimentalCyborg:
I'll give it a shot this weekend / next week 🫡

ExperimentalCyborg:
Work stuff came up, it'll be Soon™

pjeweb:
Another idea would be to link to a page where you can see the time in all timezones?! Eg https://www.timeanddate.com/worldclock/fixedtime.html?iso=20240530T10

ExperimentalCyborg:
Nah i just need to get out of my ADHD procrastination cycle and actually do the thing properly 🥲

React is new to me but as a Professional Software Developer™ that shouldn't be an issue

PR #744, opened 2024-08-07
Add timezone toggle to schedule calendar

bl4nkfce:
## Describe your changes

Resolves https://github.com/alveusgg/alveusgg/issues/592

- Updated timezone toggle styles and semantics
- Timezone is now stored in the localStorage

I've switched the checkbox into a button. This way we don't have to write custom logic for handling focus state in order to display an outline when navigating via keyboard. The HTML for this element is based on the [W3C APG example](https://www.w3.org/WAI/ARIA/apg/patterns/switch/examples/switch-button/) implementation.

### Desktop view
image

It is a little cluttered on mobile, so maybe we can place the toggle between the arrows?

### Mobile (current)
image

### Mobile (moved down a bit)
image

1. Toggle the switch, check if there are any changes in the calendar.
2. Refresh the page and check if the toggle is in the same state as before.

Optionally check localStorage before and after refreshing the page to see if the timezone is stored correctly.


pjeweb:
Im a GH noob. Will this actually merge on my branch on not main? Can you make it a PR to the main branch instead so we can skip the double PR?

pjeweb:
@bl4nkfce this looks great. Is there an accessible way to lose the focus style after click? i always have to check again. Its just a bit confusing when the toggle switch stays dark after disabling it.

bl4nkfce:
> Im a GH noob. Will this actually merge on my branch and not main? Can you make it a PR to the main branch instead so we can skip the double PR?

Yeah, it would merge my branch into yours. I'll change the target to main.

---


> @bl4nkfce this looks great. Is there an accessible way to lose the focus style after click? i always have to check again. Its just a bit confusing when the toggle switch stays dark after disabling it.

Since we have the native focus outline applied to the element when navigating via keyboard, we don't need the additional dark background styles (we can remove the focus-within styles).

For desktop we can apply the "hover" styles only when the switch is not focused,
it would look like this:

https://github.com/user-attachments/assets/b532164b-6eb8-4a9a-860a-8853c17e3d62

pjeweb:
Since some part of this PR is from my brain, I'd love you, @MattIPv4, to review this when you have the time. I'll go through it too tho.

MattIPv4:
Love this! One thing though, can we see if we can improve the clarity of the UX around changing between local/Alveus time? I'm not sure a toggle at the top is particularly clear what it is doing, nor that by default the schedule is local time?

Perhaps we could just have a line of text? `Using your local time (UTC), [switch to Alveus time (CT)](js click handler)` if that makes sense? (and flip the sentence when clicked)

MattIPv4:
Oooh yeh, segmented button works too 👍 And feel free to play around with the locale stuff to have it work nicer

pjeweb (dismissed):
This looks great to me. Thank you

apps/website/src/components/calendar/Calendar.tsx
```diff
@@ -138,6 +143,36 @@ const getCalendarTheme = (month?: number): CalendarTheme => {
...
+
+export function useCalendarEventsQuery(selected?: MonthSelection) {
+  const start = useMemo(
+    () => selected && new Date(selected.year, selected.month, 1),
+    [selected],
+  );
+  const end = useMemo(
+    () => selected && new Date(selected.year, selected.month + 1, 1),
+    [selected],
+  );
```
MattIPv4, line 168: Do we need to consider timezone here? If an event was right at the end of the day Alveus time on the last day of the month, it wouldn't show in this query based on local time, I think?

MattIPv4: image
image
image
image

MattIPv4: (As an aside, the fading that is applied to previous days doesn't seem to be timezone-aware)

MattIPv4:
👋 @bl4nkfce any update on this?

bl4nkfce:
I looked at this briefly today.
I noticed we're using some luxon utils in some places and native dates in other.
Is it okay if i try to refactor this to use luxon only?
We definitely have to use the timezones here so I'll try to fix it either way, probably tomorrow or early next week.

MattIPv4:
👍 Luxon is fine, we tend to use it anywhere where we need to deal with timezones.

bl4nkfce:
I've rebased the branch and pushed some changes.

- In general: updated some parts that used native JS Date objects in favor of Luxon DateTime objects.
- Updated the query that fetches events with the timezone info and tested some cases (end of month, end of day times).
- The calendar should correctly set the "current day" (and past days) highlight based on the selected timezone.
- Fixed Timezone Switch button text color in admin view (should now have better contrast with button bg, previously it inherited a very bright color)

@pjeweb @MattIPv4
If you could please test this thing when you have some time that would be great. I changed a bunch of things, so it's possible I broke something that worked and missed it during testing 😅 



https://github.com/user-attachments/assets/0a12508c-d63a-44d2-973a-11b8a89d5809

apps/website/src/components/calendar/Calendar.tsx
```diff
@@ -383,20 +412,14 @@ const Calendar = ({
...
                       >
-                        {date.toLocaleString(undefined, {
-                          minimumIntegerDigits: 2,
-                        })}
-
+                        {paddedDate}
                         {/* Render the day of the week for mobile */}
                         <span className="md:hidden">
-                          {days[fullDate.getDay()]?.slice(0, 1)}
+                          {days[(i + startDay) % 7]?.charAt(0)}
```
MattIPv4, line 422: You seem to be off by one here?

MattIPv4 (dismissed):
Otherwise, LGTM

apps/website/src/components/calendar/Calendar.tsx
bl4nkfce: I think it works now?  https://github.com/alveusgg/alveusgg/pull/744/commits/a15d770e22746bb3e48fa2ddbe960b747b039271 
The date used for selected month/year was not starting on the first day of the month, it was the current date instead hence incorrect weekday value.

MattIPv4 (approved):
LGTM, tyvm.

pjeweb:
> @pjeweb did you want to review this before I merge it?

should be fine to merge. i'll take a look at it later though

bl4nkfce:
> Has anyone tried the extreme timezones (-12/+14)?

I just checked +14 and found two issues:
1. `.toLocaleTimeString()` returns this format for `KI-L` locale - we could just call `.toFormat('HH:mm')` to always display the time in a 24 hour format with padded mintues.
If not, we can leave it as is and rely on locale to take care of this formatting, but then it would be nice to do that for dates as well 🤔 
image

2. After switching from +14 to Alveus time, current day is not marked correctly (for some reason the "today" date is a month behind?)

Tbh I think my recent changes introduced more issues than solved. 😅 

> 
> One minor thing I just noticed. Would it be easier to understand if we'd call the local time "Your Time", since Alveus time is also local, just at Alveus...?

I like this idea, was also thinking about "My time (CEST)"

MattIPv4 (approved):

pjeweb (approved):
I think this is pretty solid now! If we wanna change the label for "my time", we can do that in a separate PR and get this merged first (finally).

pjeweb:
@bl4nkfce thank you again! are you in twitch chat too? or in one of the discords?

bl4nkfce:
yeah, I'm foszki on twitch and in Mayacord

PR #745, opened 2024-08-07
Add YouTube feed to homepage

### LINKED ISSUE #635: [Idea]: Social/video feed section

I think having a section of current instagram posts or youtube videos could be nice. The shop has that for example. Not sure where it would fit.

IG feed:

ig-feed

Youtube feed:

videos

Not 100% if its a good idea. We already have a lot of content, so adding more might not be the best idea, but I think it would present more of the (good) socials content on the site and could drive cross-platform engagement.

ACA30:
## Describe your changes

Resolves #635. Adds a YouTube feed near the bottom of the homepage.

Uses ISR (re-validated every 30 mins) to cache feed and avoid expensive checks. Excludes shorts.

Using public YouTube XML feed to fetch from both [@AlveusSanctuary](https://www.youtube.com/@AlveusSanctuary) and [@AlveusSanctuaryHighlights](https://www.youtube.com/@AlveusSanctuaryHighlights) on YouTube. Displays the 4 most recent videos across both channels.

Considered adding IG feed as mentioned in the issue, but it would be considerably harder. Requires you to use Meta APIs, auth, and some other stuff that seems like more of a headache than it's worth.

image

N/A


apps/website/src/pages/index.tsx
```diff
@@ -140,7 +144,80 @@ const getTwitchEmbed = (
...
+  }),
+});
+
+type Props = {
+  videos: Video[];
+};
+
+const fetchYouTubeFeed = async (channelId: string): Promise<Video[]> => {
+  const url = `https://www.youtube.com/feeds/videos.xml?channel_id=${channelId}`;
+  const response = await axios.get(url);
```
MattIPv4, line 175: Please can we avoid using axios here as it means introducing an entire new dependency? A regular fetch should be just fine?

apps/website/src/pages/index.tsx
```diff
@@ -140,7 +144,80 @@ const getTwitchEmbed = (
...
+    .map((video) => VideoSchema.parse(video)); // Validate each video against the schema
+};
+
+export const getStaticProps: GetStaticProps = async () => {
+  try {
+    const channelIds = ["UCbJ-1yM55NHrR1GS9hhPuvg", "UCfisf6HxiQr8_4mctNBm9cQ"];
+    const fetchPromises = channelIds.map((channelId) =>
+      fetchYouTubeFeed(channelId),
+    );
+    const videosArrays = await Promise.all(fetchPromises);
```
MattIPv4, line 195:
```suggestion
    const videosArrays = await Promise.all(channelIds.map((channelId) =>
      fetchYouTubeFeed(channelId),
    ));
```
This extra const seems unnecessary?

apps/website/src/pages/index.tsx
```diff
@@ -140,7 +144,80 @@ const getTwitchEmbed = (
...
+    };
+  } catch (error) {
+    return {
+      props: {
+        videos: [],
+        error: "Failed to fetch YouTube feed",
+      },
+      revalidate: 300, // revalidate after 5 mins
+    };
+  }
```
MattIPv4, line 217: Do we need to try/catch here? I believe we should actually just allow the error to be thrown which will allow Next to keep reusing the old data until it next attempts to revalidate?

apps/website/src/components/content/YouTubeCarousel.tsx
```diff
@@ -0,0 +1,72 @@
...
+  published: Date;
+  link?: string;
+  vodId?: string;
+};
+
+type YouTubeCarouselProps = {
+  videos: Video[];
+};
+
+const YouTubeCarousel: FC<YouTubeCarouselProps> = ({ videos }) => {
```
MattIPv4, line 20: Please do not use React.FC, it is considered a deprecated/bad practice

apps/website/src/pages/index.tsx
```diff
@@ -140,7 +144,80 @@ const getTwitchEmbed = (
...
+        title: z.array(z.string()).nonempty(),
+        published: z.array(z.string()).nonempty(),
+      }),
+    ),
+  }),
+});
+
+type Props = {
+  videos: Video[];
+};
```
MattIPv4, line 171: Can we just rely on inferring the props here, like we do for our other pages? This type doesn't actually match what you're returning from static props at present, as you have an error prop sometimes returned, which inferring would handle.

apps/website/src/pages/index.tsx
```diff
@@ -140,7 +143,71 @@ const getTwitchEmbed = (
...
+
+  return {
+    props: {
+      videos: latestVideos,
+    },
+    revalidate: 1800, // revalidate after 30 minutes
+  };
+};
+
+const Home: NextPage<{ videos: Video[] }> = ({ videos }) => {
```
MattIPv4, line 210: Can we use `InferGetStaticPropsType` like we do on some other pages, so we just infer the props based on what `getStaticProps` returns, rather than hard-coding the prop types here?

MattIPv4: Take a look at the show and tell index page as a reference point :)

apps/website/src/components/content/YouTubeCarousel.tsx
```diff
@@ -0,0 +1,64 @@
...
+import Heading from "@/components/content/Heading";
+import { formatDateTime } from "@/utils/datetime";
+
+type Video = {
+  videoId: string;
+  title: string;
+  published: Date;
+};
+
+const YouTubeCarousel = ({ videos }: { videos: Video[] }) => {
```
MattIPv4, line 12: I wonder if you should accept a `dark` prop here like a few of our other components -- the font color used for the date is pretty unreadable when on the green background?

apps/website/src/pages/index.tsx
ACA30: my bad i completely misunderstood what you meant haha - fixed now

MattIPv4 (approved):
LGTM, tyvm

PR #748, opened 2024-08-08
Add autoscroll to collaborations slider

### LINKED ISSUE #648: [Idea]: Add arrows for creator selector and position relative to carousel.

Add arrows you can click to navigate creators. The arrows should have 2 states, one for the ends and one for everywhere else. The end one could be smaller and with less color value, and the clickable one bigger and bolder. 
Also i would suggest adding a way of knowing where in the carousel you are relative to the rest. 

Carousel Left

Carousel mid

I feel like while Shift + Scroll and Drag are both good ways of moving, a lot of people will find it more natural in desktop to click an arrow at the end of the carousel. The second suggestion of place in carousel will help to reinforce the idea that there are more creators to search from and to "Ground" the carousel in some way. (Right now there are not that many creators outside the main view but some day there will be)

ACA30:
## Describe your changes

Resolves #648. Adds slow autoscrolling to the slider of collaborators to help indicate that there is more content if you scroll.

As Matt requested, this does not add arrows or additional UI elements, but sticks to simply adding a slow autoscroll of that slider.

The scroll should infinitely loop. Scrolling does get interrupted when you reach the end of 2 cycles, as the scroll position is reset. I implemented a different solution prior to this one that did allow infinite scrolling with no interruption or minor jumps, but it will theoretically infinitely loop repeating the slider, which seems not ideal haha. Autoscroll is paused on hover.

image

N/A

### DIFF
--- a/apps/website/src/pages/collaborations.tsx
+++ b/apps/website/src/pages/collaborations.tsx
@@ -43,9 +43,24 @@ const Creators = ({ className }: { className?: string }) => {
   const drag = useDragScroll();
   const containerRef = useRef<HTMLUListElement>(null);
   const [isHovered, setIsHovered] = useState(false);
+  const [supportsHover, setSupportsHover] = useState(false);
 
   useEffect(() => {
-    if (containerRef.current) {
+    const mediaQuery = window.matchMedia("(hover: hover)");
+    setSupportsHover(mediaQuery.matches);
+
+    const handleHoverChange = (event: MediaQueryListEvent) => {
+      setSupportsHover(event.matches);
+    };
+
+    mediaQuery.addEventListener("change", handleHoverChange);
+    return () => {
+      mediaQuery.removeEventListener("change", handleHoverChange);
+    };
+  }, []);
+
+  useEffect(() => {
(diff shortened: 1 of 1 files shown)


apps/website/src/pages/collaborations.tsx
```diff
@@ -41,16 +41,65 @@ const creators = collaborations
...
 
   return (
     <div className={classes("flex justify-center", className)}>
       <div className="relative isolate max-w-full">
         <ul
           className="scrollbar-none group/creators flex max-w-full cursor-grab flex-row gap-y-4 overflow-x-auto pb-2 pl-12 pr-8 pt-6"
           onMouseDown={drag}
+          ref={containerRef}
+          onMouseEnter={() => setIsHovered(true)}
+          onMouseLeave={() => setIsHovered(false)}
```
pjeweb, line 96: Can we also handle keyboard `onFocus`/`onBlur`? Iirc that should work with bubbling in react without attaching handlers to each link.

MattIPv4:
👋 @ACA30 any update on this?

MattIPv4:
No progress, closing.

PR #763, opened 2024-09-03
Add monthly scheduled regular calendar events

MattIPv4:
## Describe your changes

Once a month on the 21st, create all the regular calendar events that happen each week for the following month.

If running locally, you may want to swap `interval: { months: 1 },` to `interval: { minutes: 1 },`, and add the following before the final `for` loop that creates the events:

```ts
  await prisma.calendarEvent.deleteMany({
    where: {
      startAt: {
        gte: nextMonth.toJSDate(),
        lt: nextMonth.plus({ months: 1 }).toJSDate(),
      },
    },
  });
```

### DIFF
--- a/apps/website/src/data/scheduled-tasks.ts
+++ b/apps/website/src/data/scheduled-tasks.ts
@@ -23,7 +23,7 @@ const scheduledTasksConfigSchema = z.object({
   tasks: z.array(
     z.object({
       id: z.string(),
-      task: z.function().returns(z.promise(z.void())),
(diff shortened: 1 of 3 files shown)


MattIPv4:
@pjeweb thoughts on how we could make this run for September in prod?

I was thinking perhaps:

1. Add a fake previous run in the DB, so it thinks it is overdue for Sept
2. Update the logic that calls `task()` so that we pass in the computed due date for the current run (rather than the current date)

pjeweb (approved):
Yeah I think adding the task seems like the easiest solution. Can do that in preparation.

PR #767, opened 2024-09-07
Fix scheduled tasks getting stuck or not running properly

### LINKED ISSUE #765: [Bug]: Scheduled Tasks sometimes stall out and block forever

I've seen on production multiple times that scheuduled tasks do not finish, blocking following calls. 

Tasks should probably just have a max time to run after which they are considered stalled.

### Steps to reproduce

Wait :shrug: 

To see the block you can also null the `finishedAt` column for the last `TaskExecutionEvent` in db.

### Browser specs

n/a

### Anything else?

Fix should probably go somewhere here: https://github.com/alveusgg/alveusgg/blob/b0ddc2b0290c517ab2b8c08b8deca2d7235d8673/apps/website/src/server/scheduler.ts#L37-L39

pjeweb:
## Describe your changes

This fixes two problems:
- The two retry tasks (webhooks and notifications) failed because the conditions for the exponentially delayed retry were broke. The array of delays was not filled, because it was using the `Array.map()` method on a sparse array. Instead this now fills the array before mapping over it.
- When tasks failed with an exception, they never were tagged as finished, getting the task stuck. This will both catch exceptions and allow stuck tasks to be run after some time (5 minutes).

Resolves #765

This is a hard one to test. In a fresh db you should see that the tasks run through and do not leave an unfinished `TaskExecutionEvent` entry.

### DIFF
--- a/apps/website/src/server/scheduler.ts
+++ b/apps/website/src/server/scheduler.ts
@@ -18,7 +18,13 @@ async function executeTask(
       task: taskConfig.id,
     },
   });
-  await taskConfig.task(nextExecutionTime.toJSDate());
+
+  try {
+    await taskConfig.task(nextExecutionTime.toJSDate());
+  } catch (exception) {
+    console.error(`Error executing task ${taskConfig.id}`, exception);
+  }
+
   const finishTime = new Date();
   await prisma.taskExecutionEvent.update({
     where: { id: event.id },
--- a/apps/website/src/server/scheduler.ts
+++ b/apps/website/src/server/scheduler.ts
@@ -6,6 +6,8 @@ import { prisma } from "@/server/db/client";
 
 type TaskConfig = ScheduledTasksConfig["tasks"][number];
 
+const MAX_EXECUTION_DURATION = 1000 * 60 * 5; // 5 minutes
+
 async function executeTask(
   taskConfig: TaskConfig,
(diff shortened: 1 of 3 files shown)


apps/website/src/server/scheduler.ts
```diff
@@ -6,6 +6,8 @@ import { prisma } from "@/server/db/client";
 
 type TaskConfig = ScheduledTasksConfig["tasks"][number];
 
+const MAX_EXECUTION_DURATION = 1000 * 60 * 5; // 5 minutes
```
MattIPv4, line 9: Should this be something defined on each task, in case we have a known-long-running task?

pjeweb: We could do that yeah. Not sure if its worth though. I chose 5m/300s because thats the limit for vercel functions in the Pro plan. So it is kinda the maximum it could ever run. I also do not expect for tasks to fail like this anymore, so I see it as a failsafe.

pjeweb: Forgot to link the docs: https://vercel.com/docs/functions/limitations

MattIPv4: Ah cool, that's fine then 👍

MattIPv4 (approved):

PR #780, opened 2024-09-27
Refactor scheduled tasks

### LINKED ISSUE #764: [Idea]: Refactor scheduled tasks - simplify code

Reviewing #763 I noticed we have some over-engineering in https://github.com/alveusgg/alveusgg/blob/main/apps/website/src/data/scheduled-tasks.ts

I propose removing zod completely and use a typescript type/interface directly. `getScheduledTasksConfig()` also does not need to be async. This was originally set up like this to support loading the config from a file or database without having to change the signature. I do not think that will happen.

Reduce code. Improve maintainability.

matadaniel:
resolves #764

### DIFF
--- a/apps/website/src/data/scheduled-tasks.ts
+++ b/apps/website/src/data/scheduled-tasks.ts
@@ -7,7 +7,23 @@ import { retryOutgoingWebhooks } from "@/server/outgoing-webhooks";
 import { OUTGOING_WEBHOOK_TYPE_DISCORD_CHANNEL } from "@/server/discord";
 import { createRegularCalendarEvents } from "@/server/db/calendar-events";
 
-export type ScheduledTasksConfig = z.infer<typeof scheduledTasksConfigSchema>;
+export type ScheduledTasksConfig = {
+  tasks: {
+    id: string;
+    task: (date: Date) => Promise<void>;
+    label: string;
+    interval: {
+      years?: number;
+      months?: number;
+      weeks?: number;
+      days?: number;
+      hours?: number;
+      minutes?: number;
+      seconds?: number;
+    };
+    startDateTime: Date;
+  }[];
+};
 
 const durationSchema = z.object({
   years: z.number().optional(),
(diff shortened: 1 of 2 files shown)


apps/website/src/data/scheduled-tasks.ts
```diff
@@ -72,6 +64,6 @@ const config: ScheduledTasksConfig = {
   ],
 };
 
-export async function getScheduledTasksConfig() {
+export function getScheduledTasksConfig() {
```
MattIPv4, line 67: Does this even need to be a method at this point, we could just export default the config, so we're not creating new object instances all the time?

matadaniel: I thought that as well. I didn't want to make such a change in case it was some convention this repo is following. We could give a more descriptive name to the config and export it or just export default as you suggested.

matadaniel: > Does this even need to be a method at this point, we could just export default the config, so we're not creating new object instances all the time?

Does it create new object instances with each invocation? I think it always references the same instance, but the change would still go along with the spirit of "Reduce code. Improve maintainability." from the issue.

pjeweb: yeah go for it. its not creating new instances or anything.

matadaniel: Any preference for how the config is exported?

pjeweb: I think we have a mix of default export and named exports. I do not really mind either way. I'd probably name it `scheduledTasks` i guess 🤷  We do not really need the `tasks` object either. Can just be the array of tasks.

MattIPv4: > Does it create new object instances with each invocation? I think it always references the same instance

I think you're right, but it probably creates a new closure every time still.

MattIPv4 (approved):

MattIPv4:
@pjeweb want to test this anywhere, or good to go?

pjeweb:
testing in production o7

matadaniel:
this is the way

PR #784, opened 2024-10-02
Update data (Push Pop's animal quest)

MattIPv4:
## Describe your changes

cc https://github.com/alveusgg/data/pull/77

Twitch VoD works for episode, broadcast date + length are correct


MattIPv4:
#788

PR #786, opened 2024-10-02
Implement calendar layout based on user locale

### LINKED ISSUE #777: [Idea]: Change calendar layout based on user locale

The schedule should show weeks based on the user locale starting on Monday or Sunday. This should be available through for example: https://moment.github.io/luxon/#/intl?id=accessing-locale-based-week-info

I am not sure if it should always be US calendar for the alveus time, i would think not. Instead show both timezones with the week matching the users locale.

Better UX

jateez:
## Describe your changes

resolves #777 
The schedule calendar should now displays weeks based on the user’s locale starting day.

en-GB
image

en-US
image

...

Use Chrome DevTools to simulate different country locales and verify the calendar's starting day.

...

### DIFF
--- a/apps/website/src/components/calendar/Calendar.tsx
+++ b/apps/website/src/components/calendar/Calendar.tsx
@@ -8,9 +8,8 @@ import {
 
 import { Transition } from "@headlessui/react";
 import { DateTime } from "luxon";
-
-import { ListBucketInventoryConfigurationsOutputFilterSensitiveLog } from "@aws-sdk/client-s3";
 import { getWeekStartByLocale } from "weekstart/full";
+
 import { trpc } from "@/utils/trpc";
 import { classes } from "@/utils/classes";
 
--- a/apps/website/src/components/calendar/Calendar.tsx
+++ b/apps/website/src/components/calendar/Calendar.tsx
@@ -8,7 +8,7 @@ import {
 
 import { Transition } from "@headlessui/react";
 import { DateTime } from "luxon";
-import { getWeekStartByLocale } from "weekstart/full";
+import "@bart-krakowski/get-week-info-polyfill";
 
 import { trpc } from "@/utils/trpc";
 import { classes } from "@/utils/classes";
@@ -294,9 +294,13 @@ export function Calendar({
 
   if (!today || !selectedDateTime) return null;
 
-  // Get week start (0 = Sunday, 1 = Monday, etc) and adjust Sunday to 7
-  const locale = Intl.DateTimeFormat().resolvedOptions().locale;
-  const startDay = getWeekStartByLocale(locale) || 7; // Convert 0 to 7 for Sunday.
+  const locale = new Intl.Locale(
+    Intl.DateTimeFormat().resolvedOptions().locale,
+  );
+  // Get first day of the week based on Locale.
+  // Defaults to 7 if getWeekInfo is not supported.
+  // 1 = Monday, 7 = Sunday.
+  const startDay = locale.getWeekInfo?.()?.firstDay ?? 7;
   const startOffset = (7 + selectedDateTime.weekday - startDay) % 7; // Luxon uses 1-based days
   const weeks = Math.ceil((startOffset + daysInMonth!) / 7);
 
--- a/apps/website/src/components/calendar/Calendar.tsx
+++ b/apps/website/src/components/calendar/Calendar.tsx
@@ -7,8 +7,10 @@ import {
 } from "react";
 
 import { Transition } from "@headlessui/react";
-import { DateTime, Info } from "luxon";
+import { DateTime } from "luxon";
 
+import { ListBucketInventoryConfigurationsOutputFilterSensitiveLog } from "@aws-sdk/client-s3";
+import { getWeekStartByLocale } from "weekstart/full";
 import { trpc } from "@/utils/trpc";
 import { classes } from "@/utils/classes";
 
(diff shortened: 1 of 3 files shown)


pjeweb (commented):
This looks great and works for me when emulating in the chrome dev tools.

Sadly I asked some US chatters to try it out, and it weirdly did not work for them. They do have the locale set to 'en-US' though (`new Intl.DateTimeFormat().resolvedOptions().locale => 'en-US'` ).

jateez:
> This looks great and works for me when emulating in the chrome dev tools.
> 
> Sadly I asked some US chatters to try it out, and it weirdly did not work for them. They do have the locale set to 'en-US' though (`new Intl.DateTimeFormat().resolvedOptions().locale => 'en-US'` ).

okay, thank you sm for the review. will check it again asap :)

pjeweb:
> > This looks great and works for me when emulating in the chrome dev tools.
> > Sadly I asked some US chatters to try it out, and it weirdly did not work for them. They do have the locale set to 'en-US' though (`new Intl.DateTimeFormat().resolvedOptions().locale => 'en-US'` ).
> 
> okay, thank you sm for the review. will check it again asap :)

I honestly have no idea what is wrong with your code. My only idea was to either not pass a locale (maybe luxon defaults to the correct locale?) or use the snippet I pasted above (`new Intl.DateTimeFormat().resolvedOptions().locale` ) to get the locale instead of creating a DateTime instance?

If you need testers, you can ask in alveussanctuary chat or if there is a stream there maybe rather in the alveusgg chat. There should be US/non-US chatters around that can help test.

jateez:
Hi, 

I did a bit of research and testing, and from what i can gather, Luxon's ```Info.getStartOfWeek``` is not functioning properly on Firefox browsers due to Firefox does not support the ```Intl.Locale.prototype.getWeekInfo()``` method, which Luxon relies on

References:
- [MDN's documentation for Intl.Locale.prototype.getWeekInfo()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/Locale/getWeekInfo)
- [Luxon's documentation for Locale-based weeks info](https://moment.github.io/luxon/#/intl?id=locale-based-weeks)

Here’s the code I used for testing:
```ts
  console.log(
    Info.getStartOfWeek({ locale: "en-GB", }),
    "Testing of en-GB -- expected 1 (Monday)",
  );

  console.log(
    Info.getStartOfWeek({ locale: "en-US" }),
    "Testing of en-US -- expected 7 (Sunday)",
  );

  console.log(
    Info.getStartOfWeek({
      locale: Intl.DateTimeFormat().resolvedOptions().locale
    }),
    `Testing of Intl.DateTimeFormat().resolvedOptions().locale ${Intl.DateTimeFormat().resolvedOptions().locale} -- expected 7 (Sunday)`,
  );

  console.log(
    Info.getStartOfWeek({
      locale: DateTime.local().locale
    }),
    `Testing of DateTime.local().locale ${DateTime.local().locale} -- expected 7 (Sunday)`,
  );

    //Additional test
  console.log(
    Info.getStartOfWeek({ locale: "ja-JP" }),
    "Testing of ja-JP -- expected 7 (Sunday)",
  );

  console.log(
    Info.getStartOfWeek({ locale: "fa-IR" }), 
    "Testing of fa-IR -- expected 6 (Saturday)",
  );

```

I used the following for startDay:

```ts
const startDay = Info.getStartOfWeek({
    locale: Intl.DateTimeFormat().resolvedOptions().locale,
  }); // 1 = Monday, 7 = Sunday
```

Here are the results I got on different browsers:

- Chrome: Working as expected
image

- Firefox: Issue encountered
image

- Edge: Working as expected
image

I don’t have access to an Apple device, so I couldn't test on Safari.

This is my first time solving these kinds of problem, so I would greatly appreciate any guidance or feedback.

Apologies for the bad english, it is my third language. Thanks in advance!

pjeweb:
Thank you for investigating further! That makes a lot of sense. Thats a bit sad that firefox does not support it, but I think we have a two options:
1) use a separate library or polyfill for the week info like https://github.com/gamtiq/weekstart or https://github.com/bart-krakowski/get-week-info-polyfill
2) or honestly i would just keep our own small list of countries, we know start on sunday/staturday. I think with `US`, `CA`, `JP`,`IL` (Israel) and `IR` we should cover most cases?! We could do a simple feature detection to use the Intl weekInfo if available and fall back to our own list if not.

jateez:
Thank you for the feedback!

I'll explore both options and, once I've had a chance to test them, I'll implement the one that seems to be the best fit.

apps/website/src/components/calendar/Calendar.tsx
```diff
@@ -9,6 +9,8 @@ import {
 import { Transition } from "@headlessui/react";
 import { DateTime } from "luxon";
 
+import { ListBucketInventoryConfigurationsOutputFilterSensitiveLog } from "@aws-sdk/client-s3";
```
MattIPv4, line 12:
```suggestion
```
Auto-import did a bad?

apps/website/src/components/calendar/Calendar.tsx
```diff
@@ -9,6 +9,8 @@ import {
...
+import { getWeekStartByLocale } from "weekstart/full";
```
MattIPv4, line 13: Nit, can the linebreak be after this, not before it? So we're grouping external vs. internal imports?

MattIPv4:
Looking at https://github.com/gamtiq/weekstart/blob/master/src/fullLangRegionMap.js vs. https://github.com/bart-krakowski/get-week-info-polyfill/blob/main/src/weekInfoData.ts, I'm suspecting we might want to use the latter -- it also acts as a proper polyfill, so if the browser has the method, it will use that rather than the polyfill with its pre-determined (and potentially not correct) data?

jateez:
yeah, It does seem like the get-week-info-polyfill might be the better option

MattIPv4 (approved):
LGTM

MattIPv4:
Thanks for all the work on this 💚

jateez:
Thank you so much @pjeweb and @MattIPv4 for all the help and guidance, I really appreciated it!🙏🏻

PR #790, opened 2024-10-03
Community map

### LINKED ISSUE #668: [Idea]: Community map (maybe Show & Tell)

It would be nice to have a map of the alveus community. This could be collected when posting show & tell posts. Original idea by Stolen87 in chat.

Community engagement, showing global reach of Alveus.

fooji23:
## Describe your changes

Resolves #668 by adding a map to the S&T creation form. Users can now search for locations by name, reverse search by coordinates, and select a location on the map to share their post location. Additionally, a Community Map has been added to the S&T main page, displaying all the locations and simple data about unique locations and countries.

This feature is nearly finished, and I'm adding the final touches, but I've encountered a bug that I can't seem to fix. Here are the steps to reproduce it:

1) Create a post with a location.
2) Go to "My Posts," edit that post, and select another location. You can drag the marker or use the search box to update it; both methods work. Save the updated post.
3) Return to "My Posts" and edit the post a second time. The location displayed is the one you selected in the first step, not the updated one. You must refresh the map (click on the checkbox "Add post location" twice to hide and then show the map) for the updated location to appear.

This issue only occurs from the user perspective. When reviewing the post as an admin, you can edit the location, and it will show the updated location the next time you review that post.

If you spot any rough edges, please let me know, and I’ll address them. I've also tried to document everything thoroughly, but if something isn’t clear, just let me know, and I will add more information.

Cheers!


apps/website/src/components/shared/form/MapPickerField.tsx
```diff
@@ -0,0 +1,299 @@
...
+export const MapPickerField = ({
+  defaultLocation,
+  initiallyHidden = defaultLocation?.location ? false : true,
+  textToShow = "Show map",
+  minZoom = 1,
+  maxZoom = 24,
+  initialZoom = defaultLocation?.location ? maxZoom : 1,
+  center = defaultLocation?.location
+    ? [defaultLocation.longitude, defaultLocation.latitude]
+    : [0, 0],
```
pjeweb, line 70: This will create a new array on each render and since we never pass a `center` prop anyways, i would move calculating the center into the creation effect (see below).

apps/website/src/components/shared/form/MapPickerField.tsx
```diff
@@ -0,0 +1,299 @@
...
+   */
+  const createMap = useCallback(() => {
+    console.log("CREATING MAP");
+    if (mapContainerRef.current) {
+      const map = new Map({
+        container: mapContainerRef.current,
+        style: {
+          version: 8,
+          name: "Alveus Map",
+          center,
```
pjeweb, line 97: Moving this here from the function parameters (see above).

```suggestion
          center: defaultLocation?.location
              ? [defaultLocation.longitude, defaultLocation.latitude]
              : [0, 0],
```

apps/website/src/components/shared/form/MapPickerField.tsx
```diff
@@ -0,0 +1,299 @@
...
+
+  const mapContainerRef = useRef(null);
+  const mapRef = useRef<Map | null>(null);
+  const markersRef = useRef<Marker[]>([]);
+  const isDraggingRef = useRef(false);
+
+  /**
+   * Creates a map with the props provided to the Component
+   */
+  const createMap = useCallback(() => {
```
pjeweb, line 89: I do not see the benefit of moving this into its own function. If you just keep this in the other effect that should be fine. That will also reduce the indentation one level since you have the early return in the other effect.

But this is option. Just make sure all deps are set here then (see comment below).

apps/website/src/components/shared/form/MapPickerField.tsx
```diff
@@ -0,0 +1,299 @@
...
+    [coordsPrecission, allowMultipleMarkers, onLocationChange],
+  );
+
+  useEffect(() => {
+    // Ensure map container exists and map hasn't been initialized yet
+    if (!showMap || !mapContainerRef.current || mapRef.current) {
+      return;
+    }
+
+    const map = createMap();
```
pjeweb, line 235: Inline the creation here.

```suggestion
    const map = new Map({
```

apps/website/src/components/shared/form/MapPickerField.tsx
```diff
@@ -0,0 +1,299 @@
...
+      }
+    }
+
+    // Clean up on component unmount
+    return () => {
+      handleLocationClear();
+      mapRef.current?.remove();
+      mapRef.current = null;
+    };
+  }, [showMap]);
```
pjeweb, line 254: Assuming you apply the suggestions and move the map creation code into this effect, these should be all the dependencies and they should all be stable.

```suggestion
  }, [
    antialias,
    defaultLocation,
    handleLocationClear,
    handleLocationSet,
    initialZoom,
    maxZoom,
    minZoom,
    showMap,
  ]);
```

- `antialias`, `initialZoom`, `maxZoom`, `minZoom` and `showMap` are primitive values so we do not need to do anything special
- `defaultLocation` -> stabilized in the parent component
- `handleLocationClear`, `handleLocationSet` -> stabilized with useCallback

pjeweb (changes requested):

pjeweb:
One more thing that makes the map reload that should be fixed is that the form/map will rerender on window focus. We can fix that by configuring the trpc query in `apps/website/src/components/show-and-tell/ShowAndTellEntryForm.tsx`:

```typescript
  const getMyPost = trpc.showAndTell.getMyEntry.useQuery(String(postId), {
    enabled: Boolean(postId),
    refetchOnReconnect: false,
    refetchOnWindowFocus: false,
  });
```

pjeweb (approved):
Amazing work!

Issue #787, opened 2024-10-02
[Idea]: Go through collaborations and update popularity (followers)


pjeweb:
### Describe your idea / enhancement

Go through the collaborations and update follower numbers. Most of them are very similar still, but some made big jumps. https://github.com/alveusgg/alveusgg/blob/main/apps/website/src/data/collaborations.ts#L67

### How would this idea / enhancement be beneficial?

Better highlight collaborations of (now) popular content creators.

shradiphylleia:
@pjeweb just a quick check.... would you want me to manually change them or maybe write script to update them dynamically (have not figured this out just yet but this might be a good idea)

pjeweb:
> @pjeweb just a quick check.... would you want me to manually change them or maybe write script to update them dynamically (have not figured this out just yet but this might be a good idea)

I'm not sure a script is worth the effort 🤷  I think we mostly have Twitch and Youtube content creators, but ideally we'd also check IG, TikTok etc. for their biggest "reach". I'd accept either, I think, but social media APIs tend to break or change every now and then. So maybe we just do it manually every 6 months or so.

shradiphylleia:
@pjeweb thanks for the reply. the most popular of their socials should work about fine then. Later maybe looking for a 3rd party api which could fetch the info might work but for now I will do it manually based on the most popular of their social. Let me know if it works.

MattIPv4:
You can probably just use whichever social platform is currently referenced for each one, just update the actual number and the date it was updated :)

shradiphylleia:
@MattIPv4 Done with the rest.
Would need clarity regarding the collaborator Jack Manifold they have two entries:
https://github.com/alveusgg/alveusgg/blob/70eec19868987f928c9a24f5888ae463bc7b1d93/apps/website/src/data/collaborations.ts#L647  

https://github.com/alveusgg/alveusgg/blob/70eec19868987f928c9a24f5888ae463bc7b1d93/apps/website/src/data/collaborations.ts#L237 (recent collab)

should I do away with the previous collab entry? let me know and I will raise a PR. thanks

MattIPv4:
No, he is in there twice because he has had two collaborations (as the comment explains), but we only store his creator details for the most recent one so he doesn't show up in the creator list twice.

PR #792, opened 2024-10-05
Go through collaborations and update popularity (followers) #787

### LINKED ISSUE #782: [Idea]: Sync calendar with Twitch Stream Schedule

We should maybe revisit the twitch schedule feature. i was under the impression, it only works for recurring events, but it looks like you can put one-off streams on there.

https://dev.twitch.tv/docs/api/reference/#create-channel-stream-schedule-segment

Helps users using the twitch schedule find upcoming streams/events

shradiphylleia:
## Describe your changes
 resolves #782  -->Go through collaborations and update popularity (followers) #787
...
check the updated follower numbers on the collaborations page
...

### DIFF
--- a/apps/website/src/data/collaborations.ts
+++ b/apps/website/src/data/collaborations.ts
@@ -76,7 +76,7 @@ const collaborations: Collaboration[] = (
         {
           name: "itmeJP",
           image: itmeJP,
-          popularity: 370_000, // Twitch followers, 2024-08-02
+          popularity: 371_000, // Twitch followers,  2024-10-04
         },
       ],
     },
@@ -90,7 +90,7 @@ const collaborations: Collaboration[] = (
         {
           name: "Agent00",
           image: agent00,
-          popularity: 1_800_000, // Twitch followers, 2024-07-31
+          popularity: 2_000_000, // Twitch followers, 2024-10-04
         },
       ],
     },
@@ -103,7 +103,7 @@ const collaborations: Collaboration[] = (
         {
           name: "Bao",
           image: bao,
-          popularity: 540_000, // Twitch followers, 2024-06-13
+          popularity: 561_000, // Twitch followers, 2024-10-04
         },
       ],
     },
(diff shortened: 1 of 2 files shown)


shradiphylleia:
not sure what is blocking the ability to merge can the maintainers please let me know what could have been wrong. i did close my previous one since the node version i was using did not match with the required version. i think it is because of the package-lock.json on my end.

pjeweb:
@shradiphylleia hey! there seems to be nothing wrong with the PR. If you create the PR there are a few check that will prevent merging, but it needs a maintainer of the repo (like matt and me) to allow the checks to run. Just keep the PR open long enough for us to let them run ;)

Two things I noticed already is that you seem to have use `npm` which has written a `package-lock.json`. If you can use `pnpm` instead or just exclude that file from the PR that would be great.

Also I wonder why you switched from yt to twitch for some of the creators even though their YT has the bigger number, for example MoistCr1TiKaL (https://www.youtube.com/user/penguinz0).

shradiphylleia:
thanks for helping me out. I was really confused.

addressing the second thing I just went with the link which was mentioned for them if I was unsure of the Twitch account or other socials. For instance, it was 2 collaborator I was not able to find a matching twitch account(same username) so I went with what info I had at hand. [for the said creator it was their twitch link which was mentioned]  https://github.com/alveusgg/alveusgg/blob/70eec19868987f928c9a24f5888ae463bc7b1d93/apps/website/src/data/collaborations.ts#L634

will change that too in new PR

> @shradiphylleia hey! there seems to be nothing wrong with the PR. If you create the PR there are a few checks that will prevent merging, but it needs a maintainer of the repo (like matt and me) to allow the checks to run. Just keep the PR open long enough for us to let them run ;)
> 
> Two things I noticed already is that you seem to have use `npm` which has written a `package-lock.json`. If you can use `pnpm` instead or just exclude that file from the PR that would be great.
> 
> Also I wonder why you switched from yt to twitch for some of the creators even though their YT has the bigger number, for example MoistCr1TiKaL (https://www.youtube.com/user/penguinz0).

PR #799, opened 2024-10-15
Fix NWF link on Voters Guide

ComradeRamen:
Update the link to National Wildlife Federations Voters Guide.


pjeweb:
@ComradeRamen Thank you for the fix! Out formatter check is not happy with the longer link. Do you want to fix that yourself? You'd have to check it out locally and install and run prettier. Otherwise I'm happy to push the format change to your PR :)

ComradeRamen:
> Otherwise I'm happy to push the format change to your PR :)

Please do.
Thanks.

pjeweb:
@MattIPv4 I wonder if it makes sense to still link to their page. The original guide is not there anymore.

This is what I think we based our guide on: https://web.archive.org/web/20240404075824/https://www.nwf.org/Home/VOTE

MattIPv4:
Yeh, this link was added moreso as credit rather than an additional resource for folks. If the page no longer exists, might be best to just remove the link and leave the text, or remove the credit altogether (not sure that feels right)?

MattIPv4:
@ComradeRamen Would you be willing to update this PR to have `National Wildlife Federation` be linked to the NWF homepage, instead of linking to their voters guide that doesn't exist anymore? Lmk if not and I can push the change instead.

ComradeRamen:
Done.
Also apologies for not checking and mentioning that it's a different page and not just moved to a different url.

MattIPv4 (approved):

PR #803, opened 2024-10-17
Add Vercel track call to short links

MattIPv4:
## Describe your changes

To give us better visibility into the clicks through the short link system, hook it into the Vercel analytics system that we have access to with the site.

Per https://github.com/vercel/analytics/issues/88, it looks like we can pass through the request headers to improve what data Vercel can show.

Vercel now tracks short link clicks as custom events.

### DIFF
--- a/apps/website/src/pages/api/short-links/track-click.ts
+++ b/apps/website/src/pages/api/short-links/track-click.ts
@@ -8,20 +8,21 @@ const trackClickSchema = z.object({
   id: z.string().cuid(),
   slug: z.string(),
   link: z.string(),
+  headers: z.record(z.union([z.string(), z.array(z.string())])),
 });
 export type TrackClickSchema = z.infer<typeof trackClickSchema>;
 
 //API for short links tracking
 export default createTokenProtectedApiHandler(
   trackClickSchema,
-  async ({ id, slug, link }) => {
+  async ({ id, slug, link, headers }) => {
     try {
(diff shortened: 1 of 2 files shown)


apps/website/src/pages/api/short-links/track-click.ts
```diff
@@ -1,20 +1,27 @@
...
   res: NextApiResponse,
 ) {
-  const id = req.body.id;
+  const { id, slug, link } = req.body;
   await prisma.shortLinksTracking.upsert({
     where: { id: id },
     create: { id, clicks: 1 },
     update: { clicks: { increment: 1 } },
   });
+  await track("Short link clicked", { id, slug, link });
```
MattIPv4, line 25: Looking at https://vercel.com/docs/analytics/limits-and-pricing#limits-per-plan, it looks like we can't even track id + slug + link here, as we're limited to two properties

MattIPv4:
I wonder if, much as it might stress the DB a bit, the best thing to do is have a timeseries DB that stores a record per click? The Next.js request we get contains all the geo info so we can just write that to the DB?

Alternatively, we could try making this a _page_ rather than middleware, so we can make use of the standard Vercel analytics, similar to how notifications work.

@pjeweb thoughts?

MattIPv4:
It looks like it might be working 🎉 

image

pjeweb (approved):
Thats looking very good!

PR #811, opened 2024-10-30
Validate + refresh Twitch auth tokens for accounts

### LINKED ISSUE #450: [Bug]: Users getting stuck with active session connected to deleted twitch account

Users with an existing session that delete their twitch account (or get banned) can't log out and log in with their new account.

When the twitch account is deleted and a user returns, they should be logged out and possibly the account be marked as deleted.

### Steps to reproduce

- Login with account A
- Delete twitch user connected to account A (or get it banned)
- Without deleting cookies return to the site

### Browser specs

n/a

### Anything else?

Very edge case. Probably easier to send them a link to delete cookies for that case. "Try another browser" apparently did not work in the one case, deleting the user in our DB did.

MattIPv4:
## Describe your changes

Should resolve #450 as we are continually validating and attempting to refresh the Twitch token. If the token is expired for any reason and we fail to refresh, we'll force the session to be signed out.

In preparation for #782, this also adds a cron job that more proactively refreshes the Twitch tokens for configured accounts in the DB -- if we're getting near the expiry we'll attempt a refresh, and if not, we'll run the same validate/refresh check as with sessions.

Authenticate website locally. While authenticated, in a new tab, access your Twitch settings and disconnect the OAuth app. Observe signed out of website.

Configure a Twitch channel and link your account to it in the admin UI. Wait for the expiry to get within 1hr, or update the timestamp in the DB to be within 1hr of now. Hit the cron endpoint with auth. Observe token refreshed and expiry time updated in admin UI.

### DIFF
--- a/apps/website/src/pages/api/auth/[...nextauth].ts
+++ b/apps/website/src/pages/api/auth/[...nextauth].ts
@@ -7,6 +7,10 @@ import { env } from "@/env";
 import { prisma } from "@/server/db/client";
 import { getRolesForUser } from "@/server/db/users";
 import { checkIsSuperUserId } from "@/server/utils/auth";
+import {
+  ExpiredAccessTokenError,
+  refreshAccessToken,
+} from "@/server/utils/oauth2";
 import { defaultScope } from "@/data/twitch";
 
 const adapter = PrismaAdapter(prisma);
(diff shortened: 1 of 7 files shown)


MattIPv4:
I've pushed this to the preview branch for testing w/ auth

apps/website/src/pages/api/auth/[...nextauth].ts
```diff
@@ -27,9 +31,33 @@ type ProfileData = {
...
+      if (token?.access_token && token?.refresh_token) {
+        try {
+          await refreshAccessToken(
+            "twitch",
+            env.TWITCH_CLIENT_ID,
+            env.TWITCH_CLIENT_SECRET,
+            token.id,
+            token.access_token,
+            token.refresh_token,
+          );
```
Comment (pjeweb) [inline apps/website/src/pages/api/auth/[...nextauth].ts:50]: I am not sure we should run refreshAccessToken which will validate the token with twitch every time the session is accessed!?

But I am also not sure whats the best approach. Maybe we need to keep track of the last token update and check `validated_at + 1h < now` or something?

Comment (MattIPv4) [reply apps/website/src/pages/api/auth/[...nextauth].ts:50]: Yeh, it might be a bit aggressive if folks are clicking in/out of the site continually. I wonder if we can just include a value in the session itself to track last validated? Not sure if extra data there persists or not.

Comment (MattIPv4) [reply apps/website/src/pages/api/auth/[...nextauth].ts]: Turns out you can't store data in the session that persists, so put it in the DB.

pjeweb:
Pushed the DB changes to Preview. Set up a branch to deploy to Live.

pjeweb (approved):
This is great!

PR #812, opened 2024-10-31
Sync Alveus calendar events to Twitch schedule

### LINKED ISSUE #782: [Idea]: Sync calendar with Twitch Stream Schedule

MattIPv4:
## Describe your changes

Resolves #782 

...


MattIPv4:
Asked in the TwitchDev Discord about getting access to non-recurring events for testing: https://discord.com/channels/504015559252377601/523675960797691915/1301362926095175731

MattIPv4:
@pjeweb this should be good for review now -- unsure where, if at all, we want to have this running in preview (or if we should just leave it for local dev + prod).

MattIPv4:
Two follow-up items once we know this works in prod:

- Add a redirect for the Twitch iCal + link it on the page (maybe /updates/ical)
- Abstract the syncing logic to support Maya's channel/events (and add a redirect for /updates/ical/maya)

pjeweb (changes requested):

pjeweb (approved):

PR #814, opened 2024-10-31
Add MMCQ util for dominant color on Show & Tell posts

### LINKED ISSUE #754: [Idea]: Show and Tell: Make presentation mode more colorful

Right now the slides in presentation mode use the first image as a blurred background plus some green. Maybe we can make it more colorful (instead of mostly green/grey) by picking the dominant color and tinting the background with that instead of green?

More better look Smile

matadaniel:
modified mean cut quantizer by Leptonica

resolves #754


MattIPv4 (commented):
👋 Thanks for doing this -- two main initial thoughts:

1. Could we abstract this into a dedicated util file, so we're not bloating the component itself?

2. Could we perhaps add some comments to the logic so its easier to follow what its doing to determine the color?

MattIPv4:
And I suppose as a related Q, why could we not use something like https://www.npmjs.com/package/image-q to avoid having all this logic in our own repo that we need to maintain *yes, lots of dependencies is bad, but I feel something like quantization is probably better out-sourced to a well-maintained library)?

matadaniel:
Yeah, I didn't want to introduce a dependency. I hadn't come across image-q while searching for a way to extract the dominant color of an image. I used color-thief as inspiration for my implementation. I made it type safe and used more modern js features.

matadaniel:
I tried to write "self-documenting code". Which parts of the logic are most unclear?

MattIPv4:
> I tried to write "self-documenting code". Which parts of the logic are most unclear?

I think it's less about documenting what each line of the code is doing, but more-so documenting the principles being applied in each block of code. As someone not familiar with MMCQ as a topic, sure I can read the syntax here, but I have no idea what the high-level principles are that are at play, and that makes me very hesitant to want to merge it as I then am the maintainer of it.

matadaniel:
That is completely reasonable. I added some things to improve readability.

matadaniel:
Before I forget: I use oklch to adjust the dominant color's lightness to ensure there is enough contrast with the text. It is currently supported by [86% of users](https://caniuse.com/mdn-css_types_color_oklch_relative_syntax). Is that high enough?

MattIPv4:
@pjeweb would appreciate your thoughts on this 👀

pjeweb:
I do the like the effect!

I wonder if this is something we can go once when creating/updating the post? We already resize the image with canvas before uploading. Maybe we can pass the dominant color there and save it in the database?

See https://github.com/alveusgg/alveusgg/blob/752b822c6f7e33c14a61c63a62da05cec0acd42a/apps/website/src/components/shared/form/UploadAttachmentsField.tsx#L231 and https://github.com/alveusgg/alveusgg/blob/752b822c6f7e33c14a61c63a62da05cec0acd42a/apps/website/src/utils/resize-image.ts#L31

I think I would be fine with the code if we maybe attribute the original author/"inventor". I did not find a decent looking typescript port. 

Looks like it originally stems from this: https://github.com/DanBloomberg/leptonica/blob/master/src/colorquant2.c

You explanation of steps is fine, but I think they are not really enough to explain the algorithm.

matadaniel:
Is there a "proper" way to attribute the author? Should I include the copyright notice from the original file?

The color extracted this way is always off by 2 or less in each channel compared to the previous method. I think this is more interesting than concerning.

What is the workflow of approving posts? The moderator needs to make sure the featured image is acceptable.

matadaniel:
No worries! The attribution question still stands. Other than that, I think I found the answer to the approval question and I implemented a solution for it along with the solutions for the previous requests.

MattIPv4:
> Is there a "proper" way to attribute the author? Should I include the copyright notice from the original file?

That's essentially what we did for https://github.com/alveusgg/alveusgg/tree/main/apps/website/src/third-party/qrcodegen

matadaniel:
I already added some attribution to the mmcq.ts file. It is basically equivalent to the one in icons. I just want to make sure I give credit properly.

MattIPv4:
👍 I think what you have is probably fine, unless there's a specific repo/file we want to point to as the source.

@pjeweb any concerns from you on that front?

MattIPv4:
Would you be able to resolve the conflicts and type errors so that we can get this tested + reviewed?

matadaniel:
Done!

pjeweb:
> 👍 I think what you have is probably fine, unless there's a specific repo/file we want to point to as the source.
> 
> @pjeweb any concerns from you on that front?

no i think this is fine

apps/website/prisma/schema.prisma
```diff
@@ -458,15 +459,18 @@ model ImageMetadata {
...
+  title               String  @db.VarChar(200)
+  alternativeText     String  @db.VarChar(300)
+  caption             String  @db.VarChar(200)
+  description         String  @db.VarChar(200)
+  url                 String  @db.VarChar(200)
+  fileStorageObjectId String  @unique
+  entryId             String? @unique
+  dominantColor       String?
+
+  entry                      ShowAndTellEntry?           @relation(fields: [entryId], references: [id], onDelete: Cascade)
```
MattIPv4, line 473: Why do we need to store an entryId + entry relation -- you can already get this via the showAndTellEntryAttachment relation?

apps/website/src/utils/video-urls.ts
```diff
@@ -52,7 +52,12 @@ export const videoPlatformConfigs = {
     regex: /(?:https?:)?(?:\/\/)?(?:(?:www|i)\.)?imgur\.com\/([a-z0-9]+)/i,
     normalizedRegex: /https:\/\/imgur\.com\/([a-z0-9]+)/,
   },
-} as const;
+} as const satisfies {
+  [key: string]: {
+    previewUrl?: (id: string) => `${string}.jpg`;
+    [key: string]: unknown;
```
MattIPv4, line 58: Can we write a full type here instead of just having it being unknown (ack that this is a const so the actual type is correct already)?

matadaniel: I didn't want any future changes to the object to need to be repeated on the type when it is a const. Maybe the type validation should be in `ShowAndTellEntryForm`.

MattIPv4: Ah I see -- I'd be fine with definitely a full type (using `as const satisfies <type>`) here so that we know the const has a consistent set of values (I would think most logic that consumes this const relies on that fact already even though it isn't typed as such).

apps/website/prisma/schema.prisma
matadaniel: I'm probably missing something, but what I've tried gives me errors until I define another relation anyway.
```
featuredImage   ShowAndTellEntryAttachment?
```

MattIPv4: Ohhh, no I missed what this was doing -- you're using `ImageAttachment` directly from `ShowAndTellEntry` rather than via `ShowAndTellEntryAttachment`. Would it make sense to make it use `ShowAndTellEntryAttachment` for consistency? I'm worried it'd be confusing with some `ImageAttachment`s being directly related to the `ShowAndTellEntry` and some not?

matadaniel: I tried to change it to `ShowAndTellEntryAttachment`, and it creates an ambiguous relation.

MattIPv4: I imagine you'll need to store a featuredImageId on the entry so that you can disambiguate the relationship and know which specific attachment is the featured one -- the relationship is currently just defined by all the attachments having the entry id

matadaniel: What about:
```prisma
// ShowAndTellEntry
  featuredImage ShowAndTellEntryAttachment?  @relation("FeaturedImage")

// ShowAndTellEntryAttachment
  featuredImageEntryId String? @unique

  featuredImageEntry ShowAndTellEntry? @relation(name: "FeaturedImage", fields: [featuredImageEntryId], references: [id], onDelete: Cascade)
```

MattIPv4: I'm confused why the Attachment would define the relation? The Entry is the one that needs to go "hey that attachment with that id is the featured image attachment"?

MattIPv4: @pjeweb am I off base here?

matadaniel: So something like:
```prisma
// ShowAndTellEntry
  featuredImageId     String?   @unique

  attachments   ShowAndTellEntryAttachment[] @relation("Attachments")
  featuredImage ShowAndTellEntryAttachment?  @relation(name: "FeaturedImage", fields: [featuredImageId], references: [id], onDelete: NoAction, onUpdate: NoAction)

// ShowAndTellEntryAttachment
  entry              ShowAndTellEntry  @relation(name: "Attachments", fields: [entryId], references: [id], onDelete: Cascade)
  featuredImageEntry ShowAndTellEntry? @relation("FeaturedImage")
```

MattIPv4: I don't think `ShowAndTellEntryAttachment` would need anything, you'd just need to define it in `ShowAndTellEntry`?

matadaniel: Unless I'm missing something, that would create an ambiguous relation.

matadaniel: Another approach could be storing the featured image in the attachments and giving it a "featuredImage" attachmentType. I don't really think it belongs there though, and it would no longer be a constant time lookup.

MattIPv4: Hm yeh. Would agree with that. I'm just confused why this is still resulting in an ambiguous relationship -- we're storing an explicit ID so it shouldn't.

What happens if we go back to the idea of using `ImageAttachment` directly without `ShowAndTellEntryAttachment`, but define the relationship only from `ShowAndTellEntry` with a `featuredAttachmentId`?

I think my issue with the implementation as you've committed is that we're introducing a new relationship into `ImageAttachment` with `entry` that I really worry is going to trip us up down the road -- it looks like `entry` is always going to be populated (and it makes sense it would be given it is always tied to an entry via the attachment intermediary model), but actually won't be as that relationship is direct to the entry only used for the featured image.

matadaniel: I can definitely see that causing confusion. There would still need to be an opposite relation field on `ImageAttachment` with the relation defined on `ShowAndTellEntry`, no?

pjeweb: Sorry for not looking at this earlier. But do we actually need the featured image in the database? I think the function to figure out the first image/video with thumbnails is not too expensive, so we can just run it if needed? Or did we want the featured image to be selectable?

matadaniel: If we don't store the featured image, should the dominant color be a field on `ShowAndTellEntry`? Does a video url always respond with the same thumbnail? YouTube videos can have multiple thumbnails.

MattIPv4: I actually don't know how the thumbnail URL works with the a/b testing feature now -- I would guess you get a consistent image though.

Storing the color directly on the entry might be easiest, so its still computed only once. You could even have a field that only shows in review mode (like mod comments) to allow mods to manually change the color if needed.

matadaniel: I'll make the dominant color a field on `ShowAndTellEntry`, and I'll extract the featured image selection logic to try to guarantee the color will match the image.

Why would a color need to be changed? Just curious

MattIPv4: I can't imagine we'll ever need to actually change it, just planning ahead for if the logic inadvertently picks a really bland colour or smth for a post (not that it should given what its doing)

matadaniel: I could add a minimum chroma to the color. Either way, I implemented a very basic review field with DominantColorFieldset.tsx.

apps/website/src/components/show-and-tell/ShowAndTellEntry.tsx
```diff
@@ -293,13 +272,16 @@ export const ShowAndTellEntry = ({
...
+    splitAttachments(entry.attachments);
+
   return (
     <article
       key={entry.id}
       className={classes(
         "relative flex flex-shrink-0 flex-col transition-opacity delay-500 duration-500 focus:outline-none",
         isPresentationView &&
-          "h-[calc(100svh-6em)] h-[calc(100vh-6em)] w-[80%] snap-center overflow-hidden bg-alveus-green text-white shadow-xl",
+          "h-[calc(100svh-6em)] h-[calc(100vh-6em)] w-[80%] snap-center overflow-hidden bg-alveus-green text-white shadow-xl transition-[background-color,_opacity]",
```
MattIPv4, line 284:
```suggestion
          "h-[calc(100svh-6em)] h-[calc(100vh-6em)] w-[80%] snap-center overflow-hidden bg-alveus-green text-white shadow-xl transition-[background-color,opacity]",
```
Just a nit

apps/website/src/components/show-and-tell/ShowAndTellEntryForm.tsx
```diff
@@ -210,6 +229,24 @@ export function ShowAndTellEntryForm({
       }
     }
 
+    if (data.dominantColor === null) {
+      const { featuredImage } = splitAttachments(
+        // @ts-expect-error - only essential properties are defined
```
MattIPv4, line 234: If these are the only essential properties for the method, can we update the type on the method to only need those properties?

apps/website/src/components/show-and-tell/ShowAndTellEntryForm.tsx
```diff
@@ -200,6 +215,10 @@ export function ShowAndTellEntryForm({
         ),
       };
 
+      if (data.dominantColor === null) {
```
MattIPv4, line 218:
```suggestion
      if (!data.dominantColor) {
```
Nit

apps/website/src/components/show-and-tell/ShowAndTellEntryForm.tsx
```diff
@@ -163,10 +169,18 @@ export function ShowAndTellEntryForm({
     { allowedFileTypes: imageMimeTypes },
   );
 
-  const handleSubmit = (e: FormEvent<HTMLFormElement>) => {
+  const handleSubmit = async (e: FormEvent<HTMLFormElement>) => {
     e.preventDefault();
     const formData = new FormData(e.currentTarget);
     const hours = parseFloat(formData.get("giveAnHour") as string);
+    let dominantColor = formData.get("dominantColor") as string | null;
+    if (dominantColor !== null) {
```
MattIPv4, line 177:
```suggestion
    if (dominantColor) {
```
Nit

apps/website/src/components/show-and-tell/ShowAndTellEntryForm.tsx
```diff
@@ -210,6 +229,24 @@ export function ShowAndTellEntryForm({
       }
     }
 
+    if (data.dominantColor === null) {
```
MattIPv4, line 232:
```suggestion
    if (!data.dominantColor) {
```
Nit

apps/website/src/components/show-and-tell/ShowAndTellEntryForm.tsx
```diff
@@ -210,6 +229,24 @@ export function ShowAndTellEntryForm({
...
+        // @ts-expect-error - only essential properties are defined
+        videoLinksData.videoUrls.map((url) => ({
+          attachmentType: "video",
+          linkAttachment: {
+            url,
+          },
+        })),
+      );
+
+      if (typeof featuredImage !== "undefined") {
```
MattIPv4, line 243:
```suggestion
      if (featuredImage) {
```
Nit

apps/website/src/server/db/show-and-tell.ts
```diff
@@ -149,6 +150,7 @@ const showAndTellSharedInputSchema = z.object({
   location: z.string().max(MYSQL_MAX_VARCHAR_LENGTH).nullable(),
   longitude: z.number().nullable(),
   latitude: z.number().nullable(),
+  dominantColor: z.string().min(5).max(11).nullable(),
```
MattIPv4, line 153: Worth using a pattern here to ensure we have a comma-separated value of digits and not some arbitrary other string?

apps/website/src/components/show-and-tell/ShowAndTellEntryForm.tsx
matadaniel: The function's return type is nicely derived from the parameter type. I'd rather fill the argument with dummy properties if anything.

MattIPv4: Ah okay. Yeh I'd like us to avoid relying on expecting TS errors, please -- let's have the correct types in place or the correct data

MattIPv4:
@pjeweb would you be able to take a look at this and get the schema change onto preview for testing?

matadaniel:
Currently, any update will recalculate the dominant color and overwrite any saved color, even if it was manually changed by a mod. Even mod updates require the color input to be set to the saved color if they want to keep it.

pjeweb:
> Currently, any update will recalculate the dominant color and overwrite any saved color, even if it was manually changed by a mod. Even mod updates require the color input to be set to the saved color if they want to keep it.

i mean thats not ideal, but i dont really care about the override rn. we might need to test a few images and see, but if at all possible i'd stick with the automatic color.

also thank you again for working on this! its deployed on preview.alveus.gg right now

matadaniel:
No problem! Hope you like the results!

MattIPv4 (approved):
@pjeweb would appreciate a +1 from you on this when you have a chance to look (and then shipping the prod schema change), ty.

PR #818, opened 2024-11-06
Use optimized webm for Peepo updates animation

### LINKED ISSUE #698: [Idea]: Replace Animated Webp Bell Peepo on Updates Page With Video or Still Image

MattIPv4:
## Describe your changes

Resolves #698 by swapping the webp for a webm (massively reduced file size + more optimal playback as a looping video).

I also added partial webm support into the video loader, so that we can generate a poster image for this webm (I did not add support for processing the actual video, so it just passes the raw through).

Peepo continues to be animated on the updates page, ideally more performant than production.


pjeweb (approved):
pogg

PR #819, opened 2024-11-06
Support /go/ for short links alongside /l/

MattIPv4:
## Describe your changes

Recently we've seen some users confused by `/l/` vs. `/I/`, so to help disambiguate that this adds support for `/go/` for short URLs, and uses that as the default in the UI.

Accessing `/l/<link>` continues to work, as does accessing `/go/<link>`. In both cases, a click should be tracked in the DB + Vercel.


pjeweb (approved):
lets go

PR #820, opened 2024-11-07
Show progress in timeline on scroll

MattIPv4:
## Describe your changes

Using a sticky element we can make it look like the bar in the timeline is filling as you scroll down the page, making the page feel a bit more alive.

This works perfectly in Chrome, but has some visual oddities in Firefox (see https://github.com/webcompat/web-bugs/issues/143746).

Bar fills as you scroll on the about Alveus + Maya pages.


pjeweb (approved):
cute

PR #821, opened 2024-11-07
Add new location badge to Show and Tell (new country)

### LINKED ISSUE #801: [Idea]: Badge for first time new locations on Show and Tell

When a post is the first post from a new location on Show and Tell, show some sort of badge indicating that.

Given we relatively precisely track location, we'll probably want to do some sort of fuzziness on this -- we could rely on the same logic as on the page for doing the "near", and use that value to determine a new location. Or, if it's easier, we could just do this based on country, though that's definitely not as exciting.

Let's Maya, and the community, more directly celebrate the diversity and distribution of the Alveus community all over the world.

matadaniel:
resolves #801

> we could rely on the same logic as on the page for doing the "near", and use that value to determine a new location

I reused the country extraction logic because I was unsure of how to handle location edge cases. From what I could tell, there is no input validation for the location. The logic can easily be adjusted in extractInfoFromMapFeatures.


pjeweb:
Thank you for the PR. Will review it asap!

apps/website/src/pages/show-and-tell/my-posts/[postId]/preview.tsx
```diff
@@ -72,13 +75,31 @@ const PreviewShowAndTellPage: NextPage = () => {
...
+            ) : (
+              getMyPost.data && (
+                <ShowAndTellEntry
+                  entry={getMyPost.data}
+                  newLocation={getPostsFromANewLocation.data.has(
+                    getMyPost.data.id,
+                  )}
+                />
+              )
             )}
```
Comment (MattIPv4) [inline apps/website/src/pages/show-and-tell/my-posts/[postId]/preview.tsx:102]: I'm not sure I'm a fan of the refactor into a chained ternary here -- I think it was cleaner as a set of standalone conditionals

apps/website/src/pages/admin/show-and-tell/review/[entryId]/preview.tsx
```diff
@@ -39,6 +39,11 @@ const AdminPreviewShowAndTellPage: NextPage<
 
   const entry = getEntry.data;
 
+  const getPostsFromANewLocation =
+    trpc.showAndTell.getPostsFromANewLocation.useQuery();
+
+  const postsFromANewLocation = getPostsFromANewLocation.data;
```
Comment (MattIPv4) [inline apps/website/src/pages/admin/show-and-tell/review/[entryId]/preview.tsx:45]:
```suggestion
  const { data: postsFromANewLocation } =
    trpc.showAndTell.getPostsFromANewLocation.useQuery();
```
Two consts seems excessive, can we not just destructure?

MattIPv4:
@matadaniel Any update here on the comments I left? I realise PJE hasn't had a chance to look at this fully yet, but getting those resolved plus the conflict that now exists might be a good starting point?

apps/website/src/pages/show-and-tell/my-posts/[postId]/preview.tsx
Comment (matadaniel) [reply apps/website/src/pages/show-and-tell/my-posts/[postId]/preview.tsx:102]: I believe the query states (e.g. isLoading) are mutually exclusive. I wanted to ensure the user was not shown an error and an incomplete entry to avoid confusion.

Comment (MattIPv4) [reply apps/website/src/pages/show-and-tell/my-posts/[postId]/preview.tsx]: Yeh, I do believe they're mutually exclusive, and agree that the nested ternary helps enforce that. That being said, I do think the individual blocks is a lot cleaner to read than a nested ternary, and given they should be mutually exclusive from the hook already, I'd prefer sticking to the old style, please?

Comment (matadaniel) [reply apps/website/src/pages/show-and-tell/my-posts/[postId]/preview.tsx]: What do you mean by mutually exclusive from the hook?

I could create a couple variables that track the states of both queries and revert the nested ternary.

Comment (MattIPv4) [reply apps/website/src/pages/show-and-tell/my-posts/[postId]/preview.tsx]: As in, if a given hook is in the loading state, I don't _think_ it'll also be in the error state.

If one of the hooks is loading and the other is erroring, then I think we'd want to show the user both messages, not a cascading ternary where we'd just show a loading message and no error (or visa-versa).

I don't think any additional variables would be needed, your first block would just check if either is loading and show a loading message, the second checking if either are erroring and show an error message, and the third if the data is present to actually render the entry?

Comment (matadaniel) [reply apps/website/src/pages/show-and-tell/my-posts/[postId]/preview.tsx]: That's my assumption as well.

I thought it might be confusing to show a loading state and display an error, but maybe I'm overthinking it.

Comment (MattIPv4) [reply apps/website/src/pages/show-and-tell/my-posts/[postId]/preview.tsx]: I think it's probably fine, and hopefully very much an edge-case we'll never run into, so I'd prefer to lean into cleaner/easier-to-read code with the isolated blocks rather than a nested tenary :nod:

Comment (matadaniel) [reply apps/website/src/pages/show-and-tell/my-posts/[postId]/preview.tsx]: Is f37dff3 what you had in mind?

Comment (MattIPv4) [reply apps/website/src/pages/show-and-tell/my-posts/[postId]/preview.tsx]: Yes, thank you!

matadaniel:
@MattIPv4 This should be ready for review. If there's anything specific that's holding up the merge, please let me know so I can assist or make necessary changes.

MattIPv4:
👀 Will take another look -- had been waiting for @pjeweb to take a pass through this as well to double-check the DB logic and whatnot.

MattIPv4 (approved):
Would love us to refine the unique logic down the road to potentially factor in region (US state, UK county, etc.) as well as country -- this is a great starting point though!

pjeweb (approved):

PR #823, opened 2024-11-09
Use fallback image for transparent WebM in Safari

### LINKED ISSUE #698: [Idea]: Replace Animated Webp Bell Peepo on Updates Page With Video or Still Image

pjeweb:
## Describe your changes

Use fallback image for transparent WebM in Safari cause it plays back bad (frame rate and missing transparency).

Fixes #698 

This time check more browsers ;)


MattIPv4 (commented):
This doesn't seem to fix the issue on mobile? All iOS browsers seem to fail to render WebM in the same way?

pjeweb:
> This doesn't seem to fix the issue on mobile? All iOS browsers seem to fail to render WebM in the same way?

Well it does not fix it, it just does not load the video

pjeweb:
Oh you mean Chrome on iOS. So we should check OS instead of Browser, yeah?

MattIPv4:
Yeh -- all browsers on iOS are webkit under the hood, so they're all going to have the same rendering issue. So I think we need the UA check for Safari on desktop, and then detecting iOS as a whole for mobile.

MattIPv4:
Honestly part of me wonders if we should just choose a nice static frame and do away with this animation given how much of a headache it has been.

pjeweb:
> Honestly part of me wonders if we should just choose a nice static frame and do away with this animation given how much of a headache it has been.

I still like it. Having it on the major desktop browsers and android seems like a win to me.

MattIPv4 (approved):

Issue #796, opened 2024-10-11
[Idea]: Improvements on Community Map


fooji23:
### Describe your idea / enhancement

- **Searchbox and Geolocation marker behavior:** When using the searchbox or geolocation, the map should zoom in on the precise location without automatically setting a marker. The marker should only be placed when the user clicks on the map. The Community Map would still place a marker at a rounded coordinate.

- **Address formatting of searchbox results:** The address returned by the searchbox is too long, often including unnecessary details like the zip code. This can be improved by slightly changing the onResult event in the MaplibreGeocoder. Instead of using place_name, we can destructure properties: {address} from the result and call createAddress on that address property to generate a cleaner, more concise address.

- **Post creation form:** Instead of just displaying the location string, we should allow users to modify the non-country part of the address. This keeps the country data for stats while giving users more control over the displayed address, allowing for more precision or creativity. For example:

`<input type="text" placeholder={result_from_geocoding_except_country} />, {country_from_geocoding}`

- **Community map modal behavior:** When viewing a post with pictures, if a picture is maximized, clicking outside of it currently sends the user back to the map. Instead, the expected behavior is to return to the modal. Additionally, the arrows for cycling through photos don’t work. Clicking outside a maximized picture but inside the gallery should close the picture and return to the modal, rather than loading the clicked image full-screen in the same tab.

### How would this idea / enhancement be beneficial?

These changes will improve the user experience in two ways:

1.  **For posters:** It allows them to modify the address slightly without affecting the core data points.

2. **For the community map users:** It will significantly enhance the experience, as the current modal behavior is not very intuitive.

pjeweb:
Thank you! Another point brought up a couple of times was clustering multiple posts with the same or very close location so you can see how many "pins" are on the same spot and ideally with a way to open all posts there.

PR #824, opened 2024-11-10
Improve Show and Tell location (centroids & shorter names)

pjeweb:
## Describe your changes

This will make the map search use a shorter name for locations consistent with the reverse geocode and use the centroid instead of south-east bounding box coordinate.

This also makes the map zoom to the bounding box when available.

See #796

Check every type of map input still works as expected: Select on map, search by text (point of interest, address, area/region).

### DIFF
--- a/apps/website/src/components/shared/form/MapPickerField.tsx
+++ b/apps/website/src/components/shared/form/MapPickerField.tsx
@@ -172,10 +172,10 @@ export const MapPickerField = ({
           ({
             result: {
               geometry: { coordinates },
-              place_name,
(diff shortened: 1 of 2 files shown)


MattIPv4:
@pjeweb not sure if the result of this PR, but search does not seem to work for me in the submit form?

MattIPv4:
Nevermind, I take that back, it was just slow and I was pressing enter (which submits the form -- maybe it shouldn't do that)

MattIPv4 (approved):

PR #826, opened 2024-11-11
Add dots to indicate more creators

### LINKED ISSUE #648: [Idea]: Add arrows for creator selector and position relative to carousel.

MattIPv4:
## Describe your changes

Resolves #648 hopefully by adding a set of dots below the scroller to give a visual indication/hint that there are more creators that can be scrolled to.

Clicking on the dots works as expected, scrolling through the creators. Resizing to different breakpoints works as expected, with the dots adjusting.

### DIFF
--- a/apps/website/src/pages/collaborations.tsx
+++ b/apps/website/src/pages/collaborations.tsx
@@ -39,11 +39,32 @@ const creators = collaborations
   )
   .sort((a, b) => b.popularity - a.popularity);
 
-const dots = 16;
-
 const Creators = ({ className }: { className?: string }) => {
   const ref = useRef<HTMLUListElement>(null);
   const drag = useDragScroll();
+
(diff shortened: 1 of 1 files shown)


pjeweb (approved):
oh i havent sent the review. i love it!

PR #830, opened 2024-11-14
Small fixes syncing Alveus calendar events to Twitch schedule

pjeweb:
## Describe your changes

When testing #812 I found two problems:

1) The API will 404 when no segments exist (this will happen at least once). This will handle the 404 gracefully.
2) Segments are not allowed to overlap and will throw an Error. This does not actually handle this, but filters events by URL instead of category such das overlapping collaboration streams do not lead to overlapping segments.

...


MattIPv4:
Hmm; not sure we want to filter down to just Alveus events like this -- I intentionally wanted to show collaborations on here hence the title formatting logic.

I was aware that we might run into collisions every so often and I think that's fine. Perhaps we should just wrap each call to create a segment in a try/catch so it doesn't block the whole sync?

MattIPv4 (approved):

PR #842, opened 2024-11-26
Remove invalid MIME type

matadaniel:
Found out the convoluted way that a canvas converts an image to image/png when an unsupported MIME type is provided.

I'm not sure of everywhere these types are used or what knock-on effects this could have.


MattIPv4:
image

Doesn't appear to be used in many places, and all the usages seem to be for validation against a MIME type provided by the browser/user, so I don't see this change causing issues given we're removing an invalid type.

pjeweb (approved):

PR #843, opened 2024-11-26
Update Node.js + pnpm to latest versions

MattIPv4:
## Describe your changes

Updates Node.js to 22.x as this is now the active LTS version and is supported on Vercel. As Vercel does not let us set a specific version, I've opted to just define it as `22.x` everywhere to match the uncertainty we have with Vercel.

Updates pnpm to 9.x as this is the latest release and has been out for some time now.

Stack continues to work as expected, dependencies all install without issue, Vercel deployments continue to work.


MattIPv4:
The preview here was deployed using Node.js 22.x in Vercel. I've since reverted the setting back to 20.x to allow other PRs to continue building.

PR #839, opened 2024-11-23
Chatbot: Fix Planetscale drizzle driver initialization

pjeweb:
## Describe your changes

Change how we initialize the Drizzle Planetscale driver for updated deps (#836).

## Notes for testing your change

Check the bot works.

PR #845, opened 2024-11-27
Use pnpm cache built into actions/setup-node

MattIPv4:
## Describe your changes

Per https://github.com/pnpm/action-setup#use-cache-to-reduce-installation-time we can use actions/setup-node to cache pnpm dependencies.

## Notes for testing your change

pnpm dependency caching is still happening in GHA workflow runs.

PR #846, opened 2024-11-27
Run linting + type-checking in PR GitHub Actions

MattIPv4:
## Describe your changes

A thought from #839, this wasn't caught by type-checking because we don't run it -- we should!

Unfortunately, due to https://github.com/vercel/next.js/discussions/47010, we have to rely on a bit of a hack to generate the `next-env.d.ts` file that is needed for type-checking -- we lint an empty directory, which triggers next to generate the required file.

Types are checked correctly when running `pnpm run types` in either of the packages.


MattIPv4:
This will need rebasing once #845 is in

apps/website/package.json
```diff
@@ -24,10 +24,15 @@
...
+    "lint:next": "next lint",
+    "lint:prisma:validate": "prisma validate",
+    "lint:prisma:format": "prisma format --check",
+    "lint:prisma": "run-p lint:prisma:*",
+    "lint": "run-p lint:*",
+    "types": "tsc --noEmit",
     "start": "next start",
     "test": "vitest",
-    "format": "prettier . --write && prisma format"
+    "format": "prisma format"
```
pjeweb, line 35: What is the background behind removing prettier here?

I personally use prettier in the editor but having a command sounds good too unless there is a problem with it?

MattIPv4: `format` in the top-level package runs prettier -- I figured we only needed it in one place (plus the chatbot doesn't define it as `format` either).

If you prefer to have it in here as well, happy to add it back.

pjeweb: sounds good

pjeweb (approved):

PR #864, opened 2024-12-04
Fix PopoverButton clipping with anchor positioning

### LINKED ISSUE #822: [Bug]: Share thing in announcements gets cut off

The share thing that opens when being clicked is cut off by the stay updated thing.
image

The share thing should be fully shown

### Steps to reproduce

1. Open https://www.alveussanctuary.org/updates#announcements
2. Click on share

### Browser specs

https://www.whatsmybrowser.org/b/C5Q8U

### Anything else?

On smaller screens it works as expected and the share thing stays on top

alisondy:
## Describe your changes

Resolves #822 by relying on modern anchor positioning instead of absolute positioning

Applies to all usages of popover button, not just the announcement share button

- Pull branch down locally
- Run website
- Create test announcement
- Check share button works in Firefox (Uses fallback js positioning)
- Check share button works in Chrome (Uses anchor positioning)

### DIFF
--- a/apps/website/src/components/shared/AddEventButton.tsx
+++ b/apps/website/src/components/shared/AddEventButton.tsx
@@ -1,5 +1,5 @@
 import { Fragment } from "react";
-import { PopoverButton as PopeoverButtonHeadless } from "@headlessui/react";
+import { PopoverButton as PopoverButtonHeadless } from "@headlessui/react";
 
 import { env } from "@/env";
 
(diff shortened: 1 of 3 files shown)


MattIPv4 (approved):
Firefox:

| Preview | Production |
|-|-|
| image | image |

Chrome (with chrome://flags/#web-share disabled):

| Preview | Production |
|-|-|
| image | image |

MattIPv4:
As a nice side benefit, because this is an anchor, it'll move itself when there isn't enough space on the page:

https://github.com/user-attachments/assets/2e361513-aa35-424f-81b1-813e6880b48a

Looks like that even works in Firefox where headlessui relies on a JS polyfill.

PR #887, opened 2024-12-11
Update to Next.js 15.1.0

MattIPv4:
## Describe your changes

Updates Next.js manually as dependabot couldn't do it with the patch. We can finally remove the patch now as this version include the Satori + @vercel/og bump with the fix.

This version also makes React 19 stable, so we should consider updating. An experimental feature for inlining CSS is also available which might be worth exploring.

Site continues to build.


pjeweb (approved):
nice

Issue #565, opened 2024-02-27
[Idea]: Show and tell admin: Open videos and pictures on click


pjeweb:
### Describe your idea / enhancement

It would help if you could open video links and pictures in the review/edit view of show and tell entries. This probably also applied to the creation of posts. They could either open in a modal or just in a new tab.

### How would this idea / enhancement be beneficial?

Make reviews easier and allow people to verify they entered the correct links/uploaded the correct pictures when submitting.

PR #910, opened 2024-12-22
Update show and tell image/video attachments to be clickable

shawnhugginsjr:
## Describe your changes

resolves issue https://github.com/alveusgg/alveusgg/issues/565.
- Clicking on a image attachment will display the image in a dialog
- Video attachments are now clickable hyperlinks instead of just text

I tested the changes locally.
Screenshot 2024-12-22 at 17 32 23

### DIFF
--- a/apps/website/src/components/shared/form/ImageUploadAttachment.tsx
+++ b/apps/website/src/components/shared/form/ImageUploadAttachment.tsx
@@ -1,4 +1,4 @@
-import type { ReactNode } from "react";
+import type { ReactNode, MouseEvent } from "react";
 
 import { classes } from "@/utils/classes";
 
@@ -66,18 +66,32 @@ export function ImageUploadAttachment({
   fileReference,
   children = "",
   showRemoveButton = true,
+  onClick,
 }: {
   removeFileReference: (id: string) => void;
   fileReference: FileReference;
   children?: ReactNode | ReactNode[];
   showRemoveButton?: boolean;
+  onClick?: (e: MouseEvent<HTMLAnchorElement>) => void;
 }) {
+  const src =
+    fileReference.status === "saved" || fileReference.status === "upload.done"
+      ? fileReference.url
+      : "dataURL" in fileReference
+        ? fileReference.dataURL
+        : undefined;
+
   return (
     <div className="flex flex-row gap-5 rounded-lg bg-white p-2 px-4 shadow-lg">
       <div className="py-2">
-        <div className="relative size-32 overflow-hidden rounded-lg bg-gray-200">
+        <a
+          href={src}
+          className="relative block size-32 overflow-hidden rounded-lg bg-gray-200"
+          title="Click to view image"
+          onClick={onClick}
+        >
           <ImageUploadFilePreview fileReference={fileReference} />
-        </div>
+        </a>
 
         {showRemoveButton && (
           <div className="p-2">
--- a/apps/website/src/components/show-and-tell/ShowAndTellEntryForm.tsx
+++ b/apps/website/src/components/show-and-tell/ShowAndTellEntryForm.tsx
@@ -1,10 +1,5 @@
-import type {
-  MouseEvent,
-  type FormEvent,
-  useCallback,
-  useMemo,
-  useState
-} from "react";
+import { useCallback, useMemo, useState } from "react";
+import type { FormEvent , MouseEvent } from "react";
 import { useRouter } from "next/router";
 import Image from "next/image";
 import type { ShowAndTellEntry } from "@prisma/client";
--- a/apps/website/src/components/show-and-tell/ShowAndTellEntryForm.tsx
+++ b/apps/website/src/components/show-and-tell/ShowAndTellEntryForm.tsx
@@ -1,5 +1,12 @@
-import { type FormEvent, useCallback, useMemo, useState } from "react";
+import type {
+  MouseEvent,
+  type FormEvent,
+  useCallback,
+  useMemo,
+  useState
+} from "react";
 import { useRouter } from "next/router";
+import Image from "next/image";
 import type { ShowAndTellEntry } from "@prisma/client";
 
 import type {
@@ -47,6 +54,7 @@ import {
 import Link from "../content/Link";
 import type { MapLocation } from "../shared/form/MapPickerField";
 import { MapPickerField } from "../shared/form/MapPickerField";
+import { ModalDialog } from "../shared/ModalDialog";
 import { DominantColorFieldset } from "./DominantColorFieldset";
 
 type ShowAndTellEntryFormProps = {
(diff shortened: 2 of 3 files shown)


pjeweb:
@shawnhugginsjr Thank you! This is great.

I would love for the image to be bigger though, the default classes on the modal dialog panel are a bit restrictive. From what i tried, you can probably override the `panelClassName` prop on the dialog and increase the image width/height.

And this might be a weird request, but could we use a elements for the images too so you can cmd-click/middle-mouse-click to open them in a new tab?

shawnhugginsjr:
@pjeweb 
> @shawnhugginsjr Thank you! This is great.

Thanks! 


> I would love for the image to be bigger though, the default classes on the modal dialog panel are a bit restrictive. From what i tried, you can probably override the `panelClassName` prop on the dialog and increase the image width/height.

I do agree that `max-w-md` is a bit small. Is there a concern about scaling a low-res image too high that it looks blurry? Always opening the images in a new tab seems simpler in hindsight.

> And this might be a weird request, but could we use a elements for the images too so you can cmd-click/middle-mouse-click to open them in a new tab?

Is this `cmd-click` functionality supposed to be for the image inside the dialog box or the image inside the `ImageUploadAttachment`?  Or perhaps both?

MattIPv4:
> Is there a concern about possibly scaling a low-res image too high that it looks blurry?

I probably wouldn't worry about that being an issue -- if it does become a problem, we can just ensure we're only setting the max width but allowing the image to use its intrinsic size if smaller.

> Is this `cmd-click` functionality supposed to be for the image inside the dialog box or the image inside the `ImageUploadAttachment`? Or perhaps both?

I think this is just referring to the native browser functionality you'd get with it being a link, where you can cmd/middle click to open the link in a new tab.

shawnhugginsjr:
>I think this is just referring to the native browser functionality you'd get with it being a link, where you can cmd/middle click to open the link in a new tab.

Yes. Wrapping an anchor tag around the image inside the modal dialog is fine. But the image inside the `ImageUploadAttachment` componenet is already wrapped by a button that opens the modal. So I don't think the anchor tag will work in this case.

MattIPv4:
Could we remove the button and just wrap it in an anchor tag? We can still bind to the on click for the anchor to do what the button would, but the anchor provide all the other native functionality we're looking for?

shawnhugginsjr:
>  We can still bind to the on click for the anchor to do what the button would, but the anchor provide all the other native functionality we're looking for?

When I tried this the `onClick` handler would always execute before doing the link action (opening link in new tab or redirecting user to link). So the modal will open and the link will be used at the same time.

Also, it looks like modern browsers don't support `target="_blank"` or `window.open` with data urls anymore. Chrome kept giving me this error. `Not allowed to navigate top frame to data URL: `. Using `Open Link in New Tab` is fine though. 

We can keep the button and anchor tag. And the user can use `Open Link in New Tab` if they want to see the source.

MattIPv4:
@pjeweb did you have any thoughts on this? If not, I'm happy with whatever works best to allow the modal to be opened while also allowing middle-click etc. to open in new tab.

apps/website/src/components/shared/form/ImageUploadAttachment.tsx
```diff
@@ -32,7 +34,11 @@ export function ImageUploadAttachment({
   return (
     <div className="flex flex-row gap-5 rounded-lg bg-white p-2 px-4 shadow-lg">
       <div className="py-2">
-        <div className="relative size-32 overflow-hidden rounded-lg bg-gray-200">
+        <button
+          className="relative size-32 cursor-pointer overflow-hidden rounded-lg bg-gray-200"
+          title="Click to view image"
+          onClick={onClick}
+        >
```
pjeweb, line 41:
```suggestion
        <a
          href={src}
          className="relative size-32 cursor-pointer overflow-hidden rounded-lg bg-gray-200"
          title="Click to view image"
          onClick={onClick}
        >
```

apps/website/src/components/shared/form/ImageUploadAttachment.tsx
```diff
@@ -13,10 +13,12 @@ export function ImageUploadAttachment({
   removeFileReference,
   fileReference,
   children = "",
+  onClick,
 }: {
   removeFileReference: (id: string) => void;
   fileReference: FileReference;
   children?: ReactNode | ReactNode[];
+  onClick?: () => void;
```
pjeweb, line 21:
```suggestion
  onClick?: (e: MouseEvent<HTMLAnchorElement>) => void;
```

apps/website/src/components/shared/form/ImageUploadAttachment.tsx
```diff
@@ -54,7 +60,7 @@ export function ImageUploadAttachment({
               <span>Failed</span>
             </div>
           )}
-        </div>
+        </button>
```
pjeweb, line 63:
```suggestion
        </a>
```

apps/website/src/components/show-and-tell/ShowAndTellEntryForm.tsx
```diff
@@ -352,10 +365,20 @@ export function ShowAndTellEntryForm({
...
                     : undefined;
 
+                const handleImageClick = () => {
+                  if (
+                    fileReference.status === "upload.done" ||
+                    fileReference.status === "saved"
+                  ) {
+                    openImageModal(fileReference.url);
+                  }
+                };
```
pjeweb, line 375:
```suggestion
                const handleImageClick = (e: MouseEvent<HTMLAnchorElement>) => {
                  if (e.button !== 0 || e.ctrlKey || e.shiftKey || e.altKey || e.metaKey) {
                    return;
                  }

                  if (
                    fileReference.status === "upload.done" ||
                    fileReference.status === "saved"
                  ) {
                    e.preventDefault();
                    openImageModal(fileReference.url);
                  }
                };
```

pjeweb (commented):
I think its fine if the link does not work while uploading. Instead of trying to explain my idea, I put it in code suggestions. This is not tested and is probably missing some imports etc.

In any case I'm happy to merge it as is.

apps/website/src/components/shared/form/ImageUploadAttachment.tsx
```diff
@@ -13,10 +13,12 @@ export function ImageUploadAttachment({
...
   children = "",
+  onClick,
 }: {
   removeFileReference: (id: string) => void;
   fileReference: FileReference;
   children?: ReactNode | ReactNode[];
+  onClick?: () => void;
 }) {
   const src =
     fileReference.status === "saved"
```
shawnhugginsjr, line 24: @pjeweb 
For your suggestion to work we will need to also update the image source like this. Otherwise we will get the `Not allowed to navigate top frame to data URL` error because we're using a data url in the anchor link.
```suggestion
    fileReference.status === "saved" || fileReference.status === "upload.done"
```

pjeweb: That seems fine, yeah. Sorry for not responding earlier. I really only care about the review, but yeah ideally it should also work immediately after the upload. I think I did not use the uploaded url so its not having to fetch the image again (that was just uploaded), but it does make sense!

MattIPv4:
👋 Hey @shawnhugginsjr just wanted to check in on how things are going here? Do you need anything else from us to keep moving?

shawnhugginsjr:
@MattIPv4 
My bad. I waiting for pjeweb's reply but missed it. Let me look back into this.

apps/website/src/components/shared/form/ImageUploadAttachment.tsx
shawnhugginsjr: @pjeweb 
I updated the code to support command clicking.

MattIPv4:
Hey, why did this get closed?

PR #912, opened 2024-12-25
Use de-duplicated species data methods

flakey5:
Re alveusgg/data#87

Draft until alveusgg/data#100 is merged.

TODO
* [x] Update `@alveusgg/data` version
* [x] Revert changes to `@alveusgg/data` imports

### DIFF
--- a/apps/website/src/pages/animal-quest/index.tsx
+++ b/apps/website/src/pages/animal-quest/index.tsx
@@ -14,8 +14,8 @@ import {
   getClassification,
   sortAmbassadorClassification,
 } from "@alveusgg/data/src/ambassadors/classification";
-
 import { getSpecies } from "@alveusgg/data/src/ambassadors/species";
+
 import useGrouped, { type GroupedItems, type Options } from "@/hooks/grouped";
 
 import { formatDateTime } from "@/utils/datetime";
--- a/apps/website/src/pages/animal-quest/index.tsx
+++ b/apps/website/src/pages/animal-quest/index.tsx
@@ -15,6 +15,7 @@ import {
   sortAmbassadorClassification,
 } from "@alveusgg/data/src/ambassadors/classification";
 
+import { getSpecies } from "@alveusgg/data/src/ambassadors/species";
 import useGrouped, { type GroupedItems, type Options } from "@/hooks/grouped";
 
 import { formatDateTime } from "@/utils/datetime";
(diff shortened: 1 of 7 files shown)


MattIPv4:
👀 we're going to need to think about how this impacts the API endpoint for the ambassadors and how we can roll this out without breaking the existing live version of the extension that's reliant on the current schema for the endpoint (while also thinking about how we can release a new version of the extension that can then start using the new data format if it is different)

MattIPv4:
Late-night thinking on that... could definitely be solved on either end.

My initial gut take was that it might be easier to solve in the extension. That is, update the extension first with the new data/types, while having fallback logic in it to parse against the old types/schema as well. Once that's released to Twitch, the site can update its data/types and the endpoint to match.

Chewing on that more though, that puts the ball in Twitch's court so to speak and blocks the site doing any data updates until that new version is released -- it would therefore be much better to have backwards compatibility on the website instead so that it can keep iterating rapidly and the extension can consume the new types/schema whenever it is ready (and once released the endpoint backwards compatibility can be removed).

flakey5:
> it would therefore be much better to have backwards compatibility on the website instead

+1

But it is noteworthy that this doesn't *need* to change the api response (imo it should just for consistency, but still just something to mention)

How would you feel about versioning the api routes? I.e. making a `src/pages/api/v2/stream/ambassadors.ts` that returns the updated format, that way the current endpoint can continue returning the same response and can easily be removed when it's no longer needed

MattIPv4:
Yah, I imagine versioning is the way we want to go with it. And indeed we could not change the endpoint at all, though I agree I think we should try to match the new data structure (I imagine having a `species` prop on each ambassador with the full species data).

My initial thought was versioning via an accept header, before I once again remembered Vercel doesn't support the vary header.

Route versioning would definitely work and is pretty standard, though my only hesitation is that it implies the whole API is version while actually it is just this endpoint (and makes the repo structure that bit more complex).

Slightly more non-standard, but thoughts on just returning a `v2` prop in the existing response? That'd still pass the existing schema parsing which just looks for the `ambassadors` prop, and the new schema can just look for the `v2` prop instead?

MattIPv4:
Looks like this needs a rebase, and maybe some prettier fixes, and then it's good to go? I've tested this locally against the current extension version, but will test the PR preview here once deployed as well to make sure we're not going to break anything.

apps/website/src/pages/animal-quest/index.tsx
```diff
@@ -15,6 +15,7 @@ import {
   sortAmbassadorClassification,
 } from "@alveusgg/data/src/ambassadors/classification";
 
+import { getSpecies } from "@alveusgg/data/src/ambassadors/species";
```
MattIPv4, line 18:
```suggestion
import { getSpecies } from "@alveusgg/data/src/ambassadors/species";

```

apps/website/src/pages/api/stream/ambassadors.ts
```diff
@@ -10,36 +10,53 @@ import {
   type AmbassadorImage,
 } from "@alveusgg/data/src/ambassadors/images";
 import { getClassification } from "@alveusgg/data/src/ambassadors/classification";
+import type { Species } from "@alveusgg/data/src/ambassadors/species";
+import { getSpecies } from "@alveusgg/data/src/ambassadors/species";
```
MattIPv4, line 14:
```suggestion
import { getSpecies, type Species } from "@alveusgg/data/src/ambassadors/species";
```

apps/website/src/pages/api/stream/ambassadors.ts
```diff
@@ -10,36 +10,53 @@ import {
...
 import { getIUCNStatus } from "@alveusgg/data/src/iucn";
-
-import { env } from "@/env";
-
+import { createImageUrl } from "@/utils/image";
```
MattIPv4, line 16:
```suggestion

import { createImageUrl } from "@/utils/image";
```

MattIPv4 (commented):
A few nits

MattIPv4:
:( looks like one of my suggestions angered prettier, sorry

MattIPv4 (approved):
TYVM

PR #920, opened 2024-12-30
Do not error when sending pushes fails

pjeweb:
## Describe your changes

Instead of returning a 500 error the send API endpoint will now respond with 200 even if the push was not delivered, so the queue does not retry the send action in addition to the internal retry system. We suspect that might have led to too many push requests to the vendors (apple, google, microsoft etc.).

Test notification pushes still work.


pjeweb:
Currently deployed to Preview.

MattIPv4 (approved):

PR #932, opened 2025-01-06
Improve calendar text contrast for Opera GX's dark mode

### LINKED ISSUE #766: [Idea]: Make schedule look better with Opera GX "Force dark page" mode

I suggest checking, if we can make the schedule look better on with Opera GX "Force dark page" mode. Maybe its worth checking other browser extensions too.

Biggest problem to me is the white on light color contrast.

Screenshot 2024-09-06 at 16 52 51

Improve readability/accessibility when using Opera GX's "Force dark page" mode.

brianpeiris:
## Describe your changes

Fixes #766 

### TLDR: 
This was trickier than you might think. Opera GX's dark mode is just weird, and there's not much we can do about it. In this PR, I've lightened the background colors on calendar events so that Opera GX's dark mode has more contrast. I've also added a gray border around the calendar legend swatches.

#### Comparison:

Before (light mode):
light-before

After (light mode):
light-after

Before (Opera GX dark mode):
dark-before

After (Opera GX dark mode):
dark-after

---

### Longer explanation:

Opera GX's dark mode appears to be a post-rendering step that cannot be controlled or detected in CSS/JS. It seems that they arbitrarily picked a lightness threshold for a small set of background colors which are turned to darker shades. For text, they lighten all dark text, with a proportional band of grey in the middle of the spectrum.

Picking a darker text color will result a lighter text in dark mode. If we wanted to pick more suitable background colors, you'd have to go even lighter with the backgrounds.

Here's a series of swatches showing text shades (most of the first row) and background colors, and their comparison in Opera GX's dark mode.
opera-gx-dark-mode.webm

Here's a comparison with the calendar's tailwind colors and examples of lighter backgrounds:
opera-gx-dark-mode-tw.webm

---
...

This is just a visual change to the calendar, so I guess just makes sure it looks okay  on the schedule page and the calendar admin.

...


MattIPv4:
👀 Looking at your comparison of the change (not sure which is before/after), both seem very low contrast still?

brianpeiris:
The text is brighter after, still pretty low contrast, but I think it's a good improvement.

We could try a darker shade of grey, but there isn't a way to keep the text dark, as far as I can tell. The problem is that we'd have to change it in a way that doesn't negatively affect the regular/light mode. What do you think about the "lighter backgrounds" tailwind example?

Opera has a bug/suggestion form for GX here: https://opera.atlassian.net/servicedesk/customer/portal/18
They also have a forum for GX: https://forums.opera.com/category/50/opera-gx

Before:
Screenshot from 2025-01-05 23-03-53

After:
Screenshot from 2025-01-05 23-04-03

MattIPv4:
I think the lighter background colors are probably fine instead of changing the text color -- @pjeweb thoughts?

pjeweb:
> I think the lighter background colors are probably fine instead of changing the text color -- @pjeweb thoughts?

yeah i agree

MattIPv4:
👋 Hey @brianpeiris, checking in on this?

brianpeiris:
Sorry for the delay. I'll get back to this after work today.

brianpeiris:
@MattIPv4 I've changed the code to use lighter background colors on the calendar events. I've also updated the PR description with comparison images.

MattIPv4 (approved):

pjeweb:
@brianpeiris Thank you!

brianpeiris:
Thanks for the review and merge!

Issue #928, opened 2025-01-01
[Bug]: Upstash/QStash calls (especially notification pushes) need to be throttled (max 100/s)


pjeweb:
### What is the current behavior?

Currently we just send requests to qstash without throttling. Since we do not retry those (thats the whole point of delegation to that service), we run into Upstashes 100 requests per second burst rate limiting for the Pay as you Go plan.

### What is the expected behavior?

Throttle request (either globally - how?) or at least in the notification push logic to less than 100 per second. Maybe also consider setting an explicit function duration limit (https://vercel.com/docs/functions/configuring-functions/duration).

### Steps to reproduce

n/a

### Browser specs

n/a

pjeweb:
This has been mostly "fixed" by #947, but that PR only handles push notifications, not all QStash calls in general, so I'll keep it open for now.

PR #946, opened 2025-01-11
Add naive delay to Upstash QStash calls to avoid ratelimit

MattIPv4:
## Describe your changes

A very naive workaround to alleviate (but not properly fix) #928.

Notifications still work.

### DIFF
--- a/apps/website/src/server/utils/queue.ts
+++ b/apps/website/src/server/utils/queue.ts
@@ -1,6 +1,6 @@
 import { env } from "@/env";
 
-export function callEndpoint<T>(endpoint: string, body: T) {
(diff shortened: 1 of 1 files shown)


pjeweb (approved):
I'm pretty sure this will not help cause there will be many concurrent calls being awaited at the same time in `apps/website/src/pages/api/notifications/batched-create-notification-pushes.ts`. 

I started to refactor the whole thing but have not had enough brain time to think it through yet.

We can try this, but I do not think it will help much.

pjeweb:
This is similarly naive, but I think this *should* do the trick: #947

PR #947, opened 2025-01-12
Throttle notification pushes to prevent QStash rate limiting

pjeweb:
## Describe your changes

To prevent rate limiting from QStash this will throttle notification pushes by delaying batches by 2 seconds.

This also removes the unnecessary `data/env/push.ts` file by setting env defaults in `env/index.js` directly.

This also explicitly sets a (longer) duration for the functions.

Test notifications still work


MattIPv4 (commented):
Does the `callEndpoint` in `batched-create-notification-pushes` also need the delay -- as I understand it, that will write to QStash again after the call you've added the delay to here?

pjeweb:
> Does the `callEndpoint` in `batched-create-notification-pushes` also need the delay -- as I understand it, that will write to QStash again after the call you've added the delay to here?

Assuming batches of 50 with 1sec of delay between we'd have one call to `batched-create-notification-pushes` followed by 50 calls to `send-push` per second, so a total of 51/s, which leaves some headroom.

MattIPv4 (approved):

PR #948, opened 2025-01-12
Random Twitch clips stream overlay

MattIPv4:
## Describe your changes

Fetches clips from the Alveus Twitch channel from the last year that have over 100 views, excluding the last week (to allow time for mods to catch any bad clips). Refreshes the list of clips via Next.js revalidation every 30 minutes (similar to the homepage refresh). Uses the Twitch clips player directly to avoid needing funky workarounds to download the clip videos themselves, though this does mean the logic to rotate the clips is a naive timer that doesn't factor in clip buffering/loading.

Clips load/play as expected.

### DIFF
--- a/apps/website/src/pages/stream/clips.tsx
+++ b/apps/website/src/pages/stream/clips.tsx
@@ -200,34 +200,36 @@ const ClipsPage: NextPage<InferGetStaticPropsType<typeof getStaticProps>> = ({
   );
 
   return (
-    <div className="flex h-screen w-full p-8">
+    <div className="flex h-screen w-full items-center justify-center p-8">
       {clip && (
-        <div className="relative my-auto aspect-video h-auto w-full">
-          <div className="absolute -inset-2 -z-10 rounded-xl bg-alveus-green shadow-lg" />
+        <div className="flex aspect-video h-full max-w-full items-center justify-center">
+          <div className="relative flex aspect-video w-full items-center justify-center">
+            <div className="absolute -inset-2 -z-10 rounded-xl bg-alveus-green shadow-lg" />
 
-          <Transition show={details}>
-            <div className="absolute left-2 top-2 rounded-lg bg-black/25 px-4 py-2 text-white backdrop-blur transition-opacity data-[closed]:opacity-0 data-[enter]:duration-700 data-[leave]:duration-300">
-              <h1 className="text-4xl">
-                {clip.title}
-                <span className="ml-1 text-2xl">
-                  {" "}
-                  (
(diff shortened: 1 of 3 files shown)


apps/website/src/pages/stream/clips.tsx
```diff
@@ -0,0 +1,233 @@
...
+    if (!clip) return;
+
+    if (fallbackTimer.current) clearTimeout(fallbackTimer.current);
+    fallbackTimer.current = setTimeout(
+      () => increment(),
+      clip.duration * 1000 * 1.5,
+    );
+  }, [clip, increment]);
+
+  // As soon as the clip loads, start a timer to randomize when the clip ends
```
MattIPv4, line 156:
```suggestion
  // As soon as the clip loads, start a timer for when the clip ends
```

apps/website/src/pages/stream/clips.tsx
```diff
@@ -0,0 +1,233 @@
...
+    loadedTimer.current = setTimeout(
+      () => increment(),
+      clip.duration * 1000 + 2 * 1000, // Fudge factor of 2 seconds for clip loading
+    );
+
+    if (detailsTimer.current) clearTimeout(detailsTimer.current);
+    detailsTimer.current = setTimeout(() => setDetails(false), 10 * 1000);
+  }, [clip, increment]);
+
+  // If the clip fails to load, randomize after 2 seconds
```
MattIPv4, line 176:
```suggestion
  // If the clip fails to load, show another after 2 seconds
```

MattIPv4:
Hm @pjeweb I'm realising the cron doesn't run in preview, and so the Twitch tokens expire -- with this PR, that's going to block every deploy going forward. Should I just try/catch this page and have it revalidate frequently if it failed (similar to what is already does if there is no token in the DB), or do we want to try getting the cron to work (I would think we'd need to have the prod cron have a task that calls the preview cron)?

pjeweb:
> Hm @pjeweb I'm realising the cron doesn't run in preview, and so the Twitch tokens expire -- with this PR, that's going to block every deploy going forward. Should I just try/catch this page and have it revalidate frequently if it failed (similar to what is already does if there is no token in the DB), or do we want to try getting the cron to work (I would think we'd need to have the prod cron have a task that calls the preview cron)?

I mean yeah it's kinda ugly, but I think doing both would be the best. I set up a schedule with QStash/Upstash too. So it should work now …

MattIPv4:
@pjeweb this should be good for a review now, tyty

pjeweb (approved):
very cool

PR #949, opened 2025-01-13
Add endpoint to get latest YouTube videos

MattIPv4:
## Describe your changes

Allows us to replace the use of `https://decapi.me/youtube/latest_video?id=UCbJ-1yM55NHrR1GS9hhPuvg&no_shorts&no_livestream&exclude=livecams&exclude=live` in our chat bot, which excludes live streams based on `livecams` or `live` in the title, with this endpoint which only fetches long-form videos for the channels.

Endpoint returns videos, which are not the live stream.


pjeweb (approved):
clean

PR #957, opened 2025-01-15
Disable admin auth for easier local development

brianpeiris:
## Describe your changes

The local development instructions ask you to create a Twitch application in order to use the admin dashboard. This PR makes it easier to get started by disabling authentication for local development, so that you can use the admin dashboard immediately, without having to create a Twitch application. It adds a `DISABLE_ADMIN_AUTH` environment variable that is set to `false` by default in `.env.example`. When this variable is set to true, and the `NODE_ENV` environment variable is set to "development", the admin dashboard will not require authentication.

I've also added corresponding instructions in the README.

1. Comment out the `TWITCH_CLIENT_ID` and `TWITCH_CLIENT_SECRET` environment variables from your local `.env` file.
2. Set `DISABLE_ADMIN_AUTH` to `true`
3. See that you still access the admin dashboard in local dev.

### DIFF
--- a/README.md
+++ b/README.md
@@ -58,12 +58,12 @@ Hey there! Welcome to Alveus.gg! There's a few ways that you can help contribute
 ## Development setup
 
 > [!NOTE]
-> If you only want to work on the front end, you may skip the prerequisite and skip setting up a database or file storage (steps 3 and 4.iv).
+> If you only want to work on the front end, you may skip the prerequisite and skip setting up a database or file storage (step 3).
 > But you may encounter some errors when running the website without a database or file storage.
 
 ### Prerequisite
 
-Create a [Twitch application](https://dev.twitch.tv/console/apps/create), setting the OAuth callback to be `http://localhost:3000/api/auth/callback/twitch`. Note down your client ID and client secret.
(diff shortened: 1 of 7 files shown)


MattIPv4:
@pjeweb how do you feel about this? While a bit more hassle, my gut says the local setup should always be as close to prod as possible so that we're exercising all the code and flows?

pjeweb:
> @pjeweb how do you feel about this? While a bit more hassle, my gut says the local setup should always be as close to prod as possible so that we're exercising all the code and flows?

I'm on the fence. I think it might be worth, but I'd default to `false`?!

MattIPv4:
👍 Off by default seems fine to me

brianpeiris:
Thanks. I've set the flag to `false` by default in `.env.example`, and I've updated the instructions in the README to mention the flag.

MattIPv4 (commented):
👀 It looks like your stubbed session doesn't match the expended type signature of a session. Could I suggest moving the `{ user: {} }` into a util, and then using the actual Session type to make sure your stubbed object matches the expected signature?

brianpeiris:
Whoops! Yup, just did that with a constant session structure in utils.

MattIPv4 (approved):
Thanks.

@pjeweb would appreciate a +1 from you on this as well

pjeweb (approved):

PR #962, opened 2025-01-17
Allow mods to set public + private notes on Show and Tell entries

### LINKED ISSUE #716: [Idea]: Mod-only comments for Show & Tell posts

Allow mods to add mod-only comments to posts (e.g. "held back review cause x")

Help collaborating on reviews

### LINKED ISSUE #717: [Idea]: Public mod comments on Show & Tell posts

Allow mods to add public comments to show and tell posts ("[removed x]"), not counting towards the char limit.

Allow transparent edits where changes are made public.

MattIPv4:
## Describe your changes

Resolves #716 and resolves #717 by allowing mods to set both public + private notes on the Show and Tell entry review screen. Private notes can only be seen by moderators on the review screen, while public notes will be shown on the show and tell page itself.

Moderators can set public + private notes during review. They persist when saved.

If a user edits their post after public + private notes have been added, they still persist when the review screen is accessed.

As a regular user, accessing "my posts" does not reveal the private notes (even in the underlying network data). Similarly, accessing an individual post does not reveal private notes.

As a regular user, or signed out, accessing the show and tell page does not reveal private notes (even in the underlying network data). Any public notes on approved posts are shown on the page.

### DIFF
--- a/apps/website/prisma/schema.prisma
+++ b/apps/website/prisma/schema.prisma
@@ -404,6 +404,8 @@ model ShowAndTellEntry {
   location            String?
   longitude           Float?
   latitude            Float?
+  notePrivate         String?   @db.Text
+  notePublic          String?   @db.Text
 
   attachments ShowAndTellEntryAttachment[]
   user        User?                        @relation(fields: [userId], references: [id])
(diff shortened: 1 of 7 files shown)


MattIPv4:
Build failed due to the DB schema not being updated -- @pjeweb would you mind pushing the new schema to preview?

pjeweb:
> Build failed due to the DB schema not being updated -- @pjeweb would you mind pushing the new schema to preview?

DB change is pushed and re-deployment started.

pjeweb (changes requested):
The mod notes wont show unless there is text content. Otherwise this is great

MattIPv4:
Will hold off on merging this until after the Show and Tell stream today, so we don't risk breaking smth right before that.

pjeweb (approved):

PR #964, opened 2025-01-17
Use waitUntil() for notification APIs

pjeweb:
## Describe your changes

Makes use of the `waitUntil` util provided by vercel to keep the "Function" running after responding to the client. This will await the actual creation of the notification, but respond early, and letting outgoing requests run after.

Send test notification. Observe a quick response, but all notifications to be sent.

(This might be hard with the small amount of test subscribers)

### DIFF
--- a/apps/website/src/server/notifications.ts
+++ b/apps/website/src/server/notifications.ts
@@ -1,4 +1,5 @@
 import type { Notification } from "@prisma/client";
+import { waitUntil } from "@vercel/functions";
 
 import { env } from "@/env";
 
@@ -80,7 +81,7 @@ export async function createNotification(data: CreateNotificationData) {
     tasks.push(createDiscordNotifications(notification));
   }
 
-  await Promise.all(tasks);
+  await Promise.allSettled(tasks);
 }
 
 export async function resendNotification(notificationId: string) {
@@ -194,7 +195,7 @@ async function createPushNotifications(notification: Notification) {
     );
   }
 
-  return Promise.allSettled(requests);
+  waitUntil(Promise.allSettled(requests));
 }
 
 async function createDiscordNotifications({
@@ -247,5 +248,5 @@ async function createDiscordNotifications({
     );
   }
 
-  return Promise.all(tasks);
+  waitUntil(Promise.allSettled(tasks));
 }
--- a/apps/website/src/server/notifications.ts
+++ b/apps/website/src/server/notifications.ts
@@ -1,5 +1,4 @@
 import type { Notification } from "@prisma/client";
-import { waitUntil } from "@vercel/functions";
 
 import { env } from "@/env";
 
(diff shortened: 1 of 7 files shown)


apps/website/src/server/notifications.ts
```diff
@@ -80,7 +81,7 @@ export async function createNotification(data: CreateNotificationData) {
     tasks.push(createDiscordNotifications(notification));
   }
 
-  await Promise.all(tasks);
+  await Promise.allSettled(tasks);
```
pjeweb, line 84: Snuck this one in. We do not wait to abort if one of the notifications fails, so I think its better to use `allSettled` here.

apps/website/src/server/notifications.ts
```diff
@@ -247,5 +248,5 @@ async function createDiscordNotifications({
     );
   }
 
-  return Promise.all(tasks);
+  waitUntil(Promise.allSettled(tasks));
```
pjeweb, line 251: dito

apps/website/src/server/notifications.ts
```diff
@@ -194,7 +195,7 @@ async function createPushNotifications(notification: Notification) {
     );
   }
 
-  return Promise.allSettled(requests);
+  waitUntil(Promise.allSettled(requests));
```
MattIPv4, line 198: My gut says we should only call `waitUntil` in the API handlers themselves, and these util methods should be agnostic?

pjeweb: Yeah that was my first attempt too, but we'd need to split up the functions into two or more calls in the handlers. I think there were 3 points where `createNotification` is called.

So on the call site:

```typescript
const notification = await createNotification(…);
waitUntil(sendNotification(notification));
```

Not sure if its worth? I was thinking, since it is in the context of the server it makes "enough" sense to call the `waitUntil`  in the utils?

MattIPv4: Ah okay, that makes sense -- I suppose my follow-up would be then why there are some methods in here that waitUntil and some that don't? It just feels like a bit of a landmine down the road?

pjeweb: Yeah true. Not sure if there is a good way to signal that in the function signature? Like `sendNotificationButDoNotAwaitButAlsoDoNotStopFunction()`.

It would be easier if it was actually a thread or something. But the mix of `await` and `waitUntil` is kinda bad wording in that sense.

pjeweb: `WithoutBlockingClientResponse`?

MattIPv4: Maybe just stick `WaitUntil` on the end of the method names?

pjeweb: I guess. But it does exactly the opposite of that unless you know its referring to that exact util ;)

I am happy to move the waitUntil in the handlers. I agree thats easier to understand

MattIPv4: Ah true -- okay, sounds like a plan, ty!

MattIPv4 (approved):

PR #967, opened 2025-01-20
Switch weather + youtube stream API endpoints to text

MattIPv4:
## Describe your changes

Switches the weather + youtube API endpoints for the stream chatbot from JSON responses to text responses, removing the need to parse + process the response within the stream chatbot.

As the weather endpoint now responds with text a new trpc endpoint was also needed for weather to allow the stream overlay to continue functioning.

Weather + youtube endpoints return the expected messages matching the current stream chatbot. Weather component in stream overlay continues to render data as expected.


apps/website/src/server/apis/weather.ts
```diff
@@ -84,3 +87,151 @@ export async function getCurrentObservation<T extends boolean>(
...
+  windChill: number | null,
+) => {
+  if (temperature !== null) {
+    if (temperature >= 70 && heatIndex !== null) return heatIndex;
+    if (temperature <= 61 && windChill !== null) return windChill;
+  }
+  return temperature;
+};
+
+export type Weather = {
```
pjeweb, line 103: do we really need the explicit type or can we just let it infer the data?

pjeweb (approved):

pjeweb (approved):

PR #971, opened 2025-01-21
Re-auth with additional scopes if account requires them

### LINKED ISSUE #970: [Bug]: Additional scopes lost for accounts linked to channels during auth

Any Twitch account logged into the site (with admin access) can authenticate with additional Twitch OAuth scopes and then link the account to a Twitch channel (which we use for schedule syncing etc.).

However, if the account then subsequently signs back into the site again, the scopes + token stored in the DB for the account will be updated to only be the regular auth scopes again rather than the full additional OAuth scopes we need for schedule syncing etc.

I imagine the easiest fix here is that once an account has done the initial auth flow on the site, we lookup if the account is tied to any channels, and if so immediately bounce the user back through OAuth again with the additional scopes.

### Steps to reproduce

- Sign into the site with Twitch
- In admin panel, auth with additional scopes
- In admin panel, link account to channel
- In DB, observe full scopes tied to account
- Sign out of the site
- Sign into the site with Twitch
- In DB, observe limited scopes tied to account

### Browser specs

N/A

### Anything else?

_No response_

MattIPv4:
## Describe your changes

When loading the session for the current user, if they have a linked channel check if we need additional scopes in their current session. If they do, the client will redirect them to sign in again with the additional scopes requested.

Fixes #970 

- A brand new user can successfully sign in. Sent to OAuth once.
- An existing regular user can successfully sign in. Sent to OAuth once.
- An admin user with no channels can successfully sign in. Sent to OAuth once.
- An admin user linked to a Twitch channel can successfully sign in. Sent to OAuth twice. Accessing `/admin/twitch` shows all scopes granted.

### DIFF
--- a/apps/website/src/pages/api/auth/[...nextauth].ts
+++ b/apps/website/src/pages/api/auth/[...nextauth].ts
@@ -11,7 +11,7 @@ import {
   ExpiredAccessTokenError,
   refreshAccessToken,
 } from "@/server/utils/oauth2";
-import { defaultScopes } from "@/data/twitch";
+import { botScopes, defaultScopes } from "@/data/twitch";
 import invariant from "@/utils/invariant";
 
 const adapter = PrismaAdapter(prisma);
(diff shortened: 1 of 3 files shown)


apps/website/src/pages/api/auth/[...nextauth].ts
```diff
@@ -41,9 +42,17 @@ export const authOptions: NextAuthOptions = {
           access_token: true,
           refresh_token: true,
           verified_at: true,
+          scope: true,
         },
```
Comment (pjeweb) [inline apps/website/src/pages/api/auth/[...nextauth].ts:46]: You could get the linked channels here to save the second query.

```suggestion
          scope: true,
          twitchChannelBroadcaster: true,
          twitchChannelModerator: true,
        },
```

pjeweb (approved):

pjeweb (approved):

PR #983, opened 2025-01-24
Allow selection in Show and Tell presentation mode + clarify private notes

MattIPv4:
## Describe your changes

Allows for text to be selected in the presentation mode for Show and Tell, as this tripped up Maya today when she was trying to select links in posts.

Also, clarifies the visibility of private mod notes as this tripped up the mod team today when a private note was added that should've been public.

Text can be selected, has no adverse effect.

Note clarification makes sense.


MattIPv4:
I thought about using linkify-it to make links properly clickable when users haven't, but I figured that might be problematic as we then wouldn't be able to make something _not_ a link that looks like a link -- so I think best to just leave it as plain-text unless the user explicitly makes it a link in the editor (though if we wanted to be really smart we could apply linkify-it in the editor and auto-link a "word" if it looks like a link after it is typed/pasted).

pjeweb (approved):

PR #1004, opened 2025-01-29
Fix minimum badge width

matadaniel:
The flex container shrinks badges too much.

2025-01-29T16:25:17,603977463-06:00


apps/website/src/components/show-and-tell/Badge.tsx
```diff
@@ -32,7 +32,7 @@ export const Badge = ({
   pulse?: boolean;
 }) => {
   return (
-    <div className="flex aspect-square w-[80px] items-center justify-center rotate-12">
+    <div className="flex aspect-square w-[80px] min-w-[80px] items-center justify-center rotate-12">
```
MattIPv4, line 35:
```suggestion
    <div className="flex size-20 shrink-0 items-center justify-center rotate-12">
```
(This may need to be done locally for prettier)
Instead of 80px as an arbitrary value, 20 is 5rem which should be 80px + instead of a width and square aspect we can just set width/height + instead of setting a min width we can tell any flex parent to never shrink this?

matadaniel: Haha literally all of this went through my head as I made the change. I wanted to keep the diff as small as possible. I went with min-w instead of shrink-0 since it should probably never be any smaller.

MattIPv4: Ah that's fair, and I suppose the aspect ensures that if the text is too long for 80px we're still going to be square.

Could we swap to w-20 min-w-20 then to avoid the arbitrary units (I'm surprised eslint didn't autofix that)?

MattIPv4 (approved):

PR #1059, opened 2025-03-01
Update homepage hero slideshow images

MattIPv4:
## Describe your changes

The photos used in the homepage slideshow were all from when Alveus originally started and have never been updated since, so it's about time we update them with some newer and higher-quality images.

Also, received a candid bit of feedback that with the first four images of the slideshow being the parrots, one might perceive the sanctuary as only a bird sanctuary initially, so as part of this update, I've included more variety in the ambassadors featured.

All images look good on desktop + mobile, alt captions are correct.


apps/website/src/utils/image.ts
MattIPv4, line 1: I hope this diff is safe -- I couldn't figure out why we'd want to throw an error if a width was too large?

pjeweb (approved):
very beautiful

PR #1062, opened 2025-03-01
Automatically add ambassador birthdays to calendar

MattIPv4:
## Describe your changes

When we populate the coming month with regular events, we can also use this as an opportunity to inject any upcoming ambassador birthdays!

Best way to test this at present will be to comment out the `.plus({ months: 1 })` to generate events for March locally, as this will include Winnie's birthday.


pjeweb (approved):
Love the idea. I would maybe still ask staff, if they maybe dont want to have the dates in the calendar, in case they move events around and such

MattIPv4:
I think it should be fine as these are just the birthdays linking to the site, not linking to a Twitch stream or w/e. They get injected quite a few days in advance of the new month, so should be plenty of time for them to edit them into an actual stream if that is happening.

PR #1076, opened 2025-03-11
Change YouTube embed URL to not show unrelated videos

pjeweb:
## Describe your changes

This should hopefully prevent YouTube from showing super random videos as recommendations. It will at least limit it to the channel that the video is from.

Check youtube embeds work on the page (e.g. homepage, collabs and also and especially in show and tell).


MattIPv4 (approved):
I'm not sure how perfect this'll be given it'll show videos from the Show and Tell OP's channel, but better than global recommended I imagine

PR #1088, opened 2025-03-14
Improve image optimization caching

pjeweb:
## Describe your changes

Improve caching by increasing cache TTL and limiting the cached quality variations

Observe more Vercel Cache hits in HTTP headers


apps/website/next.config.ts
```diff
@@ -30,6 +30,13 @@ const config: NextConfig = {
     dirs: ["."],
   },
   images: {
+    // limits the quality levels to limit cache variations for image optimization:
+    qualities: [50, 75, 100],
```
MattIPv4, line 34: Can we add 90 here as we use it in a few places? (100 is pass-through and doesn't get optimized at all)

MattIPv4 (approved):

PR #1113, opened 2025-03-24
feat: Add click handlers to show and tell images and videos for preview

### LINKED ISSUE #565: [Idea]: Show and tell admin: Open videos and pictures on click

DriesOlbrechts:
## Describe your changes

Adds a modal to preview the show and tell uploaded images on click.
Opens show and tell videos in a new tab when clicking the link.

resolves #565

...

I'm still considering adding preview with embeds for videos as well, but that will take some more effort!

...

### DIFF
--- a/apps/website/src/components/show-and-tell/ShowAndTellEntryForm.tsx
+++ b/apps/website/src/components/show-and-tell/ShowAndTellEntryForm.tsx
@@ -13,6 +13,7 @@ import {
   DisclosurePanel,
 } from "@headlessui/react";
 
+import Image from "next/image";
 import type {
   PublicShowAndTellEntryWithAttachments,
   ShowAndTellSubmitInput,
@@ -59,6 +60,7 @@ import {
 import Link from "../content/Link";
 import type { MapLocation } from "../shared/form/MapPickerField";
 import { MapPickerField } from "../shared/form/MapPickerField";
+import { ModalDialog } from "../shared/ModalDialog";
 import { DominantColorFieldset } from "./DominantColorFieldset";
 
 type ShowAndTellEntryFormProps = {
@@ -180,6 +182,8 @@ export function ShowAndTellEntryForm({
 }: ShowAndTellEntryFormProps) {
   const [isSubmitted, setIsSubmitted] = useState(false);
   const [successMessage, setSuccessMessage] = useState<string | null>(null);
+  const [isPreviewOpen, setIsPreviewOpen] = useState(false);
+  const [previewImageUrl, setPreviewImageUrl] = useState<string | null>(null);
   const [error, setError] = useState<string | null>(null);
   const router = useRouter();
   const create = trpc.showAndTell.create.useMutation();
@@ -190,6 +194,15 @@ export function ShowAndTellEntryForm({
   const [wantsToTrackGiveAnHour, setWantsToTrackGiveAnHour] = useState(
     !!entry?.volunteeringMinutes,
   );
+
+  function closeModal() {
+    setIsPreviewOpen(false);
+  }
+
+  function openModal(url: string) {
+    setPreviewImageUrl(url);
+    setIsPreviewOpen(true);
+  }
 
   const initialLocation = useMemo<MapLocation | undefined>(
     () =>
@@ -268,6 +281,7 @@ export function ShowAndTellEntryForm({
 
       dominantColor = [r, g, b].join();
     }
+
     const data: ShowAndTellSubmitInput = {
       displayName: formData.get("displayName") as string,
       title: formData.get("title") as string,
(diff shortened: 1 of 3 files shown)


apps/website/src/components/show-and-tell/ShowAndTellEntryForm.tsx
```diff
@@ -191,6 +195,15 @@ export function ShowAndTellEntryForm({
...
   );
 
+  function closeModal() {
+    setIsPreviewOpen(false);
+  }
+
+  function openModal(url: string) {
+    setPreviewImageUrl(url);
+    setIsPreviewOpen(true);
+  }
```
MattIPv4, line 205: Can these be arrow methods?

apps/website/src/components/show-and-tell/ShowAndTellEntryForm.tsx
```diff
@@ -71,9 +73,12 @@ type ShowAndTellEntryFormProps = {
 function ImageAttachment({
   entry,
   fileReference,
+  callback,
```
MattIPv4, line 76: Should this be called `onClick`, and then exclude `onClick` from `ComponentProps` below?

apps/website/src/components/show-and-tell/ShowAndTellEntryForm.tsx
```diff
@@ -84,8 +89,21 @@ function ImageAttachment({
...
+      fileReference.status === "saved"
+    ) {
+      callback(fileReference.url);
+    }
+  };
+
   return (
-    <ImageUploadAttachment {...props} fileReference={fileReference}>
+    <ImageUploadAttachment
+      onClick={handlePreviewClick}
```
MattIPv4, line 103: This should probably be after the spread, so that the spread can't override it?

apps/website/src/components/show-and-tell/ShowAndTellEntryForm.tsx
```diff
@@ -180,6 +198,8 @@ export function ShowAndTellEntryForm({
 }: ShowAndTellEntryFormProps) {
   const [isSubmitted, setIsSubmitted] = useState(false);
   const [successMessage, setSuccessMessage] = useState<string | null>(null);
+  const [isPreviewOpen, setIsPreviewOpen] = useState(false);
+  const [previewImageUrl, setPreviewImageUrl] = useState<string | null>(null);
```
MattIPv4, line 202: Do we need two states here -- can we not just toggle the modal based on wether there is a URL or not?

apps/website/src/components/show-and-tell/ShowAndTellEntryForm.tsx
DriesOlbrechts: Yes, I think that makes more sense!

apps/website/src/components/show-and-tell/ShowAndTellEntryForm.tsx
```diff
@@ -578,6 +609,24 @@ export function ShowAndTellEntryForm({
           )}
         </Button>
       </div>
+
+      <ModalDialog
+        isOpen={!!previewImageUrl}
+        closeModal={closeModal}
+        title="Image Preview"
```
MattIPv4, line 616:
```suggestion
        title="Image Preview"
        panelClassName="max-w-3xl"
```
I think we can make this a bit easier so the preview is more helpful

apps/website/src/components/show-and-tell/ShowAndTellEntryForm.tsx
```diff
@@ -578,6 +609,24 @@ export function ShowAndTellEntryForm({
...
+          <Image
+            src={previewImageUrl}
+            alt="Form Image"
+            width={500}
+            height={500}
+            className="object-contain"
+          ></Image>
+        ) : (
+          <p>No Image to preview</p>
+        )}
```
MattIPv4, line 628:
```suggestion
        {previewImageUrl && (
          <Image
            src={previewImageUrl}
            alt="Form Image"
            width={500}
            height={500}
            className="object-contain w-full"
          ></Image>
        )}
```
1. We shouldn't need to handle the case of no image
2. Allowing the image to fill the modal

MattIPv4 (commented):
Two final nits, otherwise this looks great.

MattIPv4 (approved):

MattIPv4:
Thank you very much for working on this! 💚

PR #1122, opened 2025-03-31
Show channel in pill for homepage YouTube videos

MattIPv4:
## Describe your changes

As we have two channels that feed videos onto the homepage, it can be a bit confusing whether a video is main channel one or a highlights one. This adds a little pill under each video showing which channel it is from, and linking to that channel for further content discovery.

Pill renders as expected on all viewports.


pjeweb:
Thats nice!

pjeweb (approved):

PR #1131, opened 2025-04-04
Add Bluesky as share option + abstract share platform data

MattIPv4:
## Describe your changes

We were missing Bluesky from our standard share options, such as on /show-and-tell/give-an-hour -- this has now been added, and while doing so I realised we have two places that use near-identical platform data, so that has now been abstracted into a shared util.

Bluesky share link works as expected, other share platforms continue to be shown.


MattIPv4:
Looks like the linebreaks aren't working due to https://github.com/bluesky-social/social-app/issues/6133, oh well.

apps/website/src/utils/share.ts
```diff
@@ -0,0 +1,99 @@
...
+  url: (data: ShareProps) => string;
+}
+
+const sharePlatforms = {
+  twitter: {
+    name: "X (Twitter)",
+    description: "Share on X (Twitter)",
+    icon: IconTwitter,
+    url: ({ url, title, text }: ShareProps) => {
+      const share = new URL("https://twitter.com/intent/tweet");
```
MattIPv4, line 26:
```suggestion
      const share = new URL("https://x.com/intent/tweet");
```

pjeweb (approved):

PR #1135, opened 2025-04-04
Change Show and Tell post order to be by creation date

pjeweb:
## Describe your changes

Changes the order of show and tell posts to be by seen status, then creation date. This should prevent them from jumping in order on changes.

Check order is correct, when updating posts.


pjeweb:
Is deployed to preview.alveus.gg

MattIPv4 (approved):

PR #1138, opened 2025-04-06
Add redirects for live stream chats

MattIPv4:
## Describe your changes

Adds a very simple redirect for the Twitch live chat, and a more complex redirect for the YouTube live chat making use of the YouTube API.

Credentials have been added for preview + production, to test locally you'll need to create your own GCP project w/ API key.


apps/website/src/app/live/youtube/chat/route.ts
```diff
@@ -0,0 +1,34 @@
...
+          "max-age=1800, s-maxage=1800, stale-while-revalidate=300",
+      },
+    });
+  } catch (err) {
+    console.error("Error getting YouTube live", err);
+    return new Response("YouTube data not available", { status: 500 });
+  }
+}
+
+// Cache the response for 30 minutes
```
MattIPv4, line 32: Is a 30m cache fine here, do we want a shorter life-span?

pjeweb (approved):

apps/website/src/app/live/youtube/chat/route.ts
pjeweb: Honestly doesnt really matter, cause I dont see a lot of traffic there. I would prefer a shorter cache time for UX reasons.

pjeweb (dismissed):

Issue #1120, opened 2025-03-31
[Idea]: Add logging for all tRPC mutations


MattIPv4:
### Describe your idea / enhancement

Add a middleware into our backend usage of tRPC that logs any mutations that're called, and which user (id + username) called them.

### How would this idea / enhancement be beneficial?

Allows us to keep track of which users are calling what endpoints, for auditability and understanding activity/debugging issues.

Arnnav-A:
Are you looking for simple console logging (as done in other parts) or a dedicated log report? I can try working on this issue.

MattIPv4:
Just console logs would be fine, Vercel will handle processing it.

Issue #969, opened 2025-01-21
[Idea]: RSS feeds for Animal Quest + Book Club


MattIPv4:
### Describe your idea / enhancement

Offer RSS feeds for the Animal Quest series + Book Club books.

### How would this idea / enhancement be beneficial?

Allows folks that enjoy subscribing to RSS feeds to know about new AQ episodes as well as when a new book is selected each month for ABC.

MattIPv4:
Potentially we could also offer (proxy) the RSS feeds for the two YT channels, as well as potentially a combined YT RSS feed similar to what's on the homepage.

MattIPv4:
It was mentioned in chat today that RSS feeds of the ambassadors, and of new staff would also be useful.

We also hold chronological data for events, so it may be simple enough to add a feed for that page as well.

glipglorp7:
Hello, I'd be interested in taking a stab at this one if that's okay!

MattIPv4:
Absolutely, go for it.

glipglorp7:
@MattIPv4 thanks! Here's what I'm thinking for starters, please feel free to lmk if I'm missing something or y'all have other preferences:

### Front end

It seems like we'd want to add buttons in `Socials`,
<img width="430" alt="Image" src="https://github.com/user-attachments/assets/4fc5d971-d6fe-48a6-a029-c5685fd2ca2a" />

Schedule and Updates,
<img width="433" alt="Image" src="https://github.com/user-attachments/assets/338ebbc2-f730-49f7-97fc-705fe0d5329b" />

and maybe even here too?
<img width="826" alt="Image" src="https://github.com/user-attachments/assets/f48399a7-45ea-4bc0-8b80-da5c2c0646db" />

I'm thinking it would open a `Popover` like the push notifs one, with rows of "\<feed name\> \<button to copy URL\>".

I think it's recommended to also add `<link rel="alternate" type="application/rss+xml" ...>` elements to `<head>` for SEO/web crawlers and RSS reader autodiscovery; maybe not necessary but couldn't hurt. The first one would be considered the default/primary feed; maybe Ambassadors for now?

### Back end

As long as it's okay to add a dependency, I'm planning to use the [feed](https://www.npmjs.com/package/feed?activeTab=readme) npm package to generate new pages under `pages/feeds/`. I think each could return 15 entries for starters, to match the YouTube feeds, which should be more than enough given that most of these feeds will be updated pretty infrequently.

[...]

PR #1146, opened 2025-04-10
Update Show and Tell carousel to show images first

### LINKED ISSUE #1022: [Bug]: featuredImage is not the first image in ShowAndTellEntry

`splitAttachments` only returns the first valid thumbnail as the featured image if there are no image attachments. Meanwhile, each `ShowAndTellEntry` presents videos before images.

`featuredImage` should probably be the first image in `ShowAndTellEntry`.

### Steps to reproduce

1. Go to https://www.alveussanctuary.org/show-and-tell
2. Enter presentation view
3. Find an entry with both images and videos

### Browser specs

https://www.whatsmybrowser.org/b/47B7O

### Anything else?

I would be happy to create a PR with a fix. We could swap the logic in `splitAttachments` and have admins update the few affected posts' `dominantColor`.

louissullivan4:
## Describe your changes

- Changed ordering of carousel to add images first and then videos as suggested in the above issue
- Agreed with issue opener as it the first image attachment is considered the "Featured Image" therefore it makes sense to have images before videos.
- Because the featured image is just the first image attachment there is no need to pass this image as a new parameter to the gallery, made more sense just to switch the order

Resolves #1022

- Locally created a Show and Tell post with a video and image. Approved the post and then verified that the order had swapped

### DIFF
--- a/apps/website/src/components/show-and-tell/gallery/ShowAndTellGallery.tsx
+++ b/apps/website/src/components/show-and-tell/gallery/ShowAndTellGallery.tsx
@@ -156,23 +156,6 @@ export function ShowAndTellGallery({
   );
 
   const carouselItems = useMemo(() => {
-    const videoItems: [string, JSX.Element][] = videoAttachments
-      .filter(({ url }) => validateNormalizedVideoUrl(url))
-      .map((videoAttachment, i) => [
-        `001-video-${i}`,
-        <VideoItem
-          key={`image-${i}`}
-          videoAttachment={videoAttachment}
-          showPreview
-          openInLightbox
-          linkAttributes={{
-            className:
-              "group/trigger pointer-events-auto flex items-center justify-center cursor-pointer select-none",
-            draggable: false,
-          }}
-        />,
-      ]);
-
     const imageItems: [string, JSX.Element][] = imageAttachments.map(
       (imageAttachment, i) => [
         `002-image-${i}`,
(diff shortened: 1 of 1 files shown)


MattIPv4:
👋 I don't think swapping the order in which the items render is the right approach here -- we want the videos to be first. I think the correct way to fix this would be to look at how we determine what the featuredImage is.

louissullivan4:
Fair enough, the current implementation is to check for imageAttachments, if there is one then the featuredImage is just the first image in that list of image attachments. A suggestion would be to allow the "poster" to set select via checkbox or some other method which image they want as featured, otherwise use the first image. Thoughts on this?

MattIPv4:
@pjeweb what're you thinking here? I'm honestly wondering if this is something we should just not worry about? I'm not convinced the average user submitting a Show and Tell post is going to understand whatever UI we present to them for selecting this, and I think it doesn't really matter than the featured image isn't the _first_ item if there are videos?

pjeweb:
> @pjeweb what're you thinking here? I'm honestly wondering if this is something we should just not worry about? I'm not convinced the average user submitting a Show and Tell post is going to understand whatever UI we present to them for selecting this, and I think it doesn't really matter than the featured image isn't the _first_ item if there are videos?

Honestly thinking the same. I do not really see a need to change anything.

louissullivan4:
Will close :) Is there any idea issues ye are particularly interested in being implemented? Just looking to contribute!

MattIPv4:
Apologies for not having a closer look at that issue was filed and having you waste your time on this!

If you do want to work on something, #969 might be interesting (especially figuring out a clean common way to do the feeds), or #1120 if you want something even more backend-y.

louissullivan4:
No not at all! Was a good issue for me to understand the code base. Will take a look over the weekend at those two, thanks.

MattIPv4:
Feel free to take a look through the other issues as well, and drop a comment on any of them if you want confirmation on it still being worth working on / direction to go with it :)

Issue #1147, opened 2025-04-11
[Bug]: Annual Report Button Styles


Mini51:
### What is the current behavior?

Currently the 2023 Annual Report button is larger and bolder than the rest.

### What is the expected behavior?

All three annual report buttons should be the same size and font weight.

### Steps to reproduce

1.  Go to the annual reports page

### Browser specs

https://www.whatsmybrowser.org/b/5ZOU3

### Anything else?

![Image](https://github.com/user-attachments/assets/1d77dc9f-5e9a-42f0-a0fc-935d5a23b348)

Mini51:
I can put together a pull request which will change the tailwind classes on the "2023 Annual Report" button to match the other's.

MattIPv4:
This is very intentional, hence the intentional code for it. It is meant to be a visual guide to the most recent annual report. https://github.com/alveusgg/alveusgg/blob/80e673b4277489cba8ba89414d31056c7ed2511e/apps/website/src/pages/about/annual-reports/index.tsx#L47

PR #1148, opened 2025-04-11
Removed conditional CSS for report buttons

Mini51:
## Describe your changes
In this pull request I have updated the tailwind classes so that all of the Annual Report buttons. This can be seen in the photos below.  This will make the page looked more balanced as previously one of the buttons was larger than the others. This is related to the issue #1147 

before: 
image

after: 
image

...

...

### DIFF
--- a/apps/website/src/pages/about/annual-reports/index.tsx
+++ b/apps/website/src/pages/about/annual-reports/index.tsx
@@ -42,10 +42,7 @@ const AnnualReportsPage: NextPage = () => {
             >
               <Heading
                 level={2}
-                className={classes(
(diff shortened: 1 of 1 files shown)


MattIPv4:
> * Removing the conditional styling eliminates the emphasis on the first report heading. Please confirm if this uniform styling is intentional from a user experience standpoint.

Copilot is correct here -- as I explained in https://github.com/alveusgg/alveusgg/issues/1147#issuecomment-2795597400, this is an intentional styling decision to emphasize the most recent report. I do appreciate your contribution, but there's no point opening a PR when the associated issue has been closed/rejected.

PR #1149, opened 2025-04-11
Enable dangerouslyAllowSVG to fix Candid (GuideStar) seal

MattIPv4:
## Describe your changes

Not sure how #1145 was working for me locally before, but this should fix it properly.

/about#transparency renders the badge.


pjeweb (approved):
weird

PR #1162, opened 2025-04-16
Explicitly set Prisma client output location

MattIPv4:
## Describe your changes

Our builds have contained `Warning: You did not specify an output path for your 'generator' in schema.prisma. This behavior is deprecated and will no longer be supported in Prisma 7.0.0. To learn more visit https://pris.ly/cli/output-path` as a warning for a while now, this fixes that by setting an explicit output location.

I've opted to have it output into the existing `prisma` directory -- as this is outside `src`, I've then re-exported everything in the `src/server/db/client` file so that usages in our code can import the client using `@/server/db/client` rather than needing to do a messy relative import to outside of `src`.

You'll want to `rm -rf node_modules apps/*/node_modules` and then `pnpm install --frozen-lockfile` to purge the old generated client, install the new adapter, and generate the new client.

Check Prisma continues to work.

### DIFF
--- a/.prettierignore
+++ b/.prettierignore
@@ -3,3 +3,4 @@ pnpm-lock.yaml
 
 apps/website/.next/
 apps/website/public/push/alveus/
+apps/website/prisma/client/
--- a/apps/website/.gitignore
+++ b/apps/website/.gitignore
@@ -6,8 +6,7 @@
 /public/sitemap.xml
 
 # database
-/prisma/db.sqlite
-/prisma/db.sqlite-journal
+/prisma/client/
 
 # next.js
 /.next/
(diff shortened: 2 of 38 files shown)


MattIPv4:
```
./src/pages/api/auth/[...nextauth].ts:16:17
Type error: Excessive stack depth comparing types 'Without<BingoCreateInput, BingoUncheckedCreateInput> & BingoUncheckedCreateInput' and 'Without<BingoCreateInput, BingoUncheckedCreateInput> & BingoUncheckedCreateInput'.
  14 | import invariant from "@/utils/invariant";
  15 |
> 16 | const adapter = PrismaAdapter(prisma);
     |                 ^
  17 |
  18 | const twitchProvider = TwitchProvider({
  19 |   clientId: env.TWITCH_CLIENT_ID,
```

pjeweb (approved):

PR #1165, opened 2025-04-17
Define explicit ordering for @/ imports

MattIPv4:
## Describe your changes

I've been trying to have a consistent order to import grouping manually as I've been working on logic, but it is far from consistent if you actually review a bunch of different files. This Prettier change will enforce it so we don't have to think about it again (grouping imports, sorting imports, sorting imported attributes).

The only thing I don't like is that there is not an option to sort `type` imported attributes before/after regular imported attributes, they're treated the same. _But_, either way, its all automatic so it'll be consistent without needing to think.

Automatic fixes commit, as the name implies, was generated from `pnpm format` so shouldn't be too much of a worry. The other commits should be individually reviewed.


apps/website/src/components/shared/form/MapPickerField.tsx
```diff
@@ -1,22 +1,27 @@
...
+import maplibregl, { GeolocateControl, Map, type Marker } from "maplibre-gl";
+import { useCallback, useEffect, useRef, useState } from "react";
+
+import "maplibre-gl/dist/maplibre-gl.css";
+// Actual map CSS
+import "@maplibre/maplibre-gl-geocoder/dist/maplibre-gl-geocoder.css";
+
+import mapStyle from "@/data/map-style";
+
+// Map search box CSS
```
pjeweb, line 11: i think we can just remove those comments?!

apps/website/src/components/show-and-tell/CommunityMap.tsx
```diff
@@ -1,17 +1,23 @@
+import { type GeoJSONSource, Map, Popup } from "maplibre-gl";
 import { useCallback, useEffect, useRef, useState } from "react";
-import { Map, type GeoJSONSource, Popup } from "maplibre-gl";
-import "maplibre-gl/dist/maplibre-gl.css"; // Import the MapLibre CSS
+
+import "maplibre-gl/dist/maplibre-gl.css";
+
+// Import the MapLibre CSS
```
pjeweb, line 6: dito

pjeweb (changes requested):

pjeweb (approved):

PR #1194, opened 2025-04-30
Fix notification images being garbage collected after upload

### LINKED ISSUE #771: [Bug]: Images uploaded for notifications expire after 1h

The notification file upload is missing the call to `checkAndFixUploadedImageFileStorageObject` so it gets garbage collected after 1h.

Valid uploads stay available at least as long as the notification is active or forever. Invalid uploads should show an error (non-image files).

### Steps to reproduce

Go to `/admin/notifications`, upload file, send notification with image, check the `FileStorageObject` in db expires within 1h.

### Browser specs

n/a

### Anything else?

_No response_

StunnaDawg:
## Describe your changes

Added the fileStorageObjectId to createNotification mutation data, which enables the call to **checkAndFixUploadedImageFileStorageObject** on the server side. This prevents notification image files from being garbage collected after 1 hour.

resolves #771  

...

1. Go to /admin/notifications
2. Upload a file and send a notification with an image
3. Check that the FileStorageObject in the database has expiresAt set to NULL instead of a timestamp

...


apps/website/src/server/db/show-and-tell.ts
```diff
@@ -215,7 +215,6 @@ async function createImageAttachment(attachment: CreateImageAttachment) {
   const { error } = await checkAndFixUploadedImageFileStorageObject(
     attachment.fileStorageObjectId,
   );
-
```
MattIPv4, line 218: Can you revert the change to this unrelated file?

MattIPv4:
@pjeweb just pushed this up to preview if you wanna review it (I can too but I figure you know this area better than I)

pjeweb (approved):
great work, thank you @StunnaDawg

PR #1209, opened 2025-05-07
Added support for  AVIF/HEIC/HEIF uploads

### LINKED ISSUE #988: [Idea]: Support AVIF/HEIC/Motion photo uploads

If possible, can we support uploading those formats, probably converting them to still photos, mostly for show and tell.

Alternatively: Add some sort of explanation/workaround for people trying to upload those.

Prevent errors for people trying to upload those.

StunnaDawg:
## Describe your changes
- Added HEIF, HEIC, and AVIF type support to ImageMimeType, 
- Added [heic-decode](https://github.com/catdad-experiments/heic-decode) dependancy so sharp can convert HEIC/HEIF images to JPEG format.
- Utilized ModalDialogue to notify users when their image is being converted to JPEG format

### New Files
website/src/pages/api/convert-image.ts - Next.js API route for image conversion 
website/src/utils/image-converter.ts - utility for handling image conversion
website/src/types/heic-decode.d.ts - types for [heic-decode](https://github.com/catdad-experiments/heic-decode)

### Flow
- File type is checked within UploadAttachmentField when a file is added
- If file type is AVIF, HEIC, or HEIF, the imageConverter function is called
- imageConverter calls the convert-image API route
- API route utilizes Sharp and heic-decode (if file is HEIC/HEIF format) to convert the image to JPEG format
- Converted image is returned to the user and uploaded

resolves #988

...

Upload AVIF/HEIC/HEIF photo files into UploadAttachmentFields

...


MattIPv4:
👀 Thanks for working on this! My immediate thought after looking at the diff here is around security -- is there a way we can do this client-side when the user attaches the image, rather than exposing an endpoint that is having to process user-submitted payloads (and is also open to the entire web to be used outside of Show and Tell)?

StunnaDawg:
> 👀 Thanks for working on this! My immediate thought after looking at the diff here is around security -- is there a way we can do this client-side when the user attaches the image, rather than exposing an endpoint that is having to process user-submitted payloads (and is also open to the entire web to be used outside of Show and Tell)?

AH makes sense, I understand! I will do a re work and do my best to make this happen @MattIPv4

apps/website/src/components/shared/form/UploadAttachmentsField.tsx
```diff
@@ -238,6 +257,23 @@ export const UploadAttachmentsField = ({
           let extractColor: PendingUploadFileReference["extractColor"] =
             async () => await extractColorFromImage(dataURL);
           let fileToUpload = file;
+
+          // Convert image to jpeg if it's a heic, avif or heif file
+          if (
+            file.type === "image/heic" ||
+            file.type === "image/avif" ||
+            file.type === "image/heif"
```
MattIPv4, line 265: Seeing this made me think, could we structure `image-converter.ts` such that its an object that maps image types to converter methods -- we could then use the keys of the object here to know whether we need to call a converter or not, rather than relying on the "magic" strings here matching the conditions handled in the single method currently in `image-converter.ts`?

apps/website/src/components/shared/form/UploadAttachmentsField.tsx
```diff
@@ -149,6 +152,21 @@ export const useUploadAttachmentsData = (
   return { files: fileReferences, dispatch };
 };
 
+async function handleImageConversion(file: File) {
+  if (
+    file.type === "image/heic" ||
+    file.type === "image/heif" ||
+    file.type === "image/avif"
```
MattIPv4, line 159: I think this condition is redundant given the identical condition wrapping this method call below, but I imagine this can go away completely if my suggestion of mapping image types to a converter makes sense to you

apps/website/src/components/shared/form/UploadAttachmentsField.tsx
StunnaDawg: Gotcha, I understand! I will work on this improvement

apps/website/src/utils/image-converter.ts
```diff
@@ -0,0 +1,91 @@
...
+  setIsConvertingFile: Dispatch<SetStateAction<boolean>>,
+  setError: Dispatch<SetStateAction<string | null>>,
+) {
+  let convertedFile: File;
+
+  const converter = conversionFunctions[file.type as SupportedMimeTypes];
+  if (converter) {
+    setIsConvertingFile(true);
+    try {
+      convertedFile = await converter(file);
```
MattIPv4, line 80: Given this is async, can we chain `.catch` and `.finally` (so we can just return the Promise) instead of a `try/catch/finally` with `await`?

apps/website/src/utils/image-converter.ts
```diff
@@ -0,0 +1,91 @@
...
+    };
+
+    img.onerror = () => {
+      URL.revokeObjectURL(url);
+      reject(new Error("Failed to load AVIF image"));
+    };
+
+    img.src = url;
+  });
+  return convertedFile;
```
MattIPv4, line 56:
```suggestion
  return new Promise((resolve, reject) => {
    const img = new Image();
    const url = URL.createObjectURL(file);

    img.onload = () => {
      const canvas = document.createElement("canvas");
      canvas.width = img.width;
      canvas.height = img.height;
      const ctx = canvas.getContext("2d");
      if (!ctx) {
        URL.revokeObjectURL(url);
        reject(new Error("Canvas context not available"));
        return;
      }
      ctx.drawImage(img, 0, 0);
      canvas.toBlob(
        (blob) => {
          URL.revokeObjectURL(url);
          if (!blob) {
            reject(new Error("Conversion to JPEG failed"));
            return;
          }
          resolve(
            new File([blob], file.name.replace(/\.avif$/i, ".jpg"), {
              type: "image/jpeg",
            }),
          );
        },
        "image/jpeg",
        0.9,
      );
    };

    img.onerror = () => {
      URL.revokeObjectURL(url);
      reject(new Error("Failed to load AVIF image"));
    };

    img.src = url;
  });
```
The constant here seems redundant, as does calling `await`?

apps/website/src/utils/image-converter.ts
```diff
@@ -0,0 +1,91 @@
+import type { Dispatch, SetStateAction } from "react";
+
+async function convertHeicToJpeg(file: File) {
+  const heic2any = (await import("heic2any")).default;
+  const blob = await heic2any({
+    blob: file,
+    toType: "image/jpeg",
+    quality: 0.9,
+  });
+  return new File([blob as Blob], file.name.replace(/\.[^.]+$/, "jpeg"), {
```
MattIPv4, line 10:
```suggestion
  return new File([blob as Blob], file.name.replace(/\.[^.]+$/, ".jpeg"), {
```
Your regex pattern includes the `.` so I imagine you want it in the replacement string too?

apps/website/src/utils/image-converter.ts
```diff
@@ -0,0 +1,91 @@
...
+      ctx.drawImage(img, 0, 0);
+      canvas.toBlob(
+        (blob) => {
+          URL.revokeObjectURL(url);
+          if (!blob) {
+            reject(new Error("Conversion to JPEG failed"));
+            return;
+          }
+          resolve(
+            new File([blob], file.name.replace(/\.avif$/i, ".jpg"), {
```
MattIPv4, line 39:
```suggestion
            new File([blob], file.name.replace(/\.avif$/i, ".jpeg"), {
```
Consistency with the `File` constructor further up

apps/website/src/components/shared/form/UploadAttachmentsField.tsx
```diff
@@ -234,18 +238,27 @@ export const UploadAttachmentsField = ({
...
       newFiles.push(
         (async () => {
-          let dataURL = await fileToBase64(file);
+          // Run through the image converter to convert the file to a jpeg if it's a heic, avif or heif file
+          // returns the original file if conversion not needed
+          const fileToUpload = await imageConverter(
+            file,
+            setIsImageConverting,
+            setError,
+          );
```
MattIPv4, line 247: If this fails to convert, does that mean we're going to pass a heic etc. file into the logic below? Will it handle that correctly?

apps/website/src/utils/image-converter.ts
```diff
@@ -0,0 +1,91 @@
...
+  setError: Dispatch<SetStateAction<string | null>>,
+) {
+  let convertedFile: File;
+
+  const converter = conversionFunctions[file.type as SupportedMimeTypes];
+  if (converter) {
+    setIsConvertingFile(true);
+    try {
+      convertedFile = await converter(file);
+      return convertedFile;
```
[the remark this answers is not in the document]
StunnaDawg: We can, I have a habit of making variables. Makes more sense to just return it directly

apps/website/src/utils/image-converter.ts
StunnaDawg: Yep for sure

apps/website/src/components/shared/form/UploadAttachmentsField.tsx
StunnaDawg: hmm, when it fails to convert the file won't be passed. Thinking twice about it now, maybe instead of setting the error with the setError I can make a setConversionError state and display that in the modal to notify the user

MattIPv4: I don't have a strong preference for how to solve it, so whatever you think best -- I just noticed your catch currently has `return file;`, which'd be passing the heic onward to the other logic without converting, if the conversion fails.

apps/website/src/components/shared/form/UploadAttachmentsField.tsx
```diff
@@ -222,6 +229,16 @@ export const UploadAttachmentsField = ({
...
       if (!file) continue;
 
+      // Run through the image converter to convert the file to a jpeg if it's a heic, avif or heif file
+      // returns the original file if conversion not needed
+      const fileToUpload = await imageConverter(
+        file,
+        setIsImageConverting,
+        setImageConversionError,
+      );
+
```
StunnaDawg, line 239: Moved this outside of the newFiles.push so I could return without pushing through a file

apps/website/src/utils/image-converter.ts
```diff
@@ -0,0 +1,103 @@
...
+  "image/heic": convertHeicToJpeg,
+  "image/heif": convertHeicToJpeg,
+  "image/avif": convertAvifToJpeg,
+};
+
+function isConvertibleMimeType(
+  mimeType: string,
+): mimeType is keyof typeof conversionFunctions & ImageMimeType {
+  return mimeType in conversionFunctions;
+}
```
StunnaDawg, line 73: Created this function to get around the "as" typing on line 81

apps/website/src/components/shared/form/UploadAttachmentsField.tsx
```diff
@@ -222,6 +229,16 @@ export const UploadAttachmentsField = ({
...
+      if (!fileToUpload) continue;
+
       if (allowedFileTypes && !allowedFileTypes.includes(file.type)) {
```
MattIPv4, line 242: I imagine this condition and the one below want to be checking against the converted file?

apps/website/src/utils/image-converter.ts
```diff
@@ -0,0 +1,103 @@
...
+
+export async function imageConverter(
+  file: File,
+  setIsConvertingFile: Dispatch<SetStateAction<boolean>>,
+  setImageConversionError: Dispatch<SetStateAction<string | null>>,
+) {
+  if (isConvertibleMimeType(file.type)) {
+    const converter = conversionFunctions[file.type];
+
+    if (!converter) return file;
```
MattIPv4, line 83: This check should be redundant now given the type-guard?

apps/website/src/utils/image-converter.ts
```diff
@@ -0,0 +1,103 @@
...
+
+    setIsConvertingFile(true);
+    return await converter(file)
```
MattIPv4, line 86:
```suggestion
    return converter(file)
```
You should be able to return the promise directly here?

apps/website/src/components/shared/form/UploadAttachmentsField.tsx
StunnaDawg: Ah yes, didn't catch that.

apps/website/src/utils/image-converter.ts
StunnaDawg: I thought the same thing but typescript insisted on me having that there. That check returns (just realized should not return a file) if the file passed into the converter is somehow undefined. I can try and make the type guard stricter to resolve this.

apps/website/src/utils/image-converter.ts
```diff
@@ -0,0 +1,102 @@
...
+    };
+
+    img.src = url;
+  });
+}
+
+const conversionFunctions: Record<
+  "image/heic" | "image/heif" | "image/avif",
+  (file: File) => Promise<File>
+> = {
```
StunnaDawg, line 64: Removed the Partial typing as it was making me add another check before the promise chain. Since we are already checking with isConveritableMimeType file type will already be one of these 3.

MattIPv4: I'm not sure I'm a fan of removing that partial -- there is now nothing stopping us accidentally adding a file type here that isn't in `ImageMimeType`?

apps/website/src/utils/image-converter.ts
```diff
@@ -0,0 +1,102 @@
...
+  "image/heic": convertHeicToJpeg,
+  "image/heif": convertHeicToJpeg,
+  "image/avif": convertAvifToJpeg,
+};
+
+function isConvertibleMimeType(
+  mimeType: string,
+): mimeType is keyof typeof conversionFunctions & ImageMimeType {
```
MattIPv4, line 72: Per my other comment, we cannt assert here that mimeType is a member of `ImageMimeType`, there is nothing in the logic that guarantees this now

apps/website/src/utils/image-converter.ts
MattIPv4: https://tsplay.dev/NrkMDm

MattIPv4: https://tsplay.dev/N9ozjW

StunnaDawg: Ya, I agree, that makes sense. I won't try for a shortcut. I will have to make this typing as strong as I can in the isConvertibleMimeType, I am determined

MattIPv4: Take a look at that playground link, it should be what you need

MattIPv4:
Hm, in preview I'm getting back `Error uploading file: Invalid image` when submitting a post that has heic/heif/avifs attached

StunnaDawg:
> Hm, in preview I'm getting back `Error uploading file: Invalid image` when submitting a post that has heic/heif/avifs attached

Weird, I was able to upload and submit a post with heic/heif/avifs hmmm

MattIPv4:
Yup, after further testing looks to be an unrelated bug to this 👍

MattIPv4 (approved):

apps/website/src/utils/image-converter.ts
```diff
@@ -0,0 +1,98 @@
...
+      .catch((error) => {
+        setImageConversionError(
+          `Error converting image (${file.type}) to JPEG`,
+        );
+        console.error(error);
+        return;
+      })
+      .finally(() => {
+        setIsConvertingFile(false);
+      });
```
alok0, line 95: This is the equivalent code.

```typescript
try {
  setIsConvertingFile(true);
  return await converter(file);
} catch (e) {
  setImageConversionError(
    `Error converting image (${file.type}) to JPEG`,
  );
  console.error(e);
  return undefined;
} finally {
  setIsConvertingFile(false);
}
```

Minor nitpick:
On error it returns undefined (does not throw). And it sets the error message via callback. But in the code where this function is called, it is called in a loop and error messages just overwrite each other. But it probably doesn't matter if every error message is the same.

StunnaDawg: Gotcha, this would make the function more reusable and I would have the UploadAttachmentsField handle the state management.

MattIPv4:
Apologies for the delay here -- if you can get this rebased again, I'll get it deployed so @pjeweb and I can take a final look tomorrow.

pjeweb:
Thank you @StunnaDawg! This is great and the code is very clean.

I am however not 100% convinced we want the dependency though. It is hard for me to review the emscripten code in the worker and it seems to not be maintained (but maybe its just done?). There seems to be a license issue: https://github.com/alexcorvi/heic2any/issues/59

It also is pretty old. There seems to be a newer implementation, but they seem to not have a proper license (either): https://github.com/hoppergee/heic-to

Did anyone check the dependency a bit closer? Have you looked at different dependencies?

alok0:
I was thinking the same thing.

I don't think anyone takes heic/heif seriously... I was thinking maybe we could drop the dependency entirely and tell people who want to use that format to use safari?

MattIPv4:
Yeah the support for HEIC/HEIF within NPM packages seems relatively limited. A cursory search suggests these are our options:

- https://www.npmjs.com/package/heic2any published 2 years ago, 200k monthly downloads, MIT license
- https://www.npmjs.com/package/libheif-js published 7 months ago, 137k monthly downloads, LGPL license
- https://www.npmjs.com/package/heic-to published 10 days ago, 20k monthly downloads, LGPL license
- https://www.npmjs.com/package/heic-convert published 1 year ago, 111k monthly downloads, ISC license

MattIPv4:
Given heic2any is MIT licensed and relatively widely used, I do think it is probably fine to use. I'd prefer that to the LGPL options.

pjeweb:
> Given heic2any is MIT licensed and relatively widely used, I do think it is probably fine to use. I'd prefer that to the LGPL options.

According to the linked issue its kinda violating the libheif dependencies LGPL though. Its just in the package without a license as compiled emscripten: https://github.com/alexcorvi/heic2any/blob/master/src/libheif.js

pjeweb:
> I was thinking the same thing.
> 
> I don't think anyone takes heic/heif seriously... I was thinking maybe we could drop the dependency entirely and tell people who want to use that format to use safari?

This sounds reasonable. Are there other ways? I mean this is a rare thing anyways.

MattIPv4:
Ah I had missed that, and given that one of the LGPL alternatives would be fine?

I'm not sure I really agree with the approach of telling users to use a different browser, that is terrible UX honestly.

alok0:
It is an odd one, the way I understand it, HEIC files would only come from people who take photos with their iPhone then transfer the file to their Mac. Otherwise the iPhone will convert it to JPEG automatically if it isn't being transferred to a Mac.

pjeweb:
> It is an odd one, the way I understand it, HEIC files would only come from people who take photos with their iPhone then transfer the file to their Mac. Otherwise the iPhone will convert it to JPEG automatically if it isn't being transferred to a Mac. (But I have no first hand experience here)

That sounds about right, yes. Uploading directly from the iPhone should also convert to JPEG.

StunnaDawg:
Okay, so I should work on implementing a libheif-js solution. I will get going on that!

StunnaDawg:
Going to fix the pnpm-lock issues in my next push. Decided to go with the heic-to dependency as it was easier to work with. It is LGPL licensed right? So if we are good with using it we have to include it in the README?

MattIPv4:
As it's **L**GPL my understanding is that we don't have to do anything special as we're not a library bundling it?

alok0:
Technically we are bundling it. I would interpret this as akin to static linking. :(

MattIPv4:
From wiki:

> The license allows developers and companies to use and integrate a software component released under the LGPL into their own (even proprietary) software without being required by the terms of a strong copyleft license to release the source code of their own components. However, any developer who modifies an LGPL-covered component is required to make their modified version available under the same LGPL license. For proprietary software, code under the LGPL is usually used in the form of a shared library, so that there is a clear separation between the proprietary and LGPL components.

I think we're fine to just consume it like any other library.

alok0:
The way javascript bundling works though... everything is transpiled and redistributed. We aren't just sending the client an unmodified copy of the npm package.

MattIPv4:
We're not in a position to re-license everything Alveus produces under LGPL. I do believe consuming an NPM package, and not distributing a modified version of the package back into the ecosystem, only serving the NPM package as part of our project, seems compliant with LGPL. If the upstream authors disagree with that, we can discuss it with them if it arises (we'll just have to remove our usage of it).

MattIPv4:
This should be deploying to preview.alveus.gg shortly

pjeweb (approved):
ty. this is great

MattIPv4 (approved):
Tyvm for all your work and patience on this!

PR #1214, opened 2025-05-09
Add ambassador rounds stream overlay

MattIPv4:
## Describe your changes

Implements a web-based version of the ambassador rounds overlay used on stream, allowing us to more easily change what checks should be done as part of the rounds.

All mods can run the `!check` commands as shown on screen (plus an extra `!check reset` command to reset all checks). Additional trusted users can be passed via the `users` query param (either pass it multiple times or as a comma-separated string for multiple users).

Watches the AlveusSanctuary + AlveusGG channels by default, but can watch other channels via the `channels` query param (again, pass multiple times or a comma-separated string for multiple channels).

...


apps/website/src/components/overlay/Checks.tsx
```diff
@@ -0,0 +1,236 @@
...
+    <>
+      <style
+        dangerouslySetInnerHTML={{
+          __html: [
+            `@keyframes checks-${id}-item { ${animation.keyframes} }`,
+            `.checks-${id}-item { animation: ${animation.duration.total}ms checks-${id}-item infinite; will-change: scale; }`,
+            `@media (prefers-reduced-motion) { .checks-${id}-item { animation: none !important; } }`,
+          ].join("\n"),
+        }}
+      />
```
alok0, line 139: you might be able to use a single static set of keyframes if you use `animation-timing-function` rather than `animation-delay` and custom keyframes.

`animation-timing-function: linear(0 0%, 0 60%, 1 70%, 1 100%);`

alok0: something like...

```js
{
  animationTimingFunction: `linear(0 0%, 0 %{timing.delay.before / totalDuration}, 1 ${(timing.delay.before + timing.duration) / totalDuration}, 1 100%)`
}
```

```css
@keyframes animation-name {
  0% { scale: 1; }
  50% { scale: 1.15; }
  100% { scale: 1; }
}

```

MattIPv4: I think you're partially right -- I'd still need to inject the keyframes using a style, but I could use a 0%, 50%, 100% keyframe set with a custom timing function applied to the items instead to shift the keyframes around to match what the logic currently does, rather than using delay + the shifted keyframes.

alok0: css variables could be used 

```css
@keyframes animation-name {
  0% { scale: var(--animation-name-scale-from); }
  50% { scale: var(--animation-name-scale-to); }
  100% { scale: var(--animation-name-scale-from); }
}
```

alok0: yea, i think i misunderstood this. I thought there was more than one set of keyframes being created. But there is only one.

`animation-delay` works fine as it is.

the only real variable as far as I can tell is the number of items in the list.

the number of items extends the animation duration, but we want the animation itself to be a fixed time.

also... technically the number of items is known at compile time.

MattIPv4: Yeh it's only one set of keyframes that's being offset for each item. I think I'll still with the current implementation unless you feel strongly otherwise?

alok0: leave it. though I wish there was a way to avoid `dangerouslySetInnerHTML`

MattIPv4: Yeh same for sure, it's the second time I believe in this codebase I've done this pattern, both for keyframes. There just isn't a nicer way to do custom keyframes on the fly afaik

alok0: maybe we could use a library like react-style-singleton

MattIPv4: I've thought about using libraries, but it feels pretty dirty to introduce a whole library just for this -- one style element won't hurt anyone.

pjeweb (approved):

PR #1231, opened 2025-05-17
Run cam presets from commands page

MattIPv4:
## Describe your changes

Building on some of what I PoC'ed in #1230, this allows users to authenticate with an additional scope to allow us to send Twitch messages on their behalf on the commands page. If we have that scope granted by the user, we'll show an extra button next to each preset that will send the command to load that preset in the AlveusGG Twitch chat.

Auth button shows all scopes if not signed in, only shows missing scopes if already signed in. Auth button hidden if user has all the required scopes.

Run button only visible if the user is authed with the required scopes. Run button sends commands to the AlveusGG chat when clicked.

Auth button on the Twitch admin page continues to work.


pjeweb (approved):

pjeweb:
provide-auth
the scopes being hidden make sense on the commands page, but maybe we can have an option to keep showing them on the twitch api admin page?

MattIPv4:
👍 Agree, should be fixed now
image

pjeweb (approved):

PR #1236, opened 2025-05-18
Add GitHub repositories + website to system overview

MattIPv4:
## Describe your changes

Given I'd added the low-latency feed coming from OBS, I figured it would also make sense to add the website as that's what consumes that feed. With the website added, and the open-source section already at the bottom of the page, I thought it'd also be useful to include where the repositories are used in the system.

As part of doing this, the Node implementations between the overview + network diagrams were becoming increasingly similar, so I decided to just abstract them into a single shared Node component that both diagrams can use -- the only major change is that the overview diagram now shows descriptions, which I think was probably an oversight on my part originally as we've been storing descriptions in the data since the start.

Diagram is correct.

### DIFF
--- a/apps/website/src/components/tech/Network.tsx
+++ b/apps/website/src/components/tech/Network.tsx
@@ -161,7 +161,7 @@ const NetworkNode = ({
       </p>
       <div className="my-auto">
         <p className="truncate text-alveus-green-900">{data.item.name}</p>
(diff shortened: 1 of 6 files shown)


pjeweb (approved):
is it possible to align YT, twitch and website in the same column? otherwise this is good

MattIPv4:
> is it possible to align YT, twitch and website in the same column?

Not easily, trying to override how dagre generates the graph layout is _painful_

PR #1238, opened 2025-05-19
Move lola feed URL to tRPC method + check access frequently

MattIPv4:
## Describe your changes

Moves the fetching of the lola URL to a tRPC method, allowing the component to be used anywhere without needing to pass a URL down from page props.

This also allows us to write logic in the component that frequently checks if the user still has access to generate lola URLs, and if not, we can stop the player (the JWT expiring does not stop an active connection, but once expired, the JWT cannot be used to reconnect).

lola works, lola stops if the tRPC method starts returning an error.

### DIFF
--- a/apps/website/src/components/admin/ptz/LiveCamFeed.tsx
+++ b/apps/website/src/components/admin/ptz/LiveCamFeed.tsx
@@ -1,14 +1,30 @@
 import { WebRTCPlayer } from "@eyevinn/webrtc-player";
-import { useCallback } from "react";
+import { useCallback, useEffect, useState } from "react";
 
-interface LiveCamFeedProps {
-  url: string;
-}
+import { trpc } from "@/utils/trpc";
 
-const LiveCamFeed = ({ url }: LiveCamFeedProps) => {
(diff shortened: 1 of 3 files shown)


changesbyjames:
Looks sensible to me! It's essentially a heartbeat that gets the new URL because you're not swapping the url in the video every 5 mins - that's the right call to me, it would take some time to reconnect properly.

It's totally up to you but a lot of the useEffect logic that you've done here can be done with a refCallback which might make it a little cleaner? This is [the post](https://tkdodo.eu/blog/avoiding-use-effect-with-callback-refs) that made it click for me. I did a pr for it here: https://github.com/alveusgg/alveusgg/pull/1239. Either way, I've ran them locally and they work for me!

MattIPv4:
Ah yeh, I keep forgetting that refs can do cleanups now which does make this much cleaner (and much closer to what you had before I pulled it into an effect).

pjeweb (approved):

pjeweb (approved):

pjeweb (approved):

PR #1239, opened 2025-05-19
Refactored to use refCallback

changesbyjames:
## Describe your changes

...

...

### DIFF
--- a/apps/website/src/components/admin/ptz/LiveCamFeed.tsx
+++ b/apps/website/src/components/admin/ptz/LiveCamFeed.tsx
@@ -21,52 +21,23 @@ const LiveCamFeed = () => {
     setUrl((prev) => prev ?? feedUrl.data?.url);
   }, [feedUrl.isError, feedUrl.data?.url]);
 
-  const [player, setPlayer] = useState<WebRTCPlayer & { loaded: boolean }>();
+  const ref = useCallback(
+    (video: HTMLVideoElement) => {
+      if (!video || !url) return;
 
-  useEffect(() => {
-    if (!player) return;
+      const player = new WebRTCPlayer({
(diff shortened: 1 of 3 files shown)


apps/website/src/components/admin/ptz/LiveCamFeed.tsx
```diff
@@ -75,10 +46,14 @@ const LiveCamFeed = () => {
           })
           .catch((error) => {
             console.error(`[LOLA] Error unloading`, error);
+          })
+          .finally(() => {
+            player.destroy();
           });
```
MattIPv4, line 52: Do we need this extra destroy call? Looking at https://github.com/Eyevinn/webrtc-player/blob/fcd2e45cddc69649c6c7dc3db86e93c2efb8a6c0/src/index.ts#L291-L306, I think it'll throw in the same scenario that unload would (where that is nothing loaded)?

changesbyjames: My logic was that the unload call can fail if the underlying adapter's fetch call fails - this just ensures that even if that happens the event listeners & connection to the video element is all cleaned up and doesn't leak memory. I think that's very rare, especially as the fetch call is to Cloudflare, but it could happen.

MattIPv4: Ahh I see, makes sense

MattIPv4: I think the first destroy can be removed then?

changesbyjames: Yes, my bad, I missed that one!

PR #1249, opened 2025-05-21
Update presets to use cards with thumbnails + camera menu

wazix11:
## Describe your changes

Update the tech commands page to have a camera list to select from and dynamically show the presets relevant to the camera. Also updates the presets list to be a grid of cards with the preset name, description, and a thumbnail.

A lot of these images will need to be updated, renamed, or removed. These are just all of the ones I have from a few months ago and I haven't sorted through them yet to see which presets have changed or been removed.

Initial draft commit for feedback, I've tested locally and confirmed that the layout is functional on a variety of screen sizes.


apps/website/src/pages/about/tech/commands.tsx
```diff
@@ -61,6 +61,10 @@ const AboutTechPage: NextPage = () => {
     ),
   });
 
+  const [selectedCamera, setSelectedCamera] = useState<keyof typeof presets>(
```
MattIPv4, line 64: Instead of using `keyof typeof presets` here, we should make `const presets` in `presets.ts` an actual `as const satisfies Record<string, Camera>`, and then you can `export type Preset = keyof typeof presets` (this'll collide with the current `interface Preset` which can probably become `interface PresetData` or smth), which can then be used here

apps/website/src/pages/about/tech/commands.tsx
```diff
@@ -61,6 +61,10 @@ const AboutTechPage: NextPage = () => {
...
+    Object.keys(presets)[0] as keyof typeof presets,
```
MattIPv4, line 65: As to avoid using an `as` type-cast here, the `typeSafeObjectKeys` util should help

apps/website/src/pages/about/tech/commands.tsx
```diff
@@ -326,62 +330,115 @@ const AboutTechPage: NextPage = () => {
...
+            {/* Preset List */}
+            <div className="col-span-1 lg:col-span-3">
+              {selectedCamera && (
+                <Fragment key={selectedCamera}>
+                  <Heading
+                    level={3}
+                    className="scroll-mt-14 text-2xl"
+                    id={`presets:${camelToKebab(selectedCamera)}`}
+                  >
+                    {presets[selectedCamera]?.title}
```
MattIPv4, line 387: As an FYI, once the type changes are done that I mentioned above, `presets[selectedCamera]` should be guaranteed to be a value so you won't need optional chaining here or anywhere else

apps/website/src/pages/about/tech/commands.tsx
```diff
@@ -326,62 +330,115 @@ const AboutTechPage: NextPage = () => {
...
+                                alt={preset.description}
+                                fill
+                                className="object-cover transition-transform group-hover:scale-105"
+                              />
+                            ) : (
+                              <div className="flex h-full w-full items-center justify-center bg-alveus-green-50 text-xs text-alveus-green-300">
+                                No Image
+                              </div>
+                            )}
+                          </div>
```
MattIPv4, line 415:
```suggestion
                          {preset.image ? (
                            <Image
                              src={preset.image}
                              alt={preset.description}
                              width={270}
                              className="aspect-video w-full object-cover transition-transform hover:scale-105"
                            />
                          ) : (
                            <div className="flex aspect-video items-center justify-center bg-alveus-green-50 text-xs text-alveus-green-300">
                              No Image
                            </div>
                          )}
```
`fill` tends to be rather problematic w/ next/image (as I imagine you discovered) due to it using `absolute` -- I think we'd be better off to just set `w-full` on the image + `aspect-video` directly. As an aside, we should also pass an explicit width to the image, as this'll determine at what size the optimization proxy renders it at (`w-full` will override this).

MattIPv4: (As an aside, not totally convinced on the hover effect here, I feel it implies that you're able to click on the image/card as a whole)

wazix11: > (As an aside, not totally convinced on the hover effect here, I feel it implies that you're able to click on the image/card as a whole)

Yeah, the hover effect is from before adding the buttons to copy to clipboard or run the command directly, clicking the card originally just ran the command.

wazix11:
I just dumped all of the cameras and presets in for this version, any thoughts on if these should be alphabetical or in some sort of popularity order? I'd probably lean toward popularity with some grouping for the cams, for the presets I'd probably want at minimum the "home" preset to always be the first option.

apps/website/src/data/tech/presets.ts
```diff
@@ -1,393 +1,71 @@
...
+  toast,
+  patchy,
+  noodle,
+  chicken,
+  chickenindoor,
+  georgie,
+  georgiewater,
+} as const satisfies Record<string, Camera>;
 
+export type Preset = keyof typeof presets;
```
MattIPv4, line 69: Having thought about this more, I think this type should really be `Camera` and really the default const being exported should be named `cameras`?

apps/website/src/data/tech/presets.ts
```diff
@@ -1,393 +1,71 @@
...
+export type { PresetData, Camera };
```
MattIPv4, line 70: Can we make these types as exported inline where they're defined, if they're needed elsewhere?

apps/website/src/data/presets/chicken.ts
```diff
@@ -0,0 +1,73 @@
...
+import right from "@/assets/presets/chicken/right.png";
+import window from "@/assets/presets/chicken/window.png";
+
+// import barrels from "@/assets/presets/chicken/barrels.png";
+// import nuggetsleep from "@/assets/presets/chicken/nuggetsleep.png";
+
+export interface Preset {
+  description: string;
+  image?: StaticImageData;
+}
```
MattIPv4, line 19: I wonder if this interface should be put in a dedicated file that can be referenced by all of these files, instead of being duplicated code in all of them?

MattIPv4 (commented):
I think I mentioned this on Discord, but once you've got all the images done, can you run em all through some optimization locally? Get them all down to a consistent resolution and ideally compress them as much as possible (tinypng or w/e).

apps/website/src/data/tech/presets.ts
```diff
@@ -1,393 +1,76 @@
...
+import roaches from "../presets/roaches";
+import toast from "../presets/toast";
+import wolf from "../presets/wolf";
+import wolfcorner from "../presets/wolfcorner";
+import wolfden from "../presets/wolfden";
+import wolfden2 from "../presets/wolfden2";
+import wolfindoor from "../presets/wolfindoor";
+import wolfswitch from "../presets/wolfswitch";
+
+interface PresetData {
```
MattIPv4, line 33: Can this pull from the shared definition in the presets dir?

apps/website/src/pages/about/tech/commands.tsx
```diff
@@ -1,16 +1,17 @@
...
 import Image from "next/image";
-import { Fragment } from "react";
+import { Fragment, useState } from "react";
 
 import commands, {
   type CommandCategoryId,
   commandCategories,
 } from "@/data/tech/commands";
+import type { Camera } from "@/data/tech/presets";
 import presets from "@/data/tech/presets";
```
MattIPv4, line 10:
```suggestion
import presets, { type Camera } from "@/data/tech/presets";
```

PR #1257, opened 2025-05-25
Support custom image uploads for rounds checks

MattIPv4:
## Describe your changes

We may want to have rounds checks with images that aren't ambassadors, so this adds support for this via the same upload mechanism that we use elsewhere in the app.

- Creating a new check with an image works
- Updating the name/command of a check that has an image works
  - Check image persisted
- Uploading a new image to a check that already has an image works
  - Check in S3 provider (localhost:9001 locally) that the old image was removed
- Switching a check with an image to an ambassador works
  - Check in S3 provider (localhost:9001 locally) that the old image was removed
- Switching a check set to an ambassador to a custom image works

### DIFF
--- a/apps/database/prisma/schema.prisma
+++ b/apps/database/prisma/schema.prisma
@@ -443,6 +443,7 @@ model FileStorageObject {
 
   imageAttachment ImageAttachment?
   imageMetadata   ImageMetadata?
+  roundsCheck     RoundsCheck?
 
   @@index([expiresAt, deletedAt], name: "FileStorageObject_cleanup")
   @@index([expiresAt], name: "FileStorageObject_expiresAt")
@@ -483,10 +484,13 @@ model TwitchChannel {
 }
 
 model RoundsCheck {
-  id         String  @id @default(cuid())
-  name       String
-  command    String  @unique
-  ambassador String
-  order      Int
-  hidden     Boolean @default(false)
+  id                  String  @id @default(cuid())
+  name                String
+  command             String  @unique
+  ambassador          String?
+  fileStorageObjectId String? @unique
+  order               Int
+  hidden              Boolean @default(false)
+
+  fileStorageObject FileStorageObject? @relation(fields: [fileStorageObjectId], references: [id], onDelete: Cascade)
 }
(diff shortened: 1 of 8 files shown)


apps/website/src/server/trpc/router/stream.ts
```diff
@@ -34,7 +35,19 @@ export const streamRouter = router({
   getWeather: publicProcedure.query(getWeather),
 
   getRoundsChecks: publicProcedure.query(({ ctx }) =>
```
MattIPv4, line 37: Is `fileStorageObjectId` fine to expose to unauthed users?

apps/website/src/components/admin/rounds-checks/RoundsCheckForm.tsx
```diff
@@ -26,20 +47,80 @@ type RoundsCheckFormProps = {
...
+    useMemo(() => {
+      const { fileStorageObjectId, fileStorageObjectUrl } = check || {};
+      if (!fileStorageObjectId || !fileStorageObjectUrl) {
+        return [];
+      }
+
+      return [
+        {
+          status: "saved",
+          id: createId(),
```
MattIPv4, line 70: Wasn't sure what this ID was meant to be, so I just generated a new one?

apps/database/prisma/schema.prisma
```diff
@@ -482,10 +483,13 @@ model TwitchChannel {
...
-  name       String
-  command    String  @unique
-  ambassador String
-  order      Int
-  hidden     Boolean @default(false)
+  id                  String  @id @default(cuid())
+  name                String
+  command             String  @unique
+  ambassador          String?
+  fileStorageObjectId String? @unique
```
pjeweb, line 490: Does unique work with multiple `null`'s?

MattIPv4: Yeh seems to be fine locally with a bunch of ambassador checks and some upload checks

MattIPv4: Yeh looks to be: https://www.prisma.io/docs/orm/reference/prisma-schema-reference#unique

> multiple rows with `NULL` values in the same column are allowed

apps/website/src/components/admin/rounds-checks/RoundsCheckForm.tsx
pjeweb: yep that sounds correct

pjeweb (approved):

PR #1260, opened 2025-05-27
Show command previews in tooltips for presets page

MattIPv4:
## Describe your changes

Explicitly show the user what command will be run/copied when they click a button within a preset card.

Commands are correct, tooltips render correctly.


pjeweb (approved):
very cute

PR #1266, opened 2025-05-29
Allow for rounds checks to be marked hidden

MattIPv4:
## Describe your changes

Instead of needing to fully delete rounds checks that we don't want to show, allow for them to be temporarily hidden in the admin UI.

Maximum number of checks only applies to visible checks -- checks created beyond the maximum are automatically hidden and cannot be unhidden.


MattIPv4:
This is going to conflict with #1257, I would prefer to land this first and then resolve the conflicts for the uploads PR. FYI @pjeweb

PR #1267, opened 2025-05-29
Replace npm-run-all2 w/ pnpm run pattern

MattIPv4:
## Describe your changes

Turns out `pnpm run` supports regex patterns, as well as `--sequential`, allowing us to replicate `run-s` + `run-p` without the need for a dependency.

All updated scripts run as expected.


apps/website/package.json
```diff
@@ -19,8 +19,8 @@
     "build:manifest": "tsx ./build-scripts/create-manifest.ts",
     "dev:next": "next dev",
     "dev:push-sw": "tsup --watch --config sw.tsup.config.ts",
-    "build": "run-s build:next build:next-sitemap build:push-sw build:manifest",
-    "dev": "run-p build:manifest dev:*",
+    "build": "pnpm run --sequential \"/^build:.*/\"",
```
pjeweb, line 22: this relies on the alphanumeric order of the names, right? maybe its worth to keep it explicit here, if pnpm supports that?

pjeweb: oh thats fine then i think

pjeweb (approved):
cleange

PR #1285, opened 2025-06-06
Update security policy link

### LINKED ISSUE #1274: [Bug]: security.txt policy link is broken

https://www.alveussanctuary.org/security.txt contains a policy link that is broken (404):
```
Policy: https://github.com/alveusgg/alveusgg/blob/main/SECURITY.md
```

this link should point to the current URL of the security policy (not sure where that is otherwise I would have opened a PR ofc)

### Steps to reproduce

Go to https://www.alveussanctuary.org/security.txt and try to open the [Policy link](https://github.com/alveusgg/alveusgg/blob/main/SECURITY.md)

### Browser specs

not applicable

### Anything else?

_No response_

jpkhawam:
## Describe your changes

Resolves #1274.

Update security policy link, as previous link was broken (404), and bump expiry date to `2026-01-31T04:00:00.000Z`. Let me know if I should bump the expiry date to something different instead.

Compare http://localhost:3000/.well-known/security.txt with https://www.alveussanctuary.org/.well-known/security.txt

### DIFF
--- a/apps/website/public/.well-known/security.txt
+++ b/apps/website/public/.well-known/security.txt
@@ -1,4 +1,4 @@
 Contact: mailto:opensource@alveussanctuary.org
-Policy: https://github.com/alveusgg/alveusgg/blob/main/SECURITY.md
-Expires: 2025-01-31T04:00:00.000Z
+Policy: https://github.com/alveusgg/alveusgg/security/policy
+Expires: 2026-01-31T04:00:00.000Z
 Preferred-Languages: en


apps/website/public/.well-known/security.txt
```diff
@@ -1,4 +1,4 @@
 Contact: mailto:opensource@alveussanctuary.org
-Policy: https://github.com/alveusgg/alveusgg/blob/main/SECURITY.md
-Expires: 2025-01-31T04:00:00.000Z
+Policy: https://github.com/alveusgg/alveusgg/security/policy
+Expires: 2026-01-31T04:00:00.000Z
```
MattIPv4:
```suggestion
```
👀 I wonder if we should just remove the expires line? This info should be timeless now?

jpkhawam: I do agree that the text doesn't seem likely to be updated often (if at all), but that's totally up to you. Let me know if I should remove it altogether

MattIPv4: Yeh, let's remove it, ty

MattIPv4 (approved):

PR #1293, opened 2025-06-11
Add multicams to presets page and update presets

wazix11:
## Describe your changes

Added all non-indoor multicams, cams, and preset changes to the presets page. The indoor cams and multicams will have to wait to be updated after they come back online.

...

I pressed save and all the things looked good on my local instance :)


apps/website/src/data/presets/pasturefeeder.ts
```diff
@@ -0,0 +1,23 @@
+import home from "@/assets/presets/pasturefeeder/home.png";
+import stompy from "@/assets/presets/pasturefeeder/stompy.png";
+
+import type { Preset } from "../tech/cameras.types.js";
```
MattIPv4:
```suggestion
import type { Preset } from "../tech/cameras.types.ts";
```

MattIPv4 (approved):
Ty as always

PR #1353, opened 2025-07-12
Add new cameras + update presets

wazix11:
## Describe your changes

Add new cameras and update presets+preset thumbnails.


MattIPv4 (approved):
Thank you as always

PR #1379, opened 2025-07-24
Provide token for Dependabot GitHub API calls

MattIPv4:
## Describe your changes

https://github.com/dependabot/dependabot-core/issues/12253#issuecomment-3113967434

🤷 No idea, this seems like a really odd config entry to set, but we'll see.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,10 +1,18 @@
 version: 2
 
 registries:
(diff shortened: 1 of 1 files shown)


MattIPv4:
Looks like we shouldn't need this now after all: https://github.com/dependabot/dependabot-core/issues/12253#issuecomment-3114394660

PR #1402, opened 2025-08-08
Move Show and Tell images following mouse when zoomed

MattIPv4:
## Describe your changes

Building off #1390, while we're zoomed in, automatically adjust the transform origin as the mouse moves to allow for intuitive exploring of the whole image. Also, fixes the snapping to center issue noted in that PR.

Click to zoom in, move mouse to explore whole zoomed in image, moving mouse outside the bounds etc. does not cause any weird issues, click to zoom out zooms out at the final location without snapping to center.


pjeweb (approved):
Very nice

PR #1407, opened 2025-08-08
Use custom DOM logic to close Lightbox on background click

MattIPv4:
## Describe your changes

An alternative approach to #1406, based on similar logic from the extension, to allow clicks to essentially pass through transparent divs within the lightbox and hit the backdrop to trigger the Lightbox to close.

Intuitive behaviour?


pjeweb:
So I get this approach, but it feels weird with the dragging area. The cursor for dragging shows, but when you drag in the transparent area, the lightbox will close upon click release (mouseup).

I was thinking we could disable dragging in the lightbox completely, cause the threshold to swipe a slide is too big for desktop/cursor interaction anyways imho. But dragging is super nice on touch devices. So not sure.

https://github.com/user-attachments/assets/2c3f37cf-5785-4076-ae6f-7c47c6b48b82

pjeweb:
I guess you could detect dragging vs clicking in the lightbox too or stop clicks from the carousel if they were detected as drag.

MattIPv4:
Ah yes good call, added some extra logic that should handle doing drag detection. Does that make this feel more intuitive?

pjeweb (approved):
this feels nice yeah

PR #1425, opened 2025-08-26
Add position swap buttons to presets page

MattIPv4:
## Describe your changes

While the site doesn't currently have awareness of what cams are on stream, so we can't offer more intelligent swapping buttons, this adds a set of buttons near the top of the presets page to allow swapping two positions with the cam layout, irrespective of what cameras are in each of the positions (generating `!swap <num> <num>` commands).

Clicking one position button highlights the others and changes them to command buttons. Clicking the same button again resets the state.

With one button clicked, clicking another runs the swap command for the two position buttons clicked, and then resets the state.

### DIFF
--- a/apps/website/src/components/content/Commands.tsx
+++ b/apps/website/src/components/content/Commands.tsx
@@ -54,7 +54,7 @@ const Commands = ({
     {commands.map((command) => (
       <div
         key={command.name}
-        className="mb-4 flex flex-col items-baseline lg:mb-0 lg:flex-row lg:gap-2"
+        className="mb-4 flex flex-col items-start lg:mb-0 lg:flex-row lg:gap-2"
       >
         <dt>
           <pre>


apps/website/src/components/shared/actions/ActionButton.tsx
```diff
@@ -48,7 +48,7 @@ const ActionButton = ({
         {tooltip.text}
       </Tooltip>
       <button onClick={onClick} title={tooltip.text}>
-        <Icon className="m-1 inline size-5 cursor-pointer transition-colors" />
+        <Icon className="m-1 size-5 cursor-pointer transition-colors" />
```
MattIPv4, line 51: I _think_ this change is safe, I'm not sure why this was ever `inline`?

pjeweb (approved):

pjeweb (approved):

PR #1437, opened 2025-09-02
Allow filtering by event categories on schedule

MattIPv4:
## Describe your changes

We've had all the categories shown below the calendar for a long time now, and they even had a hover state implying they did something, but they didn't...

Well, now they do! They're now checkboxes that allow you to filter the events shown in the calendar. You can select just one category, or multiple, or none at all to show all events.

Checkboxes can be interacted with as normal to filter events displayed on the calendar.


pjeweb (approved):
very nice. you might be able to go to 3+1 grid split instead of 2+1 for xl to make the calendar a bid wider, but this is nice

MattIPv4:
2+1 is what we had before, so figured best to stick with that

PR #1438, opened 2025-09-02
Minor styling tweaks for schedule page

MattIPv4:
## Describe your changes

Fast-follow for #1437 to fix two styling issues I noticed in prod.

...

### DIFF
--- a/apps/website/src/components/notifications/RecentNotifications.tsx
+++ b/apps/website/src/components/notifications/RecentNotifications.tsx
@@ -3,16 +3,18 @@ import { trpc } from "@/utils/trpc";
 
 import { NotificationEntry } from "@/components/notifications/NotificationEntry";
 
-const backgrounds = [
-  "bg-alveus-green/40 hover:bg-alveus-green/50",
-  "bg-alveus-green/36 hover:bg-alveus-green/46",
-  "bg-alveus-green/32 hover:bg-alveus-green/42",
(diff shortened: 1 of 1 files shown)


apps/website/src/components/notifications/RecentNotifications.tsx
```diff
@@ -4,16 +4,15 @@ import { trpc } from "@/utils/trpc";
...
+  "bg-alveus-green/40 hover:bg-alveus-green/50",
+  "bg-alveus-green/36 hover:bg-alveus-green/46",
+  "bg-alveus-green/32 hover:bg-alveus-green/42",
+  "bg-alveus-green/28 hover:bg-alveus-green/38",
+  "bg-alveus-green/24 hover:bg-alveus-green/34",
+  "bg-alveus-green/20 hover:bg-alveus-green/30",
+  "bg-alveus-green/16 hover:bg-alveus-green/26",
+  "bg-alveus-green/12 hover:bg-alveus-green/22",
+  "bg-alveus-green/10 hover:bg-alveus-green/20",
 ];
```
MattIPv4: This fade is a lot more obvious with this change, but I'm wondering if it should actually be an opacity fade, rather than a background fade, so that it includes the text?

image

It almost feels like you could take this UI to be placing more emphasis on the oldest notif, now that I think about it? And I would think that doing a full opacity fade would reverse that?

pjeweb (approved):

Issue #1434, opened 2025-08-29
[Bug]: Show and Tell: Marking posts as seen does not (reliably) clear frontend cache


pjeweb:
### What is the current behavior?

Marking (multiple) posts as seen does n't invalidate page cache as it should. The number of unseen posts does not update on the list page.

### What is the expected behavior?

Marking (multiple) posts as seen should update the list page.

### Steps to reproduce

- Note number of unseen posts on list page in presentation mode.
- Mark previously unseen posts as seen.
- DO NOT SAVE ANY POST.
- Note number of unseen posts on list page in presentation mode did NOT change. 

### Browser specs

n/a

### Anything else?

Saving a post does correctly invalidate cache.

pjeweb:
<img width="661" height="106" alt="Image" src="https://github.com/user-attachments/assets/4f254e7a-74e9-4d06-9f96-b7d20dbcee92" />

Looks like the revalidate URL is too long. Since we do not even use the page ids anymore (since we dont have cached details pages per post), we could just drop the ids from the revalidate request.

The calling code: https://github.com/alveusgg/alveusgg/blob/3c4e63b37e0ac132ef06b354a605c7c7e40bc276/apps/website/src/server/db/show-and-tell.ts#L179-L196

alok0:
How long does this take to get a count of the map features (and derive the "post from new location")? It looks like it is iterating through the entire table

https://github.com/alveusgg/alveusgg/blob/fe2809ea49b68cd565a418dd370404e5792fcd9c/apps/website/src/pages/show-and-tell/index.tsx#L81-L94

MattIPv4:
> How long does this take to get a count of the map features (and derive the "post from new location")? It looks like it is iterating through the entire table

Yeah, that is a pretty nasty loop, welcome to PRs to optimize that if you see anything obvious. I would think it could probably be done at least somewhat more efficiently by doing the evaluations as a SQL query over the entire DB, rather than in JS in memory.

alok0:
Do we have a way to profile/measure where the actual slow part is?

MattIPv4:
Not really, no. Given this seems to only happen when lots of posts are marked as seen, I do suspect it is to do with the post IDs in the URL, so presumably rendering the 404 page a bunch of times. I think the first action should be to drop those from the revalidation request and see if this continues to be a problem?

pjeweb:
Resolved by #1455

Issue #1322, opened 2025-07-01
[Idea]: S&T Map - Different approach to "New Location" badge


Taizunz:
### Describe your idea / enhancement

I propose a different approach to showing the "New Location" badge on Show & Tell Posts: If a post is from a location that is a certain distance from existing map pins, assign it the "New Location" badge.

This "distance" setting can be adjusted to indirectly fine-tune the frequency at which the badge would show up on new posts. Over time it would probably make sense to lower the setting as well, as more and more pins are added to the map.

### How would this idea / enhancement be beneficial?

At some point in time enough people will have submitted posts to S&T that we'll rarely ever see the "New Location" badges again. At some point we could also end up in a situation where there's no longer any new countries to visit, and the badge would become obsolete.
This idea allows for the badge to continue showing up, and also indirectly allows fine-tuning the frequency by specifying a distance setting.

alok0:
Tangentially related, right now unique locations and unique countries are counted based on the names of locations and is done by iterating over the entire table.

If the "New Location" badge is stored as a boolean on the row then unique locations could be calculated as count of the number of new locations.

(countries would probably need a country column)

MattIPv4:
Agree with the notion of storing badgs per-row instead of computing it continually. The only edge case I can think of that we might need to account for is say if an older post were to change its location to be the same as a newer post that is currently considered to have the "new location" badge -- should we have logic to handle that, or should the newer post still hold the badge?

PR #1453, opened 2025-09-15
make show and tell pre-render faster?

alok0:
For https://github.com/alveusgg/alveusgg/issues/1434

## Describe your changes

I'm not sure if this is the correct way to do this. Does it make sense to not query the map data in the pre-render and do it in a trpc query?

### DIFF
--- a/apps/website/src/pages/show-and-tell/index.tsx
+++ b/apps/website/src/pages/show-and-tell/index.tsx
@@ -13,7 +13,6 @@ import {
 import scrollIntoView from "smooth-scroll-into-view-if-needed";
 
 import {
-  getMapFeatures,
   getPostsCount,
   getPostsToShow,
   getPublicPosts,
@@ -22,7 +21,6 @@ import {
 
 import { classes } from "@/utils/classes";
 import { delay } from "@/utils/delay";
-import { extractInfoFromMapFeatures } from "@/utils/locations";
 import { trpc } from "@/utils/trpc";
 
 import useOnToggleNativeFullscreen from "@/hooks/fullscreen";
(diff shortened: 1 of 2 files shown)


MattIPv4:
Hm, I understand the intent here to make the initial render faster, but this is going to put substantially more load on the DB as this query now needs to hit the DB for every page load instead of being pre-baked into the page just once at render?

alok0:
I think a real solution would require a database migration. Also a re-architecture of how "post from new location" works.

Saving the country separate from "location" would make counting the unique locations and countries possible.

But the "post from new location" right now is derived based on order of the posts by time.

Might even consider how https://github.com/alveusgg/alveusgg/issues/1322 would factor into this.

But in the short term, I wonder if there is a way to make the show and tell re-validation work consistently. Perhaps a separately cached endpoint with this data?

MattIPv4:
> But in the short term, I wonder if there is a way to make the show and tell re-validation work consistently. Perhaps a separately cached endpoint with this data?

What indication do you have that this logic is what's causing the issue? The reported issue specifically focuses on when many posts are updated at once -- if a single post is updated, this same page is invalidated just fine as far as I'm aware, which'd suggest this logic isn't the problem?

MattIPv4:
> I think a real solution would require a database migration. Also a re-architecture of how "post from new location" works.
> 
> Saving the country separate from "location" would make counting the unique locations and countries possible.
> 
> But the "post from new location" right now is derived based on order of the posts by time.
> 
> Might even consider how #1322 would factor into this.

Yeah, I agree that thinking through that issue as part of reworking this logic would be sensible.

alok0:
ok, i'll scrap this and start over.

PR #1455, opened 2025-09-15
Remove postId from Show and Tell revalidation calls

alok0:
https://github.com/alveusgg/alveusgg/issues/1434

### DIFF
--- a/apps/website/src/pages/api/show-and-tell/revalidate.ts
+++ b/apps/website/src/pages/api/show-and-tell/revalidate.ts
@@ -4,6 +4,9 @@ import { env } from "@/env";
 
 import timingSafeCompareString from "@/server/utils/timing-safe-compare-string";
 
+/**
+ * @deprecated 2025-09-15
+ */
 export default async function handler(
   req: NextApiRequest,
   res: NextApiResponse,
--- a/apps/website/src/server/db/show-and-tell.ts
+++ b/apps/website/src/server/db/show-and-tell.ts
@@ -176,21 +176,11 @@ export type ShowAndTellUpdateInput =
   | z.infer<typeof showAndTellUpdateInputSchema>
   | z.infer<typeof showAndTellReviewInputSchema>;
 
-async function revalidateCache(postIdOrIds?: string | string[]) {
+async function revalidateCache() {
   const url = new URL(
     `${env.NEXT_PUBLIC_BASE_URL}/api/show-and-tell/revalidate`,
   );
   url.searchParams.set("secret", env.ACTION_API_SECRET);
-
-  if (postIdOrIds) {
-    if (typeof postIdOrIds === "string") {
-      url.searchParams.set("postId", postIdOrIds);
-    } else {
-      postIdOrIds.forEach((postId) =>
-        url.searchParams.append("postId", postId),
-      );
-    }
-  }
 
   return fetch(url);
 }
@@ -278,7 +268,7 @@ export async function createPost(
       dominantColor: input.dominantColor,
     },
   });
-  await revalidateCache(res.id);
+  await revalidateCache();
   return res;
 }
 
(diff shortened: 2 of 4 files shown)


MattIPv4:
@pjeweb do you recall why we make a secondary API call here, rather than calling res.revalidate directly?

pjeweb:
> @pjeweb do you recall why we make a secondary API call here, rather than calling res.revalidate directly?

premature optimization? honestly not sure. i think it should just work directly

alok0:
Might be because the revalidate call is deep in the database code?

MattIPv4:
Yeh that's true, though I don't see why the `revalidateCache` method couldn't call `revalidatePath` from `next/cache` twice instead of the `fetch` it currently calls.

Edit: Actually, does using the methods from `next/cache` work when this is in the pages router?

MattIPv4:
Want to try that as part of this to remove another moving piece, or just land this for now given we know its a no-op change?

alok0:
It looks like the only places where it would be called are from trpc.

MattIPv4:
Sweet. Having looked at this logic more, I'm not really sure this'll change anything as these params were a no-op, but this at least cleans the logic up to make it a bit easier to trace what is happening.

The only potential time saved here is removing:

```
postIdOrIds.forEach((postId) =>
  url.searchParams.append("postId", postId),
);
```

I know from experience (https://github.com/nodejs/node/issues/51518 + https://github.com/nodejs/node/pull/51520) that `URL.searchParams` can be really expensive, though my fix has been in Node.js for a while (I don't think any other runtime has adopted this optimization though).

MattIPv4:
```
Failed to compile.
./src/server/db/show-and-tell.ts
Error:   x You're importing a component that needs "revalidatePath". That only works in a Server Component which is not supported in the pages/ directory. Read more: https://nextjs.org/docs/app/building-your-application/rendering/server-components
  |
   ,-[/vercel/path0/apps/website/src/server/db/show-and-tell.ts:2:1]
 1 | import { TRPCError } from "@trpc/server";
 2 | import { revalidatePath } from "next/cache";
   :          ^^^^^^^^^^^^^^
 3 | import { z } from "zod";
 4 | 
 5 | import {
   `----
Import trace for requested module:
./src/server/db/show-and-tell.ts
./src/pages/show-and-tell/map.tsx
> Build failed because of webpack errors
 ELIFECYCLE  Command failed with exit code 1.
```

RIP. I guess if we want to do this, we'll need to either pass through revalidate from `res` in the tRPC router, or call it directly in the tRPC router mutation definitions?

MattIPv4:
Very much a nit, but can we always do `res: NextApiResponse` rather than `resp`? Where `res` collides, I think the existing variables can be renamed to smth like `data` and still hold the same meaning?

MattIPv4:
Pushed this up to preview. @pjeweb would you mind giving this a test tomorrow?

pjeweb:
> Pushed this up to preview. @pjeweb would you mind giving this a test tomorrow?

on it now

pjeweb (approved):
works great and code looks clean

@alok0 thank you!

apps/website/src/pages/api/show-and-tell/revalidate.ts
```diff
@@ -4,6 +4,9 @@ import { env } from "@/env";
 
 import timingSafeCompareString from "@/server/utils/timing-safe-compare-string";
 
+/**
+ * @deprecated 2025-09-15
+ */
```
MattIPv4, line 9: Can you file an issue to track removing this? We can pull Vercel logs to check no other traffic is going to this that we've forgotten about.

MattIPv4 (approved):

apps/website/src/pages/api/show-and-tell/revalidate.ts
alok0: https://github.com/alveusgg/alveusgg/issues/1456

PR #1464, opened 2025-09-18
Render !recent date stamps in the correct timezone

alok0:
## Describe your changes

Because events happen in CST/CDT, events that are in the evening before midnight show up in !recent with the wrong date because they are rendered in UTC.

...

Right now this really only affects bug hunts.

...


MattIPv4 (approved):
Nice catch

PR #1485, opened 2025-10-01
Show collab counts for each year

MattIPv4:
## Describe your changes

Similar to what was done in #1474, showing counts for each of the years (groups) on the collabs page. We already show the total count in the header.

I did have to change the styling logic for the first/current year to be visible now, rather than hidden.

Counts are correct.

### DIFF
--- a/apps/website/src/pages/collaborations.tsx
+++ b/apps/website/src/pages/collaborations.tsx
@@ -19,6 +19,7 @@ import { formatDateTime } from "@/utils/datetime";
 
 import useDragScroll from "@/hooks/drag";
 import useGrouped, { type GroupedItems, type Options } from "@/hooks/grouped";
+import useLocaleString from "@/hooks/locale";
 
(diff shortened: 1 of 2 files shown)


apps/website/src/pages/collaborations.tsx
```diff
@@ -250,15 +250,15 @@ const CollaborationItems = ({
...
-          }
+          className={classes(
+            "mb-6 border-b-2 border-alveus-green-300/25 pb-2 text-4xl text-alveus-green-800 [&>a]:flex [&>a]:items-end [&>a]:justify-between",
+            index === 0 ? "mt-0" : "mt-8",
+          )}
           id={`${option}:${group}`}
           link
         >
           {name}
+          <small className="text-lg font-medium">{` (${items.length.toLocaleString()})`}</small>
```
pjeweb: Would this break on different locales? This server rendered, yes?

MattIPv4: Ah yes, that is true

PR #1493, opened 2025-10-06
Add excluded title phrases for clip player

wazix11:
## Describe your changes

Adds a few phrases to an exclusion list for the clip player. This filters out some of the less desirable clips automatically without needing to keep adding them to the env file. Going forward, we can try to have clips named including at least one of these key phrases to be confident it wont end up in the rotation.

...

Tested in local environment, current filters only remove 9 out of around 1,100 clips. I let the player run for a bit over an hour in the background and didn't see any other clips that should be excluded at this time.


apps/website/src/pages/stream/clips.tsx
```diff
@@ -91,7 +91,25 @@ export const getStaticProps: GetStaticProps<{
...
+      /frew up/i,
+      /throw up/i,
+      /throwing up/i,
+    ];
+
+    const filtered = clips.filter(
+      (clip) =>
+        !excluded.has(clip.id) &&
+        !excludedTitlePatterns.some((regex) => regex.test(clip.title)),
+    );
```
MattIPv4, line 116:
```suggestion
    const excludedClips = new Set(env.TWITCH_EXCLUDED_CLIPS);
    const excludedPhrases = [
      "animal ambassador 24\/7",
      "seizure",
      "for staff",
      "nilla",
      "limp",
      "puke",
      "vomit",
      "frew up",
      "throw up",
      "throwing up",
    ];
    const excludedTitle = new RegExp(`\\b(${excludedPhrases.join("|"})\\b`, "i");

    const filtered = clips.filter(
      (clip) =>
        !excludedClips.has(clip.id) &&
        clip.title.length > 1 &&
        !excludedTitle.test(clip.title),
    );
```
Thoughts on this approach, where we have an array of strings, and then generate a single RegExp? I think we probably want all of these phrases to have a word boundary on both sides of them?

(And the single character check can just be a length check instead of a RegExp)

MattIPv4: And then as a follow-up thought to this if this seems logical, should this list of phrases be an env var?

wazix11: I think at least for "limp" and "vomit", we'd want them to include "limping" and "vomiting" so the word boundary at the end doesn't work for those. The one Nilla clip that this is filtering currently (because it is only the last year) has "Nilla's" so the word boundary on each end works, but if that punctuation wasn't there, then it wouldn't.

MattIPv4: I see, I think we probably want to add those as separate phrases? You've got puke here, but not puking ;)

MattIPv4: And, fwiw, there's nothing stopping these phrase strings having regex in them, so you could do `"nillas?"`

wazix11: Yeah true, I'm also not all that worried about covering all possibilities, just ideally clearing most out so very few have to be added to the manual exclude list. I'd also think it would be an improvement to just try this setup you suggested out and see if any "bad" clips end up in the rotation again.

MattIPv4: 👍 Sounds good. Do you want to commit this then? If we need to add more phrases, that seems like a good time to then move this to an env var.

MattIPv4:
Looks like this filters out 9 clips, vs. 2 clips on current prod.

PR #1514, opened 2025-10-13
Add new parrot cams with some basic initial presets

wazix11:
Adds "Littles" and "Macaws" cams and puts them into a grouping of "Parrot Cameras", also adds a few initial presets and images.


MattIPv4:
👀 Would you mind also updating `apps/website/src/data/tech/network.ts` very quickly as part of your PR to add the new cam models + remove the old one?

PR #1538, opened 2025-10-20
Add draggable slider for scrolling pixels in fullscreen

CtrlNate:
The fullscreen pixel view wasn't super user-friendly for desktop users without trackpads, as horizontal scrolling isn't intuitive with a standard mouse wheel.

image

### Changes:
- Added a custom scrollbar slider below the pixel grid for easier horizontal navigation
- Slider syncs bidirectionally with scroll position (drag slider or scroll directly)
- Added ESC key handler to exit fullscreen
- Restored full touch interaction for pixel tooltips

The native scroll functionality is preserved for trackpad/touch users, while the slider provides a clear visual affordance for desktop mouse users.

### DIFF
--- a/apps/website/src/pages/institute/pixels.tsx
+++ b/apps/website/src/pages/institute/pixels.tsx
@@ -3,7 +3,14 @@ import { type NextPage } from "next";
 import Image from "next/image";
 import { useRouter } from "next/router";
 import pluralize from "pluralize";
-import { useCallback, useEffect, useMemo, useRef, useState } from "react";
+import {
+  type ChangeEvent,
+  useCallback,
+  useEffect,
+  useMemo,
+  useRef,
+  useState,
+} from "react";
 
 import { classes } from "@/utils/classes";
 
@@ -84,7 +91,6 @@ const InstitutePixelsPage: NextPage = () => {
         document.body.style.overflow = "";
       } else {
         document.body.style.overflow = "hidden";
-        setScrollProgress(0.5);
         window.requestAnimationFrame(() => {
           if (!fullscreenRef.current) return;
           const { scrollWidth, clientWidth } = fullscreenRef.current;
@@ -96,7 +102,7 @@ const InstitutePixelsPage: NextPage = () => {
   }, []);
 
   const handleScrollbarChange = useCallback(
-    (e: React.ChangeEvent<HTMLInputElement>) => {
+    (e: ChangeEvent<HTMLInputElement>) => {
       const progress = parseFloat(e.target.value);
       setScrollProgress(progress);
       if (!fullscreenRef.current) return;
@@ -105,15 +111,6 @@ const InstitutePixelsPage: NextPage = () => {
     },
     [],
   );
-
-  const handleScroll = useCallback(() => {
-    if (!fullscreenRef.current) return;
-    const { scrollLeft, scrollWidth, clientWidth } = fullscreenRef.current;
-    const maxScroll = scrollWidth - clientWidth;
-    if (maxScroll > 0) {
-      setScrollProgress(scrollLeft / maxScroll);
-    }
-  }, []);
 
   const pixelsRef = useCallback(
     (node: HTMLDivElement | null) => {
@@ -129,9 +126,19 @@ const InstitutePixelsPage: NextPage = () => {
         e.preventDefault();
       };
 
+      const onScroll = () => {
+        const { scrollLeft, scrollWidth, clientWidth } = node;
+        const maxScroll = scrollWidth - clientWidth;
+        if (maxScroll > 0) {
+          setScrollProgress(scrollLeft / maxScroll);
+        }
+      };
+
       node.addEventListener("wheel", onWheel, { passive: false });
+      node.addEventListener("scroll", onScroll);
       return () => {
         node.removeEventListener("wheel", onWheel);
+        node.removeEventListener("scroll", onScroll);
       };
     },
     [fullscreen],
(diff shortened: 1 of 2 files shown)


apps/website/src/pages/institute/pixels.tsx
```diff
@@ -131,7 +165,7 @@ const InstitutePixelsPage: NextPage = () => {
         <div
           className={classes(
             fullscreen
-              ? "fixed inset-0 isolate z-100 flex h-screen w-screen touch-none flex-col gap-8 bg-alveus-green-900 p-4 ring-8 ring-alveus-green"
+              ? "fixed inset-0 isolate z-100 flex h-screen w-screen flex-col gap-4 bg-alveus-green-900 p-4 ring-8 ring-alveus-green"
```
MattIPv4: Removing `touch-none` makes me _very_ nervous. That's there to prevent a whole bunch of weird touch interaction edge-cases where users can try to zoom in/out the page etc.

apps/website/src/pages/institute/pixels.tsx
```diff
@@ -170,6 +205,29 @@ const InstitutePixelsPage: NextPage = () => {
...
+            <div className="relative z-10 shrink-0 rounded-lg bg-alveus-green-900/50 p-3 backdrop-blur-xs">
+              <div className="relative">
+                <input
+                  type="range"
+                  min="0"
+                  max="1"
+                  step="0.001"
+                  value={scrollProgress}
+                  onChange={handleScrollbarChange}
+                  className="relative z-10 h-3 w-full cursor-grab appearance-none rounded-full bg-alveus-green-800 shadow-inner active:cursor-grabbing [&::-moz-range-thumb]:size-6 [&::-moz-range-thumb]:appearance-none [&::-moz-range-thumb]:rounded-md [&::-moz-range-thumb]:border-0 [&::-moz-range-thumb]:bg-alveus-tan [&::-moz-range-thumb]:shadow-lg [&::-moz-range-thumb]:transition-transform [&::-moz-range-thumb]:hover:scale-110 [&::-moz-range-thumb]:active:scale-95 [&::-webkit-slider-thumb]:size-6 [&::-webkit-slider-thumb]:appearance-none [&::-webkit-slider-thumb]:rounded-md [&::-webkit-slider-thumb]:bg-alveus-tan [&::-webkit-slider-thumb]:shadow-lg [&::-webkit-slider-thumb]:transition-transform [&::-webkit-slider-thumb]:hover:scale-110 [&::-webkit-slider-thumb]:active:scale-95"
```
MattIPv4: Can we add a custom Tailwind variant that handles `::-webkit-slider-thum` + `::-moz-range-thumb` so we don't have a bunch of duplicate styles here?

apps/website/src/pages/institute/pixels.tsx
```diff
@@ -84,6 +87,37 @@ const InstitutePixelsPage: NextPage = () => {
...
+  useEffect(() => {
+    const handleEscape = (e: KeyboardEvent) => {
+      if (e.key === "Escape" && fullscreen) {
+        fullscreenToggle();
+      }
+    };
+
+    window.addEventListener("keydown", handleEscape);
+    return () => window.removeEventListener("keydown", handleEscape);
+  }, [fullscreen, fullscreenToggle]);
```
MattIPv4, line 158: You can just short-circuit this useEffect when `fullscreen` is false and not bind the event listener?

apps/website/src/pages/institute/pixels.tsx
CtrlNate: ah - i don't think this is actually necessary, i thought it was blocking the tooltips on mobile but that was my skill issue

i think we can leave it exactly for the reasons you said

apps/website/src/pages/institute/pixels.tsx
CtrlNate: let me know if this implementation is what you were going for, i haven't used custom tailwind variants before 😅

MattIPv4: Ah -- I would keep the styles themselves inline here, but just use the new `slider-thumb:` variant for them? Also, functional Tailwind plugins are rather deprecated, I think you should be able to do this w/ a CSS one? https://tailwindcss.com/docs/adding-custom-styles#adding-custom-variants

CtrlNate: ty for docs lol, will update

MattIPv4:
👀 @ACA30 would you mind rebasing/merging here -- I just landed a small fix to allow regular scrolling in fullscreen, which I think combined with this should make it much more intuitive to control.

CtrlNate:
rebased - should be good for merge now @MattIPv4

apps/website/src/pages/institute/pixels.tsx
```diff
@@ -183,6 +219,7 @@ const InstitutePixelsPage: NextPage = () => {
                   : "shadow-xl ring-4 ring-alveus-green",
               )}
               ref={pixelsRef}
+              onScroll={handleScroll}
```
MattIPv4: Can we use the ref to bind this only in fullscreen mode, instead of passing it through as a prop?

apps/website/src/pages/institute/pixels.tsx
```diff
@@ -84,6 +87,26 @@ const InstitutePixelsPage: NextPage = () => {
     });
   }, []);
 
+  const handleScrollbarChange = useCallback(
+    (e: React.ChangeEvent<HTMLInputElement>) => {
```
MattIPv4: Nit, import this type please rather than relying on the React global

apps/website/src/pages/institute/pixels.tsx
```diff
@@ -68,12 +68,15 @@ const InstitutePixelsPage: NextPage = () => {
...
   const fullscreenRef = useRef<HTMLDivElement>(null);
+  const [scrollProgress, setScrollProgress] = useState(0.5);
+
   const fullscreenToggle = useCallback(() => {
     setFullscreen((v) => {
       if (v) {
         document.body.style.overflow = "";
       } else {
         document.body.style.overflow = "hidden";
+        setScrollProgress(0.5);
```
MattIPv4: Is this needed, or will the logic below setting the scroll position feed back into this by itself?

apps/website/src/pages/institute/pixels.tsx
```diff
@@ -183,6 +219,7 @@ const InstitutePixelsPage: NextPage = () => {
                   : "shadow-xl ring-4 ring-alveus-green",
               )}
               ref={pixelsRef}
+              onScroll={handleScroll}
             />
 
             {fullscreen && (
```
MattIPv4: While we're here, could you add a z-50 to this div with the two arrows in it? Looks like it got buried at some point

MattIPv4: Or maybe this isn't even needed now with this slider that makes it very obvious there is more of the canvas to explore?

apps/website/src/pages/institute/pixels.tsx
```diff
@@ -193,6 +230,29 @@ const InstitutePixelsPage: NextPage = () => {
             )}
           </div>
 
+          {fullscreen && (
+            <div className="relative z-10 shrink-0 rounded-lg bg-alveus-green-900/50 p-3 backdrop-blur-xs">
+              <div className="relative">
```
MattIPv4: Why do we need a relative inside a relative?

apps/website/src/pages/institute/pixels.tsx
```diff
@@ -193,6 +230,29 @@ const InstitutePixelsPage: NextPage = () => {
...
+                <input
+                  type="range"
+                  min="0"
+                  max="1"
+                  step="0.001"
+                  value={scrollProgress}
+                  onChange={handleScrollbarChange}
+                  className="relative z-10 h-3 w-full cursor-grab appearance-none rounded-full bg-alveus-green-800 shadow-inner active:cursor-grabbing slider-thumb:size-6 slider-thumb:appearance-none slider-thumb:rounded-md slider-thumb:border-0 slider-thumb:bg-alveus-tan slider-thumb:shadow-lg slider-thumb:transition-transform slider-thumb:hover:scale-110 slider-thumb:active:scale-95"
```
MattIPv4: Why does this also need relative, and why does it need z-10 when the surrounding div is already z-10?

apps/website/src/pages/institute/pixels.tsx
```diff
@@ -193,6 +230,29 @@ const InstitutePixelsPage: NextPage = () => {
...
+                  title="Scroll horizontally"
+                />
+                <div className="pointer-events-none absolute inset-y-0 left-0 flex items-center px-1 text-alveus-tan/50">
+                  <IconArrowRight className="size-4 -scale-x-100" />
+                </div>
+                <div className="pointer-events-none absolute inset-y-0 right-0 flex items-center px-1 text-alveus-tan/50">
+                  <IconArrowRight className="size-4" />
+                </div>
```
MattIPv4: These are currently stacked behind the slider, but I suspect removing the z-10 on the slider will fix this?

MattIPv4: That being said, I'm also not sure if these are needed? Maybe if they were anchored to the thumb it'd look better, but I don't think that's possible?

MattIPv4 (commented):
Sorry, a few more thoughts on this now

PR #1542, opened 2025-10-20
Add self-service pixel rename page

pjeweb:
## Describe your changes

This adds a page that allows logged in users to edit Pixels (aka rename them aka change the identifier) that are linked to their twitch account used to log in or through full first and last names and email address as used to donate directly through PayPal donations.

...

Also this is currently deployed to https://preview.alveus.gg/institute/pixels/one/rename

### DIFF
--- a/apps/website/src/server/db/donations.ts
+++ b/apps/website/src/server/db/donations.ts
@@ -166,7 +166,7 @@ function filterDonationByPayPalVerification(verification: PayPalVerification) {
       {
         donatedBy: {
           path: "$.email",
-          equals: verification.email,
+          equals: verification.email.trim().toLowerCase(),
         },
       },
       {


apps/website/src/components/institute/EditPixelForm.tsx
```diff
@@ -0,0 +1,147 @@
...
+
+import MyPixelPreview from "@/components/institute/MyPixelPreview";
+import { coordsToGridRef } from "@/components/institute/Pixels";
+import DonationProviderIcon from "@/components/shared/DonationProviderIcon";
+import { Button } from "@/components/shared/form/Button";
+
+const confirmationMessage = (oldIdent: string, newIdent: string) =>
+  `Are you sure you want to change your Pixel label from ${oldIdent} to ${newIdent}? ` +
+  `You will only be able to change your Pixel every ${PIXEL_RENAME_LOCK_DURATION_TEXT}. ` +
+  `The last date to change Pixel labels will be announced on Alveus channels!`;
```
pjeweb: Very unsure about this confirmation message. Too much? Better as a modal or text box?

apps/website/src/components/institute/EditPixelForm.tsx
```diff
@@ -0,0 +1,147 @@
...
+            y={pixel.row}
+            identifier={
+              isLocked ? pixel.identifier : overrideIdentifier || identifier
+            }
+          />
+        </div>
+      </div>
+      <div className="w-full">
+        <label htmlFor={`${pixel.id}-identifier`} className="block text-sm">
+          Your label:
```
pjeweb: I would love some more clear language in general:
- pixel vs Pixel (cause its 3x3 pixels)
- Label / name / identifier / attribution?
- Rename / change / update

but maybe overthinking

apps/website/src/components/institute/EditPixelForm.tsx
```diff
@@ -0,0 +1,147 @@
...
+          id={`${pixel.id}-identifier`}
+          name="identifier"
+          autoComplete="username nickname given-name"
+          required
+          minLength={PIXEL_IDENTIFIER_MIN_LENGTH}
+          maxLength={PIXEL_IDENTIFIER_MAX_LENGTH}
+          value={identifier}
+          wrap="hard"
+          rows={2}
+          cols={20}
```
pjeweb: min/max length and rows/cols are kinda guessed. not sure how well we can prep for the printing, but this seemed reasonable for the size we have

apps/website/src/components/institute/EditPixelsForm.tsx
```diff
@@ -0,0 +1,238 @@
...
+  TAllMutation extends AnyRenameAllPixelsMutation,
+> = {
+  pixelsQuery: {
+    isError: boolean;
+    isPending: boolean;
+    isLoading?: boolean;
+    isSuccess: boolean;
+    data?: DonationPixel[];
+    error: { message: string } | null;
+  };
```
pjeweb, line 41: this is kinda ugly but i could not get the types to cooperation, cause the pixels are either coming from the search mutation OR a proper query (for "My pixels")

apps/website/src/components/institute/EditPixelsForm.tsx
```diff
@@ -0,0 +1,238 @@
...
+                  wrap="hard"
+                  rows={2}
+                  cols={20}
+                  className="block max-h-[calc(2lh+2*var(--sizing))] resize-none overflow-hidden rounded border border-gray-300 bg-white p-2"
+                  onChange={(e) => {
+                    setOverrideIdentifier(
+                      normalizePixelIdentifier(e.target.value),
+                    );
+                  }}
+                />
```
pjeweb: DRY label inputs

apps/website/src/components/shared/DonationProviderIcon.tsx
```diff
@@ -0,0 +1,20 @@
...
+  provider,
+  ...iconProps
+}: { provider: string } & IconProps) {
+  switch (provider) {
+    case "twitch":
+      return <IconTwitch {...iconProps} />;
+    case "paypal":
+      return <IconPayPal {...iconProps} />;
+    default:
+      return <IconGift {...iconProps} />;
```
pjeweb, line 16: I don think its worth adding giving block now. Maybe for the future.

apps/website/src/hooks/pixels.tsx
```diff
@@ -1,23 +1,61 @@
...
 export const PIXEL_GRID_HEIGHT = 50;
 export const PIXEL_TOTAL = PIXEL_GRID_WIDTH * PIXEL_GRID_HEIGHT;
 
+export const PIXEL_IDENTIFIER_MIN_LENGTH = 2;
+export const PIXEL_IDENTIFIER_MAX_LENGTH = 32;
+
+// FIXME: INCREASE LOCK DURATION TO 24 HOURS
+export const PIXEL_RENAME_LOCK_DURATION_TEXT = "1 minute"; // For display purposes
+export const PIXEL_RENAME_LOCK_DURATION_MS = 60 * 1000;
+
```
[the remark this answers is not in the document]
pjeweb: Thoughts on this?

apps/website/src/components/institute/PixelIdentifierInput.tsx
```diff
@@ -0,0 +1,49 @@
...
+        id={id}
+        name="identifier"
+        autoComplete="username nickname given-name"
+        required
+        minLength={PIXEL_IDENTIFIER_MIN_LENGTH}
+        maxLength={PIXEL_IDENTIFIER_MAX_LENGTH}
+        value={value}
+        wrap="hard"
+        rows={2}
+        cols={20}
```
pjeweb, line 61: min/max length and rows/cols are kinda guessed. not sure how well we can prep for the printing, but this seemed reasonable for the size we have

pjeweb:
I would love some more clear language in general:

- pixel vs Pixel (cause its 3x3 pixels)
- Label / name / identifier / attribution?
- Rename / change / update

but maybe overthinking

apps/website/src/components/institute/EditPixelsForm.tsx
pjeweb: done!

apps/website/src/hooks/pixels.tsx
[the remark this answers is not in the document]
MattIPv4: +1 for moving the duration ms to an env var and computing the text

apps/website/src/pages/institute/pixels/[mural]/rename.tsx
Comment (MattIPv4) [inline apps/website/src/pages/institute/pixels/[mural]/rename.tsx:1]: Can we move this to be `/institute/pixels/edit`? And can we add a link to this at the end of the `Can't find my pixels?` box?

Comment (MattIPv4) [reply apps/website/src/pages/institute/pixels/[mural]/rename.tsx:1]: Or, is it worth just placing these in a new box underneath the existing content on the pixels page itself?

apps/website/src/components/institute/EditPayPalPixels.tsx
```diff
@@ -0,0 +1,153 @@
...
+            <MessageBox variant="failure">
+              Failed to find your Pixels: {getPayPalPixels.error.message}
+            </MessageBox>
+          ) : null}
+
+          {getPayPalPixels.isSuccess && getPayPalPixels.data.length < 1 ? (
+            <MessageBox variant="warning">
+              No Pixels found! Please check the exact spelling including spaces
+              and capitalization. Use your PayPal donation receipt you got via
+              email as reference.
```
MattIPv4:
```suggestion
              and capitalization. Check the PayPal donation receipt you
              received via email.
```

apps/website/src/components/institute/EditPixelForm.tsx
```diff
@@ -0,0 +1,149 @@
...
+import MyPixelPreview from "@/components/institute/MyPixelPreview";
+import { PixelIdentifierInput } from "@/components/institute/PixelIdentifierInput";
+import { coordsToGridRef } from "@/components/institute/Pixels";
+import DonationProviderIcon from "@/components/shared/DonationProviderIcon";
+import { Button } from "@/components/shared/form/Button";
+
+const confirmationMessage = (oldIdent: string, newIdent: string) =>
+  `Are you sure you want to change your Pixel label from ${oldIdent} to ${newIdent}? ` +
+  `You will only be able to change your Pixel every ${PIXEL_RENAME_LOCK_DURATION_TEXT}. ` +
+  `The last date to change Pixel labels will be announced on Alveus channels!`;
```
MattIPv4:
```suggestion
const confirmationMessage = (oldIdent: string, newIdent: string) =>
  `Are you sure you want to rename your pixel from "${oldIdent}" to "${newIdent}"? ` +
  `Pixels can only be renamed once every ${PIXEL_RENAME_LOCK_DURATION_TEXT}.`;
```
I think we can reduce the wording here, and I don't think we need to reference the end here.

apps/website/src/components/institute/EditPixelForm.tsx
```diff
@@ -0,0 +1,149 @@
...
+
+export function EditPixelForm({
+  pixel,
+  provider,
+  overrideIdentifier,
+  renameMutation,
+  onRename,
+}: EditPixelFormProps) {
+  // Refresh every 30 seconds to update locked status
+  useTimestamp(30_000);
```
MattIPv4: This feels like a hack that might break with things like React Compiler auto-memoization? Do we not have a component that can render a live timestamp in the JSX where we need it (or a hook that returns a formatted value)?

apps/website/src/components/institute/EditPixelForm.tsx
```diff
@@ -0,0 +1,149 @@
...
+
+  const submit = (e: FormEvent<HTMLFormElement>) => {
+    e.preventDefault();
+
+    const formData = new FormData(e.currentTarget);
+    const newIdentifier = formData.get("identifier");
+    if (typeof newIdentifier !== "string" || !newIdentifier) return;
+
+    onRename(newIdentifier, pixel.id);
+  };
```
MattIPv4, line 53: Should we reject a submission with a new identifier that matches the current?

apps/website/src/components/institute/EditPixelForm.tsx
```diff
@@ -0,0 +1,149 @@
...
+      >
+        {renameMutation.isPending
+          ? "Saving..."
+          : isLocked
+            ? `Locked until ${formatDateTimeLocal(pixel.lockedUntil!, {
+                time: "minutes",
+                timezone: false,
+              })}`
+            : isDisabled
+              ? "- Save all Pixels -"
```
MattIPv4: I'm confused why this is the text that shows when it is disabled?

apps/website/src/components/institute/EditPixelsForm.tsx
```diff
@@ -0,0 +1,219 @@
...
+
+  const totalCount = pixelsQuery.data?.length ?? 0;
+  const overridablePixelsCount =
+    pixelsQuery.data?.filter(
+      (pixel) => !pixel.lockedUntil || pixel.lockedUntil < new Date(),
+    ).length ?? 0;
+
+  return (
+    <>
+      {totalCount > 1 ? (
```
MattIPv4, line 140: I think maybe this section should be collapsed by default?

apps/website/src/components/institute/EditPixelsForm.tsx
```diff
@@ -0,0 +1,219 @@
...
+        <>
+          <h3 className="mt-8 mb-4 font-bold">
+            Change all your Pixels at once
```
MattIPv4:
```suggestion
            Rename all your pixels
```

apps/website/src/components/institute/EditPixelsForm.tsx
```diff
@@ -0,0 +1,219 @@
...
+          </h3>
+          <Box className="bg-white p-4">
+            <div className="flex flex-row items-center justify-between leading-none">
+              <strong className="text-lg">All Pixels</strong>
+            </div>
```
MattIPv4: This labels seems redundant?

apps/website/src/components/institute/EditPixelsForm.tsx
```diff
@@ -0,0 +1,219 @@
...
+                  </Label>
+
+                  <Switch
+                    checked={enableOverridePreview}
+                    onChange={setEnableOverridePreview}
+                    className="group inline-flex h-6 w-11 items-center rounded-full bg-alveus-green-300 transition-colors data-checked:bg-alveus-green"
+                  >
+                    <span className="size-4 translate-x-1 rounded-full bg-alveus-tan transition-transform group-data-checked:translate-x-6" />
+                  </Switch>
+                </Field>
```
MattIPv4: I think perhaps the input should default to being empty (perhaps with the placeholder as the current default value you're using), and then we can automatically preview whenever there is an input, rather than having a dedicated toggle (which feels a bit too verbose)?

apps/website/src/components/institute/EditPixelsForm.tsx
```diff
@@ -0,0 +1,219 @@
...
+
+                <Button
+                  size="small"
+                  type="submit"
+                  disabled={overridablePixelsCount === 0}
+                >
+                  Change My Pixels ({overridablePixelsCount})
```
MattIPv4:
```suggestion
                  Save my pixels ({overridablePixelsCount})
```
Consistent language with the individual pixels

apps/website/src/components/institute/EditPixelsForm.tsx
```diff
@@ -0,0 +1,219 @@
...
+
+                {infoBox}
+              </div>
+            </form>
+          </Box>
+        </>
+      ) : null}
+
+      <h3 className="mt-8 mb-4 font-bold">
+        Preview and change individual Pixels
```
MattIPv4:
```suggestion
        Rename individual pixels
```

apps/website/src/components/institute/PixelPreview.tsx
MattIPv4, line 1: I suspect this should be called `PixelPreview` perhaps, given it's being used on the main `Pixels` component?

apps/website/src/components/institute/PixelIdentifierInput.tsx
```diff
@@ -0,0 +1,49 @@
...
+  id: string;
+  value: string;
+  setValue: (value: string) => void;
+  isDisabled?: boolean;
+}) {
+  return (
+    <>
+      <label htmlFor={id} className="block text-sm">
+        Your label:
+      </label>
```
MattIPv4: I don't think we really need this, perhaps `sr-only`?

apps/website/src/components/institute/PixelIdentifierInput.tsx
```diff
@@ -0,0 +1,49 @@
...
+      <textarea
```
MattIPv4, line 51: I don't think this should be a textarea w/ rows/cols, that indicates to users that they can do multi-line strings, which feels problematic? I think this should probably be a regular text input with a max length?

apps/website/src/components/institute/PixelIdentifierInput.tsx
```diff
@@ -0,0 +1,49 @@
...
+        required
+        minLength={PIXEL_IDENTIFIER_MIN_LENGTH}
+        maxLength={PIXEL_IDENTIFIER_MAX_LENGTH}
+        value={value}
+        wrap="hard"
+        rows={2}
+        cols={20}
+        disabled={isDisabled}
+        className={classes(
+          "block max-h-[calc(2lh+2*var(--sizing))] resize-none overflow-hidden rounded border border-gray-300 p-2",
```
MattIPv4: I don't understand what the max height is doing here, it seems off? As above, I think we'd be much better off with this being a regular text input?

Image

apps/website/src/hooks/pixels.tsx
```diff
@@ -11,6 +12,42 @@ export const PIXEL_GRID_WIDTH = 200;
...
+    .replace(/\r?\n+/g, "\n") // Replace multiple line breaks with a single line break
+    .replace(/[^\x20-\x7E]/g, ""); // Remove other non-printable characters
+
+  // Only keep the first line break and replace subsequent line breaks with spaces
+  const firstLineBreakIndex = normalized.indexOf("\n");
+  if (firstLineBreakIndex !== -1) {
+    const before = normalized.slice(0, firstLineBreakIndex + 1);
+    const after = normalized.slice(firstLineBreakIndex + 1).replace(/\n/g, " ");
+    normalized = before + after;
+  }
```
MattIPv4: Why are we accepting linebreaks at all, what is the intent, why would a name have linebreaks?

apps/website/src/pages/institute/pixels/edit.tsx
```diff
@@ -0,0 +1,91 @@
...
+      <Section
+        dark
+        containerClassName="flex flex-wrap gap-4 justify-between items-end"
+      >
+        <div>
+          <Heading>Alveus Research & Recovery Institute My Pixels</Heading>
+          <p>
+            You will only be able to rename your Pixel once every{" "}
+            {PIXEL_RENAME_LOCK_DURATION_TEXT}. The last date to rename Pixels
+            will be announced on Alveus channels!
```
MattIPv4: This seems like odd text to have in the page header?

apps/website/src/pages/institute/pixels/edit.tsx
```diff
@@ -0,0 +1,91 @@
...
+          </p>
+        </div>
+        <Button href="/pixels" dark className="inline-flex items-center gap-2">
```
MattIPv4:
```suggestion
        <Button href="/institute/pixels" dark className="inline-flex items-center gap-2">
```
Avoid sending the user through an extra redirect

apps/website/src/pages/institute/my-pixels.tsx
```diff
@@ -0,0 +1,91 @@
...
+          <Heading>Alveus Research & Recovery Institute My Pixels</Heading>
+          <p>
+            You will only be able to rename your Pixel once every{" "}
+            {PIXEL_RENAME_LOCK_DURATION_TEXT}. The last date to rename Pixels
+            will be announced on Alveus channels!
+          </p>
+        </div>
+        <Button href="/pixels" dark className="inline-flex items-center gap-2">
+          <IconArrowRight className="size-5 -scale-x-100" />
+          Learn More About the Pixel Project
```
MattIPv4:
```suggestion
          Explore the Completed Mural
```

apps/website/src/pages/institute/my-pixels.tsx
```diff
@@ -0,0 +1,91 @@
...
+        </Button>
+      </Section>
+
+      <Section containerClassName="space-y-4 text-lg max-w-screen-lg">
+        <Heading level={2}>Pixels connected with your Twitch account</Heading>
+        <p>
+          Here you can find all the Pixels you have placed through Twitch
+          Charity or PayPal donations using the same email address.
+        </p>
```
MattIPv4: I think this text is probably also redundant?

apps/website/src/pages/institute/my-pixels.tsx
```diff
@@ -0,0 +1,91 @@
...
+
+        <div className="mt-6">
+          {!session.data?.user ? (
+            <MessageBox>
+              <p className="mb-4">
+                You need to be logged in to view and edit your Pixels!
```
MattIPv4:
```suggestion
                You need to be logged in to view and edit pixels connected to your Twitch account.
```

apps/website/src/pages/institute/pixels/[mural]/rename.tsx
```diff
@@ -0,0 +1,91 @@
...
+      <Section
+        className="grow"
+        containerClassName="space-y-4 text-lg max-w-screen-lg"
+      >
+        <Heading level={2}>Pixels through PayPal donations</Heading>
+
+        <p>
+          Here you can search for more Pixels unlocked through PayPal donations
+          by providing the exact donation info (Full name and email address).
+        </p>
```
Comment (MattIPv4) [inline apps/website/src/pages/institute/pixels/[mural]/rename.tsx:99]:
```suggestion
        <p>
          Search for pixels to rename using the exact PayPal donation information used to unlock them.
        </p>
```
I think we can simplify this copy, though I do see the value in keeping this to emphasize the nature of needing the exact info.

apps/website/src/server/trpc/router/donations.ts
```diff
@@ -8,27 +10,108 @@ import {
...
+
+import { payPalVerificationSchema } from "@/components/institute/EditPayPalPixels";
+import { coordsToGridRef } from "@/components/institute/Pixels";
+
+export type RenameMyPixelsMutation = typeof trpc.donations.renameMyPixels;
+export type RenamePayPalPixelsMutation =
+  typeof trpc.donations.renamePayPalPixels;
+export type AnyRenamePixelsMutation =
+  | RenameMyPixelsMutation
+  | RenamePayPalPixelsMutation;
```
MattIPv4: Don't we normally create these types where we need them by importing the trpc router? This feels almost like a circular definition that these are coming from the trpc util import which presumably ultimately comes from this exact file?

MattIPv4 (commented):
I think generally all the existing copy refers to pixels in lowercase, as a regular noun rather than a proper noun, unless in a title. I think the same should apply here? It reads a bit odd to see a capitalised Pixel everywhere?

apps/website/src/components/institute/EditPixelForm.tsx
pjeweb: Its their official recommendation afaik. The hook should be clean and safe i think:
https://react.dev/reference/rules/components-and-hooks-must-be-pure#components-and-hooks-must-be-idempotent

apps/website/src/pages/institute/pixels/[mural]/rename.tsx
Comment (pjeweb) [reply apps/website/src/pages/institute/pixels/[mural]/rename.tsx:1]: I feel like I prefer it on its own page tbh.

apps/website/src/hooks/pixels.tsx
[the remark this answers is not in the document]
pjeweb: done

apps/website/src/components/institute/EditPixelForm.tsx
MattIPv4: Right, but that's for updating a string being rendered to the DOM -- a hook that's just updating itself with nothing actually exposed to the component calling it feels off. Can we use an actual hook or component that handles formatting the locked timestamp?

apps/website/src/pages/institute/pixels/[mural]/rename.tsx
Comment (MattIPv4) [reply apps/website/src/pages/institute/pixels/[mural]/rename.tsx:1]: Sure, works for me, let's just make sure it's linked to from the pixels page 👍

apps/website/src/components/institute/EditPixelForm.tsx
pjeweb: Sure but that component or hook will need to rerender every n seconds too.

pjeweb: Or I guess with a long time setTimeout

MattIPv4: > Sure but that component or hook will need to rerender every n seconds too.

Right, but that timer is then updating a state that is directly driving the DOM, whereas here we just have a hook that appears to be do nothing (and it also isolates the re-rendering so its not the entire form)

MattIPv4: a setTimeout works too tbh

pjeweb: can i do it for the whole component then or should I wrap the button and preview in components for it?

MattIPv4: I guess you'd want it to be a hook at the level of the button/preview, so that it handles updating the timestamp in the DOM with the side-effect of also causing the disable state to be re-evaluated each time?

pjeweb: See if this looks better to you

MattIPv4: Lovely, thank you!

apps/website/src/components/institute/EditPixelForm.tsx
pjeweb: We handle it server side. But we could add a check in the client too.

apps/website/src/components/institute/EditPixelForm.tsx
pjeweb: Maybe there is a better text. This is shown when the button is disabled because the user has edited the label on top in the "edit all pixels" section. Maybe it does not need to be disabled in that case?

apps/website/src/components/institute/EditPixelsForm.tsx
pjeweb: Not sure whats more common, people just want to change their name for all pixels or go in and change individual ones. I was under the impression, the former was more common.

apps/website/src/components/institute/EditPixelForm.tsx
MattIPv4: Yeh I think a client-side check would be smart, no point wasting a server request for it.

apps/website/src/components/institute/EditPixelForm.tsx
MattIPv4: I think we'd be fine to just keep showing the same text as normal, but disabled?

apps/website/src/components/institute/EditPixelsForm.tsx
pjeweb: I think it looks weird without this and/or without the headline. If you think it looks better without, we can remove it.

apps/website/src/components/institute/EditPixelsForm.tsx
MattIPv4: I was leaning toward folks wanting to see/edit individual pixels more-so than renaming all at once, and I also figured those that do want to edit all will be high intent enough on that that they'll be fine with expanding a section.

apps/website/src/components/institute/PixelIdentifierInput.tsx
pjeweb: We talked about it. We agreed two line is good. This way they have control over wrapping.

apps/website/src/components/institute/PixelIdentifierInput.tsx
pjeweb: Max height limits it to two lines.

apps/website/src/hooks/pixels.tsx
pjeweb: see above

apps/website/src/pages/institute/pixels/edit.tsx
pjeweb: Well it relates to the whole page. I didnt want another general section above the two headlines. I can also repeat it twice or somethin else. If you have any ideas?

apps/website/src/components/institute/PixelIdentifierInput.tsx
MattIPv4: I can't seem to input linebreaks into the input, no matter what I try? We're going to need to force wrapping ourselves anyway are we not, plus we're going to need to force truncation?

apps/website/src/components/institute/PixelIdentifierInput.tsx
MattIPv4: I see, but it doesn't seem to really work, I could see three lines pretty easily?

apps/website/src/pages/institute/pixels/edit.tsx
MattIPv4: Right, but we already convey this information to the user at the point they actually go to rename a pixel, it seems redundant to also have it in the header?

apps/website/src/components/institute/EditPixelForm.tsx
pjeweb: Its gonna add a bunch of code to show an error though and its just one http roundtrip. Honestly not sure its worth.

apps/website/src/components/institute/PixelIdentifierInput.tsx
pjeweb: > I can't seem to input linebreaks into the input, no matter what I try? 

Oh that seems to a bug after all the refactoring. Let me see if I can fix it. Put in a lot of effort to make that work after all.

> We're going to need to force wrapping ourselves anyway are we not, plus we're going to need to force truncation?

Thats the question. I was trying to figure out if we can find a save but low amount of chars that makes this less of a problem.

apps/website/src/components/institute/PixelIdentifierInput.tsx
pjeweb: I mean if you use the widest character that can happen lmao. I wont invest more time into it, if we wanna remove the possibility of people choosing where to wrap anyways.

apps/website/src/pages/institute/pixels/edit.tsx
pjeweb: So you would just remove it completely? I do like if I know what happens before clicking save tbh. Also I dont think we have the "The last date to rename Pixels will be announced on Alveus channels!" line anywhere else.

MattIPv4: Yeh, I think we'd be fine without this - I'm not sure that sentence about the last date really conveys that much information, I think we'd be fine saying nothing until we know when the last date actually is (I suspect it'll be a _while_ given the building doesn't even exist yet).

apps/website/src/components/institute/EditPixelsForm.tsx
pjeweb: If we change it like this, I think there are two options.

Option 1 We disable the individual pixel buttons, when the user types in the "All Pixels" input. Then I think its unclear to the user why they cant change or save the individual pixels. A verbose toggle makes this clear imho.

Option 2: We allow the individual pixels to be saved even if there is an input on "All Pixels". The preview in those pixels will not match one of the inputs then. Either its not matching the "All Pixels" input _or_ the individual input. So if you click save on that pixel, what would you expect to be saved? The input or the preview of that pixel?

apps/website/src/components/institute/PixelIdentifierInput.tsx
pjeweb: Okay got the line break fixed. Try playing around with it a bit :)

apps/website/src/components/institute/EditPixelsForm.tsx
MattIPv4: My thinking was that option 1 would be clear enough when combined with the rename all section being collapsed by default -- the user has then chosen to expand and use the rename all section, so it'd make sense that the individual buttons are then disabled while they're doing that?

Perhaps we extend that disabling even further, and completely removing the inputs/buttons while there is text in the rename all field?

apps/website/src/components/institute/PixelIdentifierInput.tsx
pjeweb: Should be fixed. It will not cut off excessive inputs anymore. A manual line break is possible but limited to one.

apps/website/src/pages/institute/pixels/edit.tsx
pjeweb: Removed it completely. Communication about the timeline will be on other channels! Locking of the pixels will just be "discovered" ;)

pjeweb:
Gonna split this up and put on hold for a bit.

apps/website/src/components/institute/EditPixelsForm.tsx
pjeweb: Implemented as suggested in last commits!

apps/website/src/pages/institute/pixels/[mural]/index.tsx
Comment (MattIPv4) [inline apps/website/src/pages/institute/pixels/[mural]/index.tsx:1]: 🤔 Why does this show as a new file, does `[mural].tsx` need removing/moving?

apps/website/src/pages/institute/pixels/[mural]/rename.tsx
```diff
@@ -0,0 +1,107 @@
...
+import { LoginWithTwitchButton } from "@/components/shared/LoginWithTwitchButton";
+import { MessageBox } from "@/components/shared/MessageBox";
+
+import IconArrowRight from "@/icons/IconArrowRight";
+
+import buildingHeroImage from "@/assets/institute/hero/building.png";
+
+import type { InstitutePixelsPageProps } from "./index";
+
+export { getStaticPaths, getStaticProps } from "./index";
```
Comment (MattIPv4) [inline apps/website/src/pages/institute/pixels/[mural]/rename.tsx:26]: Smart

apps/website/src/pages/institute/pixels/[mural]/rename.tsx
```diff
@@ -0,0 +1,107 @@
...
+
+const PixelsRenamePage: NextPage<InstitutePixelsPageProps> = ({ muralId }) => {
+  const session = useSession();
+  const mural = murals[muralId];
+
+  return (
+    <PixelProvider muralId={muralId}>
+      <Meta
+        title={`Rename your Pixels | ${mural.name} | Alveus Research & Recovery Institute`}
+        description={`View and rename your unlocked pixels of ${mural.name}.`}
```
Comment (MattIPv4) [inline apps/website/src/pages/institute/pixels/[mural]/rename.tsx]:
```suggestion
        title={`Rename Your Pixels | ${mural.name} | Alveus Research & Recovery Institute`}
        description={`View and rename your unlocked pixels on the ${mural.name} mural.`}
```

apps/website/src/pages/institute/pixels/[mural]/index.tsx
Comment (pjeweb) [reply apps/website/src/pages/institute/pixels/[mural]/index.tsx:1]: yep. fixed

apps/website/src/pages/institute/pixels/[mural]/rename.tsx
Comment (pjeweb) [reply apps/website/src/pages/institute/pixels/[mural]/rename.tsx:26]: good smart?

Comment (MattIPv4) [reply apps/website/src/pages/institute/pixels/[mural]/rename.tsx:26]: Yeh, took me a second to spot it though

apps/website/src/pages/institute/pixels/[mural]/index.tsx
```diff
@@ -258,6 +259,24 @@ const InstitutePixelsPage: NextPage<InstitutePixelsPageProps> = ({
               PayPal directly, your pixel will show your first name but you can
               also search using your PayPal email address.
             </p>
+
+            <Heading level={2} className="mt-8 text-xl">
+              Want a different name on the mural?
+            </Heading>
+            <p>
+              You can change your name on the pixels you have unlocked here.
```
Comment (MattIPv4) [inline apps/website/src/pages/institute/pixels/[mural]/index.tsx]:
```suggestion
              You can log in with Twitch or submit your PayPal donation details to change your name on any of the pixels you have unlocked on the mural.
```

apps/website/src/pages/institute/pixels/[mural]/index.tsx
```diff
@@ -258,6 +259,24 @@ const InstitutePixelsPage: NextPage<InstitutePixelsPageProps> = ({
...
+            <div className="flex justify-end">
+              <Button
+                href={`/institute/pixels/${muralId}/rename`}
+                dark
+                className="mt-4 inline-flex items-center gap-2"
+              >
+                <IconPencil className="size-5" />
+                Rename your pixels
+              </Button>
+            </div>
```
Comment (MattIPv4) [inline apps/website/src/pages/institute/pixels/[mural]/index.tsx]: I think left-align would fit more with the content here? I also wonder if this should be a <Link> instead of a full-size button, or perhaps a <Button> with reduced padding/font-size?

apps/website/src/components/institute/EditPayPalPixels.tsx
```diff
@@ -0,0 +1,162 @@
...
+              />
+            </Fieldset>
+
+            <FormButton size="small" type="submit">
+              Search
+            </FormButton>
+          </div>
+
+          {getPayPalPixels.isPending ? (
+            <MessageBox>Searching for your pixels…</MessageBox>
```
MattIPv4: Should this have a border to be consistent with the cards that show when you do have pixels?

apps/website/src/components/institute/EditPayPalPixels.tsx
```diff
@@ -0,0 +1,162 @@
...
+          ) : null}
+
+          {getPayPalPixels.isError ? (
+            <MessageBox variant="failure">
```
MattIPv4, line 123: Should this have a border to be consistent with the cards that show when you do have pixels?

apps/website/src/components/institute/EditPayPalPixels.tsx
```diff
@@ -0,0 +1,162 @@
...
+              Failed to find your pixels: {getPayPalPixels.error.message}
+            </MessageBox>
+          ) : null}
+
+          {getPayPalPixels.isSuccess && getPayPalPixels.data.length < 1 ? (
+            <MessageBox variant="warning">
```
MattIPv4, line 129: Should this have a border to be consistent with the cards that show when you do have pixels?

apps/website/src/components/institute/EditPayPalPixels.tsx
```diff
@@ -0,0 +1,162 @@
...
+              No Pixels found! Please check the exact spelling including spaces
+              and capitalization. Check the PayPal donation receipt you received
+              via email.
+            </MessageBox>
+          ) : null}
+        </div>
+      </form>
+
+      {persistedPixelsData ? (
```
MattIPv4: Should this have a length check? If you submit details with no pixels returned, you get the above error message as well as a message from inside the form?

apps/website/src/components/institute/EditPayPalPixels.tsx
```diff
@@ -0,0 +1,162 @@
...
+
+          {getPayPalPixels.isError ? (
+            <MessageBox variant="failure">
+              Failed to find your pixels: {getPayPalPixels.error.message}
+            </MessageBox>
+          ) : null}
+
+          {getPayPalPixels.isSuccess && getPayPalPixels.data.length < 1 ? (
+            <MessageBox variant="warning">
+              No Pixels found! Please check the exact spelling including spaces
```
MattIPv4:
```suggestion
              No pixels found! Please check the exact spelling including spaces
```

apps/website/src/components/institute/EditPayPalPixels.tsx
MattIPv4: Loading icon?

apps/website/src/components/institute/EditPixelsForm.tsx
MattIPv4, line 1: Ditto for all the MessageBox usages here, should they have a border to match the box styling?

apps/website/src/components/institute/EditPixelsForm.tsx
```diff
@@ -0,0 +1,220 @@
...
+  defaultIdentifier = "",
+}: EditPixelsFormProps) {
+  const submitRenameAll = (e: FormEvent<HTMLFormElement>) => {
+    e.preventDefault();
+    onRename(overrideIdentifier ?? "Anonymous");
+  };
+
+  let infoBox: ReactNode;
+  if (renameMutation.isPending) {
+    infoBox = <MessageBox>Renaming your pixels…</MessageBox>;
```
MattIPv4: Loading icon?

apps/website/src/components/institute/EditPixelsForm.tsx
```diff
@@ -0,0 +1,220 @@
...
+
+      {pixelsQuery.isError && pixelsQuery.error ? (
+        <MessageBox variant="failure">
+          Error loading pixels: {pixelsQuery.error.message}
+        </MessageBox>
+      ) : null}
+
+      {("isLoading" in pixelsQuery && pixelsQuery.isLoading) ||
+      ("isPending" in pixelsQuery && pixelsQuery.isPending) ? (
+        <MessageBox>Loading your pixels…</MessageBox>
```
MattIPv4: Loading icon?

apps/website/src/components/institute/EditPixelForm.tsx
```diff
@@ -0,0 +1,173 @@
...
+      <RenderTimeLocked lockedUntil={pixel.lockedUntil}>
+        {(isLocked) => (
+          <Button
+            type="submit"
+            size="small"
+            confirmationMessage={confirmationMessage(
+              pixel.identifier,
+              identifier,
+            )}
+            disabled={renameMutation.isPending || isLocked}
```
MattIPv4: Also disable if input is empty, or matches the current?

apps/website/src/components/institute/EditPixelsForm.tsx
```diff
@@ -0,0 +1,220 @@
...
+                  placeholder={defaultIdentifier}
+                />
+
+                <div>
+                  <div className="flex items-center gap-6">
+                    <div className="grow">
+                      <Button
+                        size="small"
+                        type="submit"
+                        disabled={overridablePixelsCount === 0}
```
MattIPv4: Disable while the input is empty as well?

apps/website/src/pages/institute/pixels/[mural]/index.tsx
Comment (pjeweb) [reply apps/website/src/pages/institute/pixels/[mural]/index.tsx]: You wanna play with it a bit? I am honestly not creative enough ;)

Comment (MattIPv4) [reply apps/website/src/pages/institute/pixels/[mural]/index.tsx]: I'd just use the Link component, with the Pencil icon after the text, that should be fine.

apps/website/src/components/institute/EditPayPalPixels.tsx
```diff
@@ -0,0 +1,167 @@
...
+
+  const verificationRef = useRef<PayPalVerification | null>(null);
+
+  const submit = async (e: FormEvent<HTMLFormElement>) => {
+    e.preventDefault();
+
+    const formData = new FormData(e.currentTarget);
+    const firstName = String(formData.get("firstName"));
+    const lastName = String(formData.get("lastName"));
+    const email = String(formData.get("email"));
```
changesbyjames, line 47: Trim & lowercase here?

PR #1544, opened 2025-10-21
Rewrite getRandomEmptySquare to have a deterministic runtime

alok0:
## Describe your changes

Instead of randomly picking squares over and over again. Pick a single random square out of possible empty squares.

...

...


MattIPv4:
Lovely! I'd just written out the same idea in Discord for tomorrow.

One related thought, if a donation gets multiple pixels, is the set of allocated pixels in this logic accounting for the previous pixels from the same donation?

changesbyjames:
This is great, thank you - yeah, if it's possible to add a fix for my bug by adding the current pixels to the "already unlocked" list we can fix this all in one PR!

```ts
const location = getRandomEmptySquare(
    [...state.pixels, ...Object.values(pixels).flat(), ...pixelsForDonation],
    this.grid,
);
```

PR #1546, opened 2025-10-21
Donations worker tags bug fix & preview configuration

changesbyjames:
## Describe your changes

This represents the already deployed fixes to the donations worker.

...

...

### DIFF
--- a/apps/donations/worker/src/managers/pixels.ts
+++ b/apps/donations/worker/src/managers/pixels.ts
@@ -166,7 +166,11 @@ export class PixelsManagerDurableObject extends DurableObject<Env> {
           const pixelsForDonation: Pixel[] = [];
           for (let i = 0; i < numberOfPixels; i++) {
             const location = getRandomEmptySquare(
-              [...state.pixels, ...Object.values(pixels).flat()],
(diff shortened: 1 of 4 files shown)


apps/donations/worker/wrangler.jsonc
```diff
@@ -112,6 +112,61 @@
         ],
       },
     },
+    "preview": {
+      "route": "https://donation-manager.preview.workers.alveussanctuary.org",
+
+      "vars": {
+        "SANDBOX": "true",
+        "SITE_URL": "https://alveusgg-private-git-private-preview-alveus.vercel.app",
```
MattIPv4:
```suggestion
        "SITE_URL": "https://preview.alveus.gg",
```
Can we point this at the main preview now?

changesbyjames: Yes! Is that a redirect or a real DNS entry, I don't want us to run into the www. problem again!

MattIPv4: Real, no redirect.

MattIPv4 (approved):

PR #1559, opened 2025-10-22
Set minimum Node.js version in engines for JSDom

MattIPv4:
## Describe your changes

Following #1555, we've just hit https://github.com/jsdom/jsdom/issues/3959 in production due to a breaking change shipped by JSDom in a **PATCH** release... Vercel uses Node.js 22.11.0 by the looks of it, and JSDOM silently required 22.12.0 or greater in the patch release.

...

### DIFF
--- a/apps/website/package.json
+++ b/apps/website/package.json
@@ -64,7 +64,7 @@
     "fast-xml-parser": "^5.3.0",
     "heic-to": "^1.3.0",
     "html-react-parser": "^5.2.7",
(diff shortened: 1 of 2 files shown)


MattIPv4:
Welp, this does absolutely nothing, Vercel just ignores this 🙃 opened a case with them to ask why they're running a Node.js 22 release from a year ago and not anything newer.

MattIPv4:
#1604 + #1696

PR #1560, opened 2025-10-22
Revert and pin JSDom to 27.0.0

MattIPv4:
## Describe your changes

Following #1555, we've just hit https://github.com/jsdom/jsdom/issues/3959 in production due to a breaking change shipped by JSDom in a **PATCH** release... this reverts us to the previous version as we do not appear to be able to force Vercel to run a supported version of Node.js 22 (#1559).

...


MattIPv4:
Admin merging to get main/production back into a good state.

PR #1568, opened 2025-10-22
Move mural to be served via site

changesbyjames:
## Describe your changes

Currently we're using a durable object to serve the mural. Now it is complete, this is unnecessary. This is step 1, moving the site to use a static, CDN cacheable, endpoint that works better with the upcoming rename feature.

Step 2 will be decommissioning the pixel durable object, that will come later.

...

### DIFF
--- a/apps/website/src/app/api/pixels/route.ts
+++ b/apps/website/src/app/api/pixels/route.ts
@@ -5,6 +5,11 @@ const headers = {
 };
 
 export async function GET() {
-  const pixels = await getPixels();
-  return Response.json(pixels, { headers });
+  try {
+    const pixels = await getPixels();
+    return Response.json(pixels, { headers });
+  } catch (err) {
+    console.error("Error getting pixels", err);
+    return new Response("Pixels not available", { status: 500 });
+  }
 }
--- a/apps/website/src/hooks/pixels.tsx
+++ b/apps/website/src/hooks/pixels.tsx
@@ -11,9 +11,12 @@ export const PIXEL_GRID_WIDTH = 200;
 export const PIXEL_GRID_HEIGHT = 50;
 export const PIXEL_TOTAL = PIXEL_GRID_WIDTH * PIXEL_GRID_HEIGHT;
 
-export const getPixels = async () => {
+const getPixels = async () => {
   const response = await fetch("/api/pixels");
   const data = await response.json();
+  if (!response.ok) {
+    throw new Error("Failed to get pixels");
+  }
   const pixels = data as Pixel[];
   return pixels.map((pixel) => {
     const location =
--- a/apps/website/.env.example
+++ b/apps/website/.env.example
@@ -113,7 +113,3 @@ CF_STREAM_LOLA_VIDEO_ID=
 
 # TRPC API keys for secure connected systems (optional)
 TRPC_API_SHARED_KEY=
-
-# Donations manager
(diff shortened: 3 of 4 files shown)


apps/website/src/server/trpc/router/donations.ts
```diff
@@ -63,7 +63,7 @@ export const donationsRouter = router({
       return createDonations(input.donations);
     }),
 
-  getPixels: sharedKeyProcedure.query(async () => {
+  getPixels: publicProcedure.query(async () => {
     const pixels = await getPixels();
     return pixels;
```
changesbyjames, line 68: What's the most appropriate way to cache this so it's not hitting the db every time?

apps/website/src/app/api/pixels/route.ts
MattIPv4, line 1: Can we follow the pattern we have in our other routes w/ a try/catch? And I imagine we might want to tell the client to cache this briefly as well?

apps/website/src/hooks/pixels.tsx
```diff
@@ -1,301 +1,31 @@
...
-      {children}
-    </PixelSyncProviderContext.Provider>
-  );
-};
-
-const usePixelSync = () => {
-  const sync = use(PixelSyncProviderContext);
-  if (!sync) throw new Error("PixelSyncProvider not found");
-  return sync;
+export const getPixels = async () => {
```
MattIPv4: Does this need to be exported?

apps/website/src/hooks/pixels.tsx
```diff
@@ -1,301 +1,31 @@
...
+  const response = await fetch("/api/pixels");
```
MattIPv4, line 15: Handle `!response.ok`?

apps/website/src/hooks/pixels.tsx
MattIPv4, line 1: Worth keeping the provider still so we don't make the API call multiple times?

apps/website/src/hooks/pixels.tsx
```diff
@@ -1,301 +1,31 @@
...
-}: {
-  children: ReactNode;
-}) => {
-  const sync = useMemo(() => {
-    if (env.NEXT_PUBLIC_DONATIONS_DEMO_MODE) {
-      return new DemoPixelSyncProvider();
-    }
-    if (!env.NEXT_PUBLIC_DONATIONS_MANAGER_URL) {
-      throw new Error("NEXT_PUBLIC_DONATIONS_MANAGER_URL is not set");
-    }
```
MattIPv4, line 235: Do these env vars want cleaning up from env/index.js, and the .env.example?

apps/website/src/hooks/pixels.tsx
changesbyjames: The `useQuery` will handle that & much more neatly than the websocket!

MattIPv4: Ah missed that this was still wrapped in a `useQuery`

apps/website/src/app/api/pixels/route.ts
changesbyjames: > And I imagine we might want to tell the client to cache this briefly as well?

It's 150ms from Vercel's CDN, I think an extra client cache would just lead to more "my name hasn't updated" issues but I'm happy to add them?

MattIPv4: Ah yeah that's true, this should be fine.

PR #1572, opened 2025-10-24
Fix stream selector bug

### LINKED ISSUE #1571: [Bug]: Stream selector hidden behind video embed

Currently the selector from the "Watch Live on …" button on the home page get covered by the live stream embed on smaller viewports.

Image

Selector should be over the video - or expand to the top?!

### Steps to reproduce

Go to home page, resize window so the live stream embed flows under the button, click the arrow to select another streaming platform. See screenshot above.

### Browser specs

n/a

### Anything else?

_No response_

CtrlNate:
Fix #1571

The platform selector dropdown from the "Watch Live on..." button was being covered by the Twitch embed on smaller viewports due to z-index stacking context issues.

- Changed `MenuItems` to use Headless UI's `anchor="bottom start"` positioning, which uses fixed positioning and portals to render the dropdown outside the normal stacking context.

### DIFF
--- a/apps/website/src/pages/institute/pixels.tsx
+++ b/apps/website/src/pages/institute/pixels.tsx
@@ -3,7 +3,14 @@ import { type NextPage } from "next";
 import Image from "next/image";
 import { useRouter } from "next/router";
 import pluralize from "pluralize";
-import { useCallback, useMemo, useRef, useState } from "react";
+import {
+  type ChangeEvent,
+  useCallback,
+  useEffect,
+  useMemo,
+  useRef,
+  useState,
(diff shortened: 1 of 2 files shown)


MattIPv4 (approved):

MattIPv4:
Tyvm as always. If you want two other really small issues to work on that I haven't filed issues for, both on the pixels page: 1. if you type in the input with the cursor not at the end, it jumps to then 2. on mobile tapping on the canvas shows the tooltip clipped, I think we should fix the z-clipping but also making tapping expand to fullscreen?

PR #1575, opened 2025-10-25
Fix search cursor jumping to end while typing

CtrlNate:
Fixed cursor position being reset on each keystroke by using local state for the input value that syncs with the URL query parameter, instead of deriving the value directly from the router query on every render.

### DIFF
--- a/apps/website/src/pages/institute/pixels.tsx
+++ b/apps/website/src/pages/institute/pixels.tsx
@@ -113,13 +113,16 @@ const InstitutePixelsPage: NextPage = () => {
     });
   }, []);
 
-  const handleScrollbarChange = useCallback((e: ChangeEvent) => {
-    const progress = parseFloat(e.target.value);
-    setScrollProgress(progress);
-    if (!fullscreenRef.current) return;
-    const { scrollWidth, clientWidth } = fullscreenRef.current;
-    fullscreenRef.current.scrollLeft = (scrollWidth - clientWidth) * progress;
-  }, []);
+  const handleScrollbarChange = useCallback(
+    (e: ChangeEvent<HTMLInputElement>) => {
+      const progress = parseFloat(e.target.value);
+      setScrollProgress(progress);
+      if (!fullscreenRef.current) return;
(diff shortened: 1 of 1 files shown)


apps/website/src/pages/institute/pixels.tsx
```diff
@@ -101,16 +113,13 @@ const InstitutePixelsPage: NextPage = () => {
...
-    },
-    [],
-  );
+  const handleScrollbarChange = useCallback((e: ChangeEvent) => {
+    const progress = parseFloat(e.target.value);
+    setScrollProgress(progress);
+    if (!fullscreenRef.current) return;
+    const { scrollWidth, clientWidth } = fullscreenRef.current;
+    fullscreenRef.current.scrollLeft = (scrollWidth - clientWidth) * progress;
+  }, []);
```
MattIPv4: Why the unrelated changes here?

```suggestion
  const handleScrollbarChange = useCallback(
    (e: ChangeEvent<HTMLInputElement>) => {
      const progress = parseFloat(e.target.value);
      setScrollProgress(progress);
      if (!fullscreenRef.current) return;
      const { scrollWidth, clientWidth } = fullscreenRef.current;
      fullscreenRef.current.scrollLeft = (scrollWidth - clientWidth) * progress;
    },
    [],
  );
```

apps/website/src/pages/institute/pixels.tsx
```diff
@@ -290,7 +299,10 @@ const InstitutePixelsPage: NextPage = () => {
                     : "text-alveus-green md:text-alveus-tan",
                 )}
               >
-                {`Found ${filtered.toLocaleString()} ${pluralize("pixel", filtered)}`}
+                {`Found ${filtered.toLocaleString()} ${pluralize(
+                  "pixel",
+                  filtered,
+                )}`}
```
MattIPv4: Why the unrelated changes here?


```suggestion
                {`Found ${filtered.toLocaleString()} ${pluralize("pixel", filtered)}`}
```

PR #1580, opened 2025-10-26
Add workflow to sync private mirror

### LINKED ISSUE #1551: [Idea]: Add workflow to sync private mirror

Add a new workflow that is gated to only run in alveusgg/alveusgg-private. It should pull the latest main from alveusgg/alveusgg, and push it to main in the private repo. It should run on a schedule every hour.

Keep the private mirror in sync w/ the public repo.

CtrlNate:
## Describe your changes

Resolves #1551 - I noticed this didn't have a PR yet, so I took a pass at it since it was a quick change :) Happy to adjust or close this if someone else was planning to handle it!

1. Public → Private Sync Workflow

- Runs hourly (on main) and can be manually dispatched on any branch
- Fetches the corresponding branch from `alveusgg/alveusgg` and directly pushes it to the private repo
- Uses a regular (non-force) push, so if the private repo has diverged unexpectedly, the workflow will fail and alert us

2. Deployment Workflow Gating

- Make prod chatbot deployments only run in `alveusgg/alveusgg`

### DIFF
--- a/.github/workflows/sync-from-public.yml
+++ b/.github/workflows/sync-from-public.yml
@@ -16,9 +16,12 @@ jobs:
     steps:
       - name: Checkout private repo
         uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5.0.0
+        with:
+          ref: ${{ github.ref }}
 
-      - name: Fetch public repo and push to main
+      - name: Fetch public repo and push
         run: |
+          BRANCH="${{ github.ref_name }}"
           git remote add upstream https://github.com/alveusgg/alveusgg.git
-          git fetch upstream main
-          git push origin upstream/main:main
+          git fetch upstream "$BRANCH"
+          git push origin "upstream/$BRANCH:$BRANCH"


.github/workflows/sync-from-public.yml
```diff
@@ -0,0 +1,90 @@
...
+
+permissions:
+  contents: write
+  pull-requests: write
+
+jobs:
+  sync:
+    if: github.repository == 'alveusgg/alveusgg-private'
+    runs-on: ubuntu-latest
+    timeout-minutes: 10
```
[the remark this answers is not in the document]
CtrlNate: I don't think this is necessary for the size of this codebase

MattIPv4:
👀 I am happy for the workflow to directly push to main on the private repo, no need for a new branch + PRs continuously creating noise. We'll always be using feature branches for anything being worked on there.

I think it should do a regular push though, not a force-push, so that if the private main did happen to have something unexpected then it'd fail to push until we rectified it.

CtrlNate:
@MattIPv4 updated to do that! I wasn't sure if that solution would work which is why I did it this way, but the simpler strategy is definitely better.

.github/workflows/sync-from-public.yml
```diff
@@ -0,0 +1,39 @@
...
+    steps:
+      - name: Checkout private repo
+        uses: actions/checkout@v4
```
MattIPv4: Pin the action to the same SHA used in other workflows, please

.github/workflows/sync-from-public.yml
```diff
@@ -0,0 +1,39 @@
...
+        with:
+          fetch-depth: 0
```
MattIPv4: Why does this need the full git history?

.github/workflows/sync-from-public.yml
```diff
@@ -0,0 +1,39 @@
...
+
+      - name: Configure git user
+        run: |
+          git config user.name "github-actions[bot]"
+          git config user.email "41898282+github-actions[bot]@users.noreply.github.com"
```
MattIPv4: Does GitHub not automatically do this? We're not actually committing, so do we even need this?

.github/workflows/sync-from-public.yml
```diff
@@ -0,0 +1,39 @@
...
+          if ! git remote | grep -q '^upstream$'; then
+            git remote add upstream https://github.com/alveusgg/alveusgg.git
+          fi
+          git fetch upstream main
+
+      - name: Merge upstream/main into main and push
+        run: |
+          git checkout main
+          git merge upstream/main --ff-only
+          git push origin main
```
MattIPv4: I'm not sure I want it to do a merge, I think I just want to push upstream/main to main, and if it fails due to something landed in main, that's desired that it'll just error?

.github/workflows/update-chatbot.yml
```diff
@@ -44,7 +44,7 @@ jobs:
     name: Update Chatbot (Preview)
     runs-on: ubuntu-latest
     environment: preview
-    if: github.ref == 'refs/heads/preview'
+    if: github.repository == 'alveusgg/alveusgg' && github.ref == 'refs/heads/preview'
```
MattIPv4: I think the preview we still want to be able to deploy from the private repo if needed, just not production as the public repo main branch handles that.

.github/workflows/sync-from-public.yml
CtrlNate: I believe it was necessary for the old way, but certainly not anymore

.github/workflows/sync-from-public.yml
```diff
@@ -0,0 +1,24 @@
...
+    timeout-minutes: 10
+    steps:
+      - name: Checkout private repo
+        uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5.0.0
+
+      - name: Fetch public repo and push to main
+        run: |
+          git remote add upstream https://github.com/alveusgg/alveusgg.git
+          git fetch upstream main
+          git push origin upstream/main:main
```
MattIPv4: One maybe silly final thought, given this has `workflow_dispatch`, thoughts on using whatever branch GitHub passes in as the branch this will sync? So on the schedule it'd be main, but it can be dispatched against any branch desired to sync?

CtrlNate: Good idea!

PR #1581, opened 2025-10-26
Add preview modal to show and tell admin review page

### LINKED ISSUE #1404: [Idea]: Show and Tell Admin: Open article preview from edit page

Add a button to open article preview from the edit view in the admin panel for show and tell.

Allow previewing easier without having to go back to the list.

CtrlNate:
Resolves #1404 - add a preview button to the Show and Tell admin review page, allowing admins to preview entries (with unsaved edits) directly from the edit view without navigating back to the list.

### New Preview Modal

- Added AdminShowAndTellPreviewModal component that displays the entry as it will appear to users
- Preview shows current form data (title, displayName, text, location) even before saving
- Modal includes action buttons for quick workflows:
    - "Save & Approve" button for pending entries (saves changes and approves in one click)
    - "Save" button for already approved entries (saves changes without affecting approval status)

### Admin Review Page Updates

- Added "Preview" button to the review page that opens the modal
- Implemented form data extraction to preview unsaved edits
- Added save/approve handlers that chain form submission with approval mutation
- Used useRef instead of useState for the approve flag to avoid closure issues

### Form Integration

- Added optional formRef prop to ShowAndTellEntryForm to enable programmatic form submission
- Added onSaveSuccess callback that fires after successful save operations

image


CtrlNate:
I've replaced the preview page in the admin side with a preview modal, imo it is far more clean and less annoying than having to open a whole new page.

Should I also replace the preview button in the user frontend so we can fully remove the `/show-and-tell/my-posts/[id]/preview` page?

apps/website/src/pages/admin/show-and-tell/review/[entryId]/index.tsx
```diff
@@ -131,6 +159,139 @@ const AdminReviewShowAndTellPage: NextPage<
...
+            imageAttachment: {
+              id: `temp-${idx}`,
+              fileStorageObjectId: null,
+              url: file.dataURL,
+              alt: null,
+              caption: null,
+              fileStorageObject: null,
+            },
+          };
+        }
```
Comment (pjeweb) [inline apps/website/src/pages/admin/show-and-tell/review/[entryId]/index.tsx]: I wonder if we should just skip those? Save some LOC…

pjeweb (changes requested):
Amazing work again! Just the one thingy, but happy to merge either way.

CtrlNate:
> Should I also replace the preview button in the user frontend so we can fully remove the `/show-and-tell/my-posts/[id]/preview` page?

@pjeweb did you have any thoughts on this too? I'll fix re: your comment later today!

pjeweb:
> > Should I also replace the preview button in the user frontend so we can fully remove the `/show-and-tell/my-posts/[id]/preview` page?
> 
> @pjeweb did you have any thoughts on this too? I'll fix re: your comment later today!

Oh i didnt see that comment. Yeah I guess that would be good. I assume you could reuse/refactor out a bunch of the preview data bits for that, yeah?

CtrlNate:
@pjeweb Image loading seems to be entirely broken on my local dev. I can't figure out the reason why this may be the case, but I don't think it should be related to any of my code changes here (it happens even on the regular pages which I did not touch). Do you have a similar issue on your end when you test?

That wasn't the case before, maybe something released on main broke stuff on local dev for some reason?

pjeweb:
@ACA30 can you check if it was fixed by #1605 ?

PR #1602, opened 2025-10-29
Added support for live & static murals

changesbyjames:
## Describe your changes

Added support for live & static murals. Also refactored donation worker to use sql storage for better observability.

...

...

### DIFF
--- a/apps/website/next.config.ts
+++ b/apps/website/next.config.ts
@@ -25,6 +25,12 @@ function urlOriginAsRemotePattern(url: string): RemotePattern {
   };
 }
 
+const s3Pattern = urlOriginAsRemotePattern(
+  process.env.FILE_STORAGE_CDN_URL ||
+    process.env.FILE_STORAGE_ENDPOINT ||
+    "https://localhost",
+);
+
 const config: NextConfig = {
   reactStrictMode: true,
   images: {
@@ -76,12 +82,10 @@ const config: NextConfig = {
         hostname: "www.alveussanctuary.org",
       },
       // S3 - File Storage
-      urlOriginAsRemotePattern(
-        process.env.FILE_STORAGE_CDN_URL ||
-          process.env.FILE_STORAGE_ENDPOINT ||
-          "https://localhost",
-      ),
+      s3Pattern,
     ],
+    // Allow localhost requests if the S3 pattern requires it
+    dangerouslyAllowLocalIP: s3Pattern.hostname === "localhost",
     // Allow SVGs to be proxied through Next.js
     dangerouslyAllowSVG: true,
     contentSecurityPolicy: "default-src 'self'; script-src 'none'; sandbox;",
--- a/apps/website/src/app/api/stream/upcoming/[channel]/route.ts
+++ b/apps/website/src/app/api/stream/upcoming/[channel]/route.ts
@@ -1,26 +1,44 @@
 import { getCalendarEvents } from "@/server/db/calendar-events";
 
 import { getFormattedTitle } from "@/data/calendar-events";
-import { channels } from "@/data/twitch";
+import {
+  channels,
+  isChannel,
+  isChannelWithCalendarEvents,
+} from "@/data/twitch";
 
 import { formatDateTimeRelative } from "@/utils/datetime";
 import { DATETIME_ALVEUS_ZONE } from "@/utils/timezone";
 
 // API for chat bot
 // Matches the implementation of src/components/overlay/Event.tsx
-export async function GET() {
+export async function GET(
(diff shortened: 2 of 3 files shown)


apps/website/src/server/trpc/router/donations.ts
```diff
@@ -63,10 +63,12 @@ export const donationsRouter = router({
...
-  getPixels: sharedKeyProcedure.query(async () => {
-    const pixels = await getPixels();
-    return pixels;
-  }),
+  getPixels: sharedKeyProcedure
+    .input(z.object({ muralId: z.string() }))
+    .query(async ({ input }) => {
+      const pixels = await getPixels(input.muralId);
+      return pixels;
+    }),
```
pjeweb: This is such a tiny nit and I am not even sure whats better, but we can save one Promise wrap and some characters by doing this. Not sure why it bothers me ;)

```suggestion
    .query({ input }) => getPixels(input.muralId)),
```

apps/website/src/server/db/donations.ts
```diff
@@ -71,8 +72,10 @@ export async function getPublicDonations({
   });
 }
 
-export async function getPixels() {
-  return prisma.pixel.findMany();
+export async function getPixels(muralId: string) {
+  return prisma.pixel.findMany({
+    where: { muralId: muralId },
+  });
 }
```
pjeweb: Same nit as below. Saving one async keyword ;)

```suggestion
export function getPixels(muralId: string) {
  return prisma.pixel.findMany({
    where: { muralId: muralId },
  });
}
```

apps/website/src/server/trpc/router/donations.ts
pjeweb: Oops missed a ( there

apps/website/src/app/api/pixels/[mural]/route.ts
Comment (MattIPv4) [inline apps/website/src/app/api/pixels/[mural]/route.ts:1]: I wonder if we should have an src/data/murals.ts file with known mural slugs, so we can gracefully 404 here on unknown murals, rather than returning empty arrays for any possible slug?

(We could then also use that same data for nice type-hinting on the provider etc)

apps/website/src/hooks/pixels.tsx
```diff
@@ -11,24 +22,377 @@ export const PIXEL_GRID_WIDTH = 200;
...
+
+    if (!env.NEXT_PUBLIC_DONATIONS_MANAGER_URL) {
+      throw new Error("NEXT_PUBLIC_DONATIONS_MANAGER_URL is not set");
+    }
+
+    // If the mural is sync, we need to connect to the websocket
+    return new PixelSyncProvider({
+      url: `${env.NEXT_PUBLIC_DONATIONS_MANAGER_URL}/pixels/sync`,
+      muralId,
+    });
```
MattIPv4, line 304: Should `PixelSyncProvider` take a `muralId`, that feels misleading to me as it'll only ever return data for whichever mural it considers to be the live one?

apps/website/src/hooks/pixels.tsx
```diff
@@ -11,24 +22,377 @@ export const PIXEL_GRID_WIDTH = 200;
...
     throw new Error("Failed to get pixels");
   }
-  const pixels = data as Pixel[];
+  const pixels = data as PublicPixel[];
   return pixels.map((pixel) => {
     const location =
-      `${pixel.column}:${pixel.row}` as keyof typeof grid.squares;
-    return { ...pixel, data: grid.squares[location] };
+      `${pixel.column}:${pixel.row}` as keyof typeof COMPLETED_GRIDS.v1.squares;
+    return { ...pixel, data: COMPLETED_GRIDS.v1.squares[location] };
```
MattIPv4: I feel like `COMPLETED_GRIDS` should perhaps be part of the src/data/murals.ts file I mentioned in another comment, so that we can use muralId for the lookup here? Currently, this technically accepts a mural id, but is only ever going to return the image data for the first mural?

apps/website/.env.example
```diff
@@ -113,3 +113,4 @@ CF_STREAM_LOLA_VIDEO_ID=
 
 # TRPC API keys for secure connected systems (optional)
 TRPC_API_SHARED_KEY=
+NEXT_PUBLIC_DONATIONS_DEMO_MODE=true
```
MattIPv4:
```suggestion

# Real-time pixels donation feed
#   Generate fake events in development when using the sync provider
NEXT_PUBLIC_DONATIONS_DEMO_MODE=true

```
Adding comments to explain + newline eof

MattIPv4 (approved):

PR #1603, opened 2025-10-29
Support multiple channels in upcoming stream api

pjeweb:
## Describe your changes

This will change the upcoming API to support all configured twitch channels and not just Alveus.

...

### DIFF
--- a/apps/website/src/data/twitch.ts
+++ b/apps/website/src/data/twitch.ts
@@ -85,6 +85,9 @@ export type ChannelWithCalendarEvents = {
     : never;
 }[keyof typeof channels];
 
+export const isChannel = (channel: string): channel is keyof typeof channels =>
+  channel in channels;
+
 export const isChannelWithCalendarEvents = (
   channel: Channel,
 ): channel is ChannelWithCalendarEvents => {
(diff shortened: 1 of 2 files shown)


apps/website/src/data/twitch.ts
```diff
@@ -51,7 +51,7 @@ export const isScopeGroup = (group: string): group is ScopeGroup =>
 interface ChannelData {
   username: string;
   id: string;
-  calendarEventFilter?: (event: CalendarEvent) => boolean;
+  calendarEventFilter: (event: CalendarEvent) => boolean;
```
MattIPv4: `ChannelWithCalendarEvents` + `isChannelWithCalendarEvents` can be cleaned up if we're making this required, or... the route above can make use of `isChannelWithCalendarEvents` to check its alveus/maya and not alveusgg

MattIPv4: Oh actually, we definitely don't want to make this required, if you look at the usages of `isChannelWithCalendarEvents`, this property drives a bunch of logic for whether webcals should be synced/shown in the UI etc.

pjeweb: ah good catch

PR #1534, opened 2025-10-20
Implement a basic admin dashboard

CtrlNate:
I keep seeing the "TODO: Add something useful here" on the admin dashboard page, so I figured I would implement something basic...

Adds a comprehensive admin dashboard with real-time statistics, trend charts, and activity feeds. Definitely let me know what you think would/wouldn't be useful, and I'd be happy to adjust.

image

### Overview Statistics (8 Cards)

- Clickable stat cards that link to their respective admin pages
- Cards: Notifications Sent, Push Subscriptions, Upcoming Events, Show & Tell Posts/Contributors, Total Users, Short Links, and Form Entries
- Auto-refreshes every 60 seconds
- Role-gated: cards only appear if user has the required permission

### Recent Activity Feeds

- **Show & Tell Posts**: 10 most recent posts with approval status
- **Notifications**: 10 most recent notifications sent
- Both feeds are role-gated and hidden if user lacks access

### Trend Charts (Recharts)

- **Donations**: Dual Y-axis chart showing $ amounts (bars) and count (line)
- **User Growth**: New users over time
- **Show & Tell Activity**: Posts submitted over time
- Switchable time ranges: 7d, 14d, 30d, 60d, 90d

## Permission Model

**Base requirement**: All users need `viewDashboard` permission to access `/admin/dashboard`

**Feature-specific permissions**:

- Cards/feeds only appear if user has the relevant permission (e.g., `manageNotifications`, `manageShowAndTell`)
- Superusers bypass all feature-specific checks and see everything
- Trend charts are visible to all dashboard users (no specific permission required)

## Performance Optimizations

- Parallel database queries with `Promise.all()`
- 60-second auto-refresh intervals with `staleTime: 30s`
- `refetchOnWindowFocus: false` to prevent excessive queries

PR #1609, opened 2025-10-31
Improve donations feed

pjeweb:
## Describe your changes

- adds a new role and permission to manage donations (adapted from #1534)
- adds new page layout version for pages under `/admin/` ending in `/popout` for pages with session login but without the normal page layout
- implement theming (dark/light by class name) (adapted from #643)
- allow passing a `onFetch` callback to `useLocalStorage` to get the initial fetched value from local storage
- donation feed:
	- protect with newly added permission
	- show donation notes
	- always show time, but in compact format (now, 5s, 5m …)
	- implement infinite scroll loading
	- implement two versions of "last seen":
		- show a divider above messages since last page view (reload)
		- add a button to mark items as seen manually
	- move filtering of "only pixel" donations to the server

...

### DIFF
--- a/apps/website/src/components/admin/donations/DonationFeed.tsx
+++ b/apps/website/src/components/admin/donations/DonationFeed.tsx
@@ -3,6 +3,7 @@ import { z } from "zod";
 
 import { trpc } from "@/utils/trpc";
 
+import { usePrependScrollLock } from "@/hooks/prepend-scroll-lock";
 import useLocalStorage from "@/hooks/storage";
 
 import { LoadMoreTrigger } from "@/components/LoadMoreTrigger";
@@ -60,6 +61,8 @@ export function DonationFeed() {
     (donation) =>
       initialLastShown !== null && donation.donatedAt <= initialLastShown,
   );
+
+  usePrependScrollLock(donations.length);
 
   const onLoadMore = useCallback(() => {
     if (
--- a/apps/website/src/hooks/prepend-scroll-lock.ts
+++ b/apps/website/src/hooks/prepend-scroll-lock.ts
@@ -1,0 +1,29 @@
+import { useLayoutEffect } from "react";
+
+const DEFAULT_SCROLL_THRESHOLD = 50; // px from top considered "at top"
+
+export function usePrependScrollLock(
+  watchValue: unknown,
+  threshold = DEFAULT_SCROLL_THRESHOLD,
+) {
+  useLayoutEffect(() => {
+    // Determine which element is used for scrolling
+    const el =
+      document.scrollingElement || document.documentElement || document.body;
+
+    const oldTop = el.scrollTop;
+    if (oldTop < threshold) return; // skip if near top
+
+    const oldHeight = el.scrollHeight;
+
+    // Wait for DOM mutations to update
+    const id = requestAnimationFrame(() => {
+      const newHeight = el.scrollHeight;
+      const diff = newHeight - oldHeight;
+
+      // Adjust the scrollTop to compensate if the user is *not at top*
+      el.scrollTop = oldTop + diff;
+    });
+    return () => cancelAnimationFrame(id);
+  }, [watchValue, threshold]);
+}


apps/website/src/pages/_app.tsx
MattIPv4, line 1: Can we simplify what's happening in here further? Can the theme script be included on all pages even if we're not using it? Can the stream pages have the session provider?

pjeweb: I thought about having the SessionProvider in the stream pages, but it might trigger an oauth redirect and stuff, if there even was a session in the OBS browser.

I guess the ThemeScript could be added everywhere. Its just unnecessary client code to run.

apps/website/src/pages/admin/donations/popout.tsx
MattIPv4, line 1: I wonder if the popout layout should handle the auth check, as I imagine that's going to be duplicated across any other popouts?

pjeweb: It does kinda. It has a login button top right. This give the option for popouts without auth.

apps/website/src/components/LoadMoreTrigger.tsx
MattIPv4, line 1: Can we use this on show and tell?

apps/website/src/components/admin/donations/DonationFeed.tsx
```diff
@@ -0,0 +1,139 @@
...
+  );
+  const [initialLastShown, setInitialLastShown] = useState<Date | null>(null);
+  const [, setLastShown] = useLocalStorage(
+    "stream/donation-feed/last-shown",
+    nullableDateSchema,
+    null,
+    (val) => {
+      setInitialLastShown(val);
+    },
+  );
```
MattIPv4, line 40: I'm not sure I like this new onFetch callback, I think a regular useeffect would be better?

```suggestion
  const [initialLastShown, setInitialLastShown] = useState<Date | null>(null);
  const [lastShown, setLastShown] = useLocalStorage(
    "stream/donation-feed/last-shown",
    nullableDateSchema,
    null,
  );
  useEffect(() => {
    setInitialLastShown(prev => prev ?? lastShown);
  }, [lastShown]);
```

apps/website/src/components/LoadMoreTrigger.tsx
pjeweb: We could technically, but my german law experience taught me to never use infinite loading if you have a footer with important stuff ;)

MattIPv4: I thought show and tell did have infinite loading?

apps/website/src/pages/admin/donations/popout.tsx
MattIPv4: Ah I see, could that not be handled by the layout too though? If you don't pass in a permission it doesn't do anything extra, but if you pass a permission it does a full interstitial checking for auth?

apps/website/src/components/admin/donations/DonationFeed.tsx
pjeweb: The thing is, lastShown might be `null` on initial load, then it gets set by the other effect and then this effect would catch. So no, sadly this doesnt work. It might be possible with some ref magic.

apps/website/src/pages/_app.tsx
MattIPv4: Ah yes that's true. Interestingly, we have the lola page under stream which uses auth, so I'm surprised that doesn't need the session provider actually?

apps/website/src/components/LoadMoreTrigger.tsx
pjeweb: Only in full screen and there we do some more magic that I dont really wanna touch with this.

apps/website/src/pages/admin/donations/popout.tsx
pjeweb: Sure we can do that!

apps/website/src/components/admin/donations/DonationFeed.tsx
MattIPv4: Ahh yes I hadn't considered that. 👍 Let's go with what you have then, but this needs a useCallback ;)

apps/website/src/components/LoadMoreTrigger.tsx
MattIPv4: 👍 Rgr, np

apps/website/src/pages/_app.tsx
pjeweb: Ah interesting. Yeah we use server side checks there with `getServerSideProps`. Just as we do with admin pages. We could use that here too … Not sure if thats better.

apps/website/src/pages/admin/donations/popout.tsx
pjeweb: implemented!

apps/website/src/components/admin/donations/DonationFeedItem.tsx
```diff
@@ -0,0 +1,132 @@
...
+        <div className="flex flex-row items-center gap-2 rounded-lg bg-carnival px-2 py-0.5 text-white">
+          <DonationProviderIcon
+            className="size-3 shrink-0"
+            provider={donation.provider}
+          />
+          {formatDonationAmount(donation.amount)}
+        </div>
+        {donation.pixels > 0 && (
+          <div className="rounded-lg bg-alveus-green-900 px-2 py-0.5 text-white">
+            🖼&nbsp;
```
MattIPv4: Weird unicode char here?

pjeweb: yeah i was too lazy to find an icon for "pixels". I think the frame emoji looks good though.

pjeweb: Replaced it with a proper icon!

PR #1612, opened 2025-11-01
Add alert back to pixels overlay

changesbyjames:
## Describe your changes

Added the pixel unlock alert back to the overlay.

...

...

### DIFF
--- a/apps/website/src/components/admin/donations/DonationFeedItem.tsx
+++ b/apps/website/src/components/admin/donations/DonationFeedItem.tsx
@@ -92,8 +92,8 @@ function DonationFeedItem({
       key={donation.id}
       className={classes(
         "border-b border-gray/50 px-2 py-1",
-        isNow && "text-red-800",
-        !isNow && isRecent && "text-red-900",
+        isNow && "text-red-800 dark:text-red-200",
+        !isNow && isRecent && "text-red-900 dark:text-red-300",
       )}
       initial={
         isNow


apps/website/src/components/overlay/PixelsAlert.tsx
MattIPv4, line 1: @pjeweb did you want to adjust the timings on this?

pjeweb: let me see if i can understand it enough to do that lmao. what do you think? // cc @changesbyjames

MattIPv4: Yah I think you'd said you wanted it to be a bit slower for a single pixel unlock? I imagine we want to keep the same speed it currently has for multiple pixels though?

changesbyjames: I'm happy to adjust - I can make it hold a little bit longer on the revealed pixels. Maybe 25% longer on multiple pixels, 50% longer on single?

apps/website/src/components/overlay/PixelsAlert.tsx
```diff
@@ -0,0 +1,303 @@
...
+              { ease: "easeInOut", delay: 0.2 },
+            ],
+          ]),
+          animate(
+            pixelTitleRef.current,
+            { opacity: 1, y: 0 },
+            { ease: "anticipate" },
+          ),
+        ]);
+
```
pjeweb, line 149: Maybe something like this? Just keep each pixel up for like a second:

```ts
import { delay } from "@/utils/delay";
```

```suggestion
        // Keep pixel displayed for a second
        await delay(1000);
```

apps/website/src/components/overlay/PixelsAlert.tsx
pjeweb: Maybe this? https://github.com/alveusgg/alveusgg/pull/1612/files/e2b8ad038924e5f9d14775b5421dac81867e1a85#r2483934509

pjeweb: Oh yeah making it different for one vs multiple pixels makes sense!

pjeweb: Maybe we can cut some of the animation between two pixels of the same donation too. Does the background have to fully disappear and come back?

changesbyjames: It doesn't _have_ to but it does make resetting the reveal a little awkward - I'll tweak and see if there's a better option!

pjeweb (approved):

PR #1622, opened 2025-11-03
Normalize PayPal donation emails

MattIPv4:
## Describe your changes

We can use `SELECT email FROM (SELECT donatedBy->>'$.email' AS email, LOWER(TRIM(donatedBy->>'$.email')) AS email_normalized FROM Donation) AS Donation_email WHERE email <> email_normalized;` to check how many existing donations have emails that need normalizing.

We can then use `UPDATE Donation SET JSON_REPLACE(donatedBy, '$.email', LOWER(TRIM(donatedBy->>'$.email'))) WHERE donatedBy->>'$.email' <> LOWER(TRIM(donatedBy->>'$.email'));` to update all those donations to have normalized emails.

We'll also need to update the metadata stored for any pixels tied to those donations, updating the hashed email to be the hashed normalized email?

...

### DIFF
--- a/apps/donations/worker/src/providers/paypal/paypal.ts
+++ b/apps/donations/worker/src/providers/paypal/paypal.ts
@@ -73,7 +73,7 @@ export class PaypalDonationProvider implements DonationProvider {
       note: payload.data.memo,
       donatedBy: {
         primary: "firstName",
-        email: payload.data.payer_email,
+        email: payload.data.payer_email.trim().toLowerCase(),
         firstName: payload.data.first_name,
         lastName: payload.data.last_name,
       },


pjeweb:
> We can then use `UPDATE Donation SET JSON_REPLACE(donatedBy, '$.email', LOWER(TRIM(donatedBy->>'$.email'))) WHERE donatedBy->>'$.email' <> LOWER(TRIM(donatedBy->>'$.email'));` to update all those donations to have normalized emails.

This was missing a `SET donatedBy = …`.

> We'll also need to update the metadata stored for any pixels tied to those donations, updating the hashed email to be the hashed normalized email?

This should work to normalize:

```sql
UPDATE Pixel p
INNER JOIN Donation d ON p.donationId = d.id
SET
	email = SHA2(LOWER(TRIM(d.donatedBy ->> "$.email")), 256)
WHERE
	NOT p.email IS NULL;
```

Currently there seem to be 219 pixels on mural 1 that arent lower case hashes yet:

```sql
SELECT
	count(p.id),
	p.muralId
FROM
	Pixel p
	INNER JOIN Donation d ON p.donationId = d.id
WHERE
    NOT p.email IS NULL AND
	SHA2(LOWER(TRIM(d.donatedBy ->> "$.email")), 256) <> p.email
GROUP BY
	p.muralId;
```

pjeweb:
Since mural two email hashes should already be lower case (https://github.com/alveusgg/alveusgg/blame/f1e6f44c9cd01f6b474fc694bc4d480475f5e523/apps/donations/worker/src/managers/pixels.ts#L288).

pjeweb:
Also if we merge this we should change #1542  to also lowercase the input and probably remove the capitalization warning there (although it should still apply for the names).

PR #1648, opened 2025-11-14
Clear old marmoset presets and add existing

wazix11:
Cleared all old marmoset presets and the relevant images, added in all the new presets and images.

### DIFF
--- a/apps/website/src/data/tech/network.ts
+++ b/apps/website/src/data/tech/network.ts
@@ -250,8 +250,8 @@ const data: RootNetworkItem[] = [
                   {
                     type: "accessPoint",
                     name: "Pasture",
-                    model: "Ubiquiti UAP-AC-M-Pro",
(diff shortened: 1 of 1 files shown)


MattIPv4 (commented):
Can we update the network diagram while we're here with the new cam models?

wazix11:
Trying to find where the comments were about it, marmoset was Q6318-LE, marmin was P5675-LE and pasture is Q6318-LE now correct?

wazix11:
I'm also not sure on which APs got switched with what

MattIPv4:
> Trying to find where the comments were about it, marmoset was Q6318-LE, marmin was P5675-LE and pasture is Q6318-LE now correct?

Yep!

> I'm also not sure on which APs got switched with what

Pasture + marm are both U7 Pro Outdoor APs now.

Ty :)

wazix11:
I think that was the correct stuff to change and correct links

MattIPv4 (approved):

PR #1673, opened 2025-12-01
Ensure that valid but irrelevant ipns are processed properly

changesbyjames:
## Describe your changes

...

...


apps/donations/worker/src/providers/paypal/paypal.ts
```diff
@@ -56,8 +56,9 @@ export class PaypalDonationProvider implements DonationProvider {
       Object.fromEntries(params.entries()),
     );
     if (!payload.success) {
-      return new Response(`Invalid IPN. Rejected. ${payload.error.message}`, {
-        status: 400,
+      console.warn(`Valid message but ignored. Schema validation failed.`);
+      return new Response(`Valid message but ignored.`, {
+        status: 200,
```
MattIPv4: I would if we should be more intelligent here -- can we detect the IPN type from `params`, and only return a 200 if it's explicitly not a type we want to handle? So that we still return an error to PayPal (which seems to eventually alert us etc.) if they changed the payload for the type we're interested in (or introduce new types)?

apps/donations/worker/src/providers/paypal/paypal.ts
```diff
@@ -56,8 +56,9 @@ export class PaypalDonationProvider implements DonationProvider {
       Object.fromEntries(params.entries()),
     );
     if (!payload.success) {
-      return new Response(`Invalid IPN. Rejected. ${payload.error.message}`, {
-        status: 400,
+      console.warn(`Valid message but ignored. Schema validation failed.`);
```
MattIPv4: Should we log the Zod error here, so we can more easily see in production what is failing? Probably should only be done for actual errors rather than just types we've decided to not support (to my other point)?

changesbyjames: There's a lot of PPI, especially for notifications that we're not handling - I didn't want to log it just in case it got leaked in the errors?

MattIPv4: Ah true. Looking at https://zod.dev/error-formatting it doesn't look like any of the actual values get logged, just the mismatched types?

apps/donations/worker/src/providers/paypal/schema.ts
```diff
@@ -13,25 +13,20 @@ const Tags = z.string().transform((val) => {
...
-    payment_date: z.coerce.date(),
-  });
+  first_name: z.string(),
+  last_name: z.string(),
+  custom: Tags.optional(),
+  business: z.string(),
+  payer_email: z.string(),
+  memo: z.string().optional(),
+  txn_id: z.string(),
+  mc_currency: z.literal("USD"),
```
MattIPv4, line 29: I think this is probably fine, but I do wonder if this might cause unexpected parse failures? Or does PayPal convert everything to USD before sending it to us always?

changesbyjames: The currency is set at the donation link level so in practice everything Paypal sends us will always be in USD. Maybe if someone set money directly to the Paypal address in a different currency? It's an edge case that I'm not sure would be worth having to do the conversion on our side.

MattIPv4: Yeh I'm not worried, and I suppose in that case us throwing a 400 back is the right thing to do 👍

PR #1689, opened 2025-12-03
Replace immer produce with spread

MattIPv4:
## Describe your changes

#1684 made me look at what we were using it for, and it seems to be for a single state mutation call that I think can just make use of a regular spread to create a new object each time.

Closes #1684 

@changesbyjames looking to you for knowledge on why this was done the way it was, and if this will be a problem or not.


changesbyjames:
I've tested this locally and all works as expected. Do you want to merge and I'll deploy it?

PR #1696, opened 2025-12-04
Remove parse5 override for jsdom

MattIPv4:
## Describe your changes

Per https://x.com/tomlienard/status/1996502284591890899, AWS Lambda, and so in turn Vercel, disables default Node.js experimental features. I've opted us back into the default behaviour by setting `NODE_OPTIONS=--experimental-require-module`. With this in place, we should be able to remove the parse5 override for JSDom, moving to the ESM-only version of parse5 instead of the CJS-compatible version.

tRPC routes work. Submitting a show and tell post works.

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -4,9 +4,6 @@ settings:
   autoInstallPeers: true
   excludeLinksFromLockfile: false
 
-overrides:
-  jsdom>parse5: ^7.3.0
-
 patchedDependencies:
   eslint-plugin-tailwindcss@3.17.5:
     hash: e68ba67a3525bc4b23056140334272fb1a90721db49b16680274a371890796d0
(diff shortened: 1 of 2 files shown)


MattIPv4:
Still hitting `ERR_REQUIRE_ESM`, even with the updated Node.js 24.11.0 runtime in Vercel... and still cannot reproduce locally.

```
⨯ Error: require() of ES Module /var/task/node_modules/.pnpm/parse5@8.0.0/node_modules/parse5/dist/index.js from /var/task/node_modules/.pnpm/jsdom@27.2.0/node_modules/jsdom/lib/jsdom/browser/parser/html.js not supported.
Instead change the require of index.js in /var/task/node_modules/.pnpm/jsdom@27.2.0/node_modules/jsdom/lib/jsdom/browser/parser/html.js to a dynamic import() which is available in all CommonJS modules.
    at <unknown> (../../../../opt/rust/nodejs.js:2:13527)
    at Module.Jr (../../../../opt/rust/nodejs.js:2:13905)
    at Ue.e.<computed>.Be._load (../../../../opt/rust/nodejs.js:2:13497) {
  code: 'ERR_REQUIRE_ESM'
}
```

Opened yet another case with their support team.

PR #1698, opened 2025-12-05
Update presets data

AlveusGGBot:
Generated via `pnpm build:presets` in `apps/website`


MattIPv4:
Nice, just need to exclude temp presets from this!

PR #1751, opened 2025-12-29
Add 'throws up' to excluded phrases list for clips

wazix11:
Just adds another phrase to the excluded phrase list for clips.


apps/website/src/pages/stream/clips.tsx
```diff
@@ -104,6 +104,7 @@ export const getStaticProps: GetStaticProps<{
       "vomit",
       "frew up",
       "throw up",
+      "throws up",
       "throwing up",
```
MattIPv4:
```suggestion
      "throw(s|ing)? up",
```
I think we could combine these into a single pattern?

apps/website/src/pages/stream/clips.tsx
```diff
@@ -104,6 +104,7 @@ export const getStaticProps: GetStaticProps<{
       "vomit",
       "frew up",
```
MattIPv4:
```suggestion
      "(th|f)rew up",
```
We could probably also handle "threw up" here?

PR #1753, opened 2025-12-29
Short-circuit Cycle when there are no items

MattIPv4:
## Describe your changes

Avoids `(prev + 1) % 0` which'll be `NaN`, causing it to end up in an infinite loop.

`/stream/overlay?hide=socials&hide=event` no longer locks up after 60s.


pjeweb (approved):
oopsie

PR #1763, opened 2026-01-02
Don't rely on JS running for transparent stream background

MattIPv4:
## Describe your changes

We're seeing a flash of the standard background in OBS before the useEffect kicks in to override the standard background we set on HTML. Instead of doing that, we can rely on modern CSS to look for a class further down in the DOM that can be present _always_ for stream pages without needing JS to run.

Stream pages have no background.


changesbyjames:
First real world &:has usage I've seen, v cool solution!

PR #1807, opened 2026-01-24
Add overlay commands for custom text + refreshing

### LINKED ISSUE #1801: [Idea]: rebuild !showtext !text commands on overlay

not sure where !showtext used to live, but it doesn't exist anymore(?)

since all the other overlay commands have been moved into this overlay, it only seems natural.

MattIPv4:
## Describe your changes

Adds a new `!text` command that will show custom text in the first slot. Can be run without an arg to hide the text. Resolves #1801.

Also adds a new `!refresh overlay` command, complementing the `!refresh extension` command (https://github.com/alveusgg/extension/pull/425), allowing the overlay to be refreshed by itself without needing to control OBS.

Running `!text hello` shows text on the overlay, centered at the bottom of slot 1 with a background. Running `!text` hides the text. Running `!refresh overlay` reloads the overlay with a query param added to cache bust.

### DIFF
--- a/apps/website/src/pages/stream/overlay.tsx
+++ b/apps/website/src/pages/stream/overlay.tsx
@@ -114,8 +114,8 @@ const OverlayPage: NextPage = () => {
     channels,
     useCallback(
       (message: ChatMessage) => {
-        const { text, userInfo } = message;
-        const [command, arg] = text.trim().toLowerCase().split(/\s+/);
+        const { text: raw, userInfo } = message;
+        const [command, arg] = raw.trim().toLowerCase().split(/\s+/);
 
         // Anyone can run the command to toggle the disclaimer
         if (command === "!disclaimer" || command === "!wolftext") {
@@ -152,7 +152,7 @@ const OverlayPage: NextPage = () => {
           // Mods (or trusted users) can run the command to set the text overlay
           if (command === "!text") {
             setText(
-              text
+              raw
                 .trimStart()
                 .slice(command.length + 1)
                 .trim(),


apps/website/src/data/tech/commands.ts
```diff
@@ -1203,24 +1216,17 @@ const commands = {
...
-        type: "string",
-        name: "text",
-        required: true,
-        variadic: false,
-      },
+      [
+        {
+          type: "string",
+          name: "text",
+          required: true,
```
pjeweb: is it required if it can be empty to hide it?

apps/website/src/pages/stream/overlay.tsx
```diff
@@ -138,6 +149,12 @@ const OverlayPage: NextPage = () => {
             return;
           }
 
+          // Mods (or trusted users) can run the command to set the text overlay
+          if (command === "!text") {
+            setText(text.slice(command.length + 1).trim());
```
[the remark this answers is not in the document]
MattIPv4: Ah yes, good point!

apps/website/src/data/tech/commands.ts
MattIPv4: There's a second `[]` arg array below this that is used to describe no string being passed

pjeweb: i see

apps/website/src/pages/stream/overlay.tsx
```diff
@@ -212,12 +229,20 @@ const OverlayPage: NextPage = () => {
...
                 )}
-              />
+              >
+                {disclaimerText}
+              </Text>
+            )}
+
+            {index === 0 && !hide.has("text") && text && (
+              <Text className="absolute inset-x-0 bottom-0 bg-black/50 p-8 pt-6 text-center text-4xl backdrop-blur-md">
+                {text}
```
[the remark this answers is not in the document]
MattIPv4: I figured it was easier to leave the disclaimer/wolftext command as-is, don't want chat/staff to need to remember the correct wording every time. If `!text` gets used, happy for it to overlap if `!disclaimer` happens to have also been run.

apps/website/src/data/tech/commands.ts
MattIPv4: Ah actually looks like we're really inconsistent on this, let me clarify it

PR #1830, opened 2026-01-30
Avoid socials/subs text overlap with PiP layouts

MattIPv4:
## Describe your changes

Adds yet more special logic to handle the disclaimer + custom text positioning when in the pip layouts.

Large disclaimer text + custom text do not collide with socials or subs when in pipbl/pipbr.


pjeweb (approved):
beauty

PR #1835, opened 2026-02-01
Break notification redirects out of iOS PWA browser

### LINKED ISSUE #402: [Idea]: Implement actual pages for updates instead of redirecting

We already have URLs for each update but they redirect to the update URL (usually twitch). To improve the user experience I suggest we show a page instead with the update info and a link/button to go to the URL instead.

TODO:
- [ ] generate nicer URLs for updates (timestamp or slugs based on the notification text) - needs to be unique
- [ ] design and implement actual page (show title, description, image, if applicable the dates and buttons for opening the link, adding to calendar, sharing) - probably can reuse existing components, but needs to "fill" the page
- [ ] change linked url for discord posts
- [ ] use update image as `og:image`

1. Improve user experience on iOS: Right now the push notifications will open the update url in the installed web app (PWA), which is a reduced web view. For twitch links this means you will see the twitch mobile UI without being logged in correctly. The deep link into the native twitch app does not work (always). If possible it should instead show a button to open twitch in a new windows which will hopefully open it in the normal safari or even directly in the native twitch app (using the universal link schema). (ps: check if the universal link schema works with and without for www.twitch.tv).
2. Improve user experience on mobile: Generally I think it might also be common that people get the notification on mobile but want to watch on desktop. So showing the info when clicking the notification might be preferred. (not 100% sure about this one)
3. Improve user experience for discord users for *Announcements*: Instead of linking twitch.tv directly I think it would be preferable to link the details page instead for events that are scheduled in the future etc. This should also allow showing the uploaded update image as an embed (og image). For stream notifications it should probably still link twitch directly!

MattIPv4:
## Describe your changes

Uses the `x-safari-http(s)://` scheme to break all notification redirects out of the iOS PWA experience and into regular Safari. Also introduces special handling for Twitch specifically, allowing those redirects to take the user directly into the Twitch app via the [`twitch://` scheme](https://dev.twitch.tv/docs/mobile-deeplinks/).

I think we can say that this resolves #402, as we already have the dedicated page, and with this change, we're now always sending folks to it (it was already what we sent folks to if they clicked on a recent notification on the updates page).

Clicking on a recent notification (<1 hr) in the iOS PWA with a Twitch stream sends you directly to the Twitch app. Clicking on a recent notification (>1 hr) in the iOS PWA with a Twitch VoD sends you directly to the Twitch app (with the correct VoD timestamp).

Clicking on a push notification (<1 hr) from the iOS PWA with a Twitch stream sends you directly to the Twitch app. Clicking on a push notification (>1 hr) from the iOS PWA with a Twitch VoD sends you directly to the Twitch app (with the correct VoD timestamp).

### DIFF
--- a/apps/website/src/pages/notifications/[notificationId].tsx
+++ b/apps/website/src/pages/notifications/[notificationId].tsx
@@ -61,8 +61,11 @@ const NotificationPage: NextPage<
           return;
         }
 
-        window.location.href = `x-safari-${link.replace(/^(?!(https?:)?\/\/)/, "//")}`;
-        return;
+        // Send http(s) links, or links with no scheme, through x-safari-http(s)
+        if (/^https?:\/\//i.test(link) || /^(?!.+:\/\/)/i.test(link)) {
+          window.location.href = `x-safari-${link.replace(/^(?!https?:\/\/)/i, "https://")}`;
+          return;
+        }
       }
 
       window.location.href = link;


pjeweb (approved):
Yes! I love it.

pjeweb (approved):

PR #1859, opened 2026-02-11
Add optional Redis caching to weather API calls

MattIPv4:
## Describe your changes

We've seen issues with the weather.com API rejecting our requests intermittently. I suspect this is load-related. As such, implementing Redis locking + caching on the underlying API request to better ensure we're only sending 1 request per minute to them.

Without Redis, weather continues to work. With Redis, weather continues to work, and cached data is populated in Redis.


alok0:
We might actaully need some caching.

The overlay queries the weather every minute
https://github.com/alveusgg/alveusgg/blob/f933f8e2439d11223fb9146af0c015ed51a1f3a7/apps/website/src/components/overlay/Weather.tsx#L9-L17 via trpc.

As far as I can tell, there is no caching in this path? So now that we have 2 overlays running, we are doing twice the number of queries

MattIPv4:
Yah, there is no caching there currently, hadn't considered that we're now running two separate overlay instances! Might be time to land this then.

apps/website/src/server/apis/weather.ts
```diff
@@ -74,18 +75,97 @@ export async function getCurrentObservation<T extends boolean>(
...
+      for (let attempt = 0; attempt < 300; attempt += 1) {
+        await new Promise((resolve) => setTimeout(resolve, 100));
+        const data = await cached();
+        if (data) return data;
+      }
+      throw new Error(
+        "Failed to get weather data from cache after waiting for lock",
+      );
+    }
   }
```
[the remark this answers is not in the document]
MattIPv4: Intentional, Redis is optionally made available here, but if it is available it will be relied upon.

apps/website/src/server/apis/weather.ts
```diff
@@ -74,18 +76,94 @@ export async function getCurrentObservation<T extends boolean>(
...
+
+    if (response.status !== 200) {
+      throw new Error(
+        `Error fetching weather data: ${response.status} ${response.statusText}`,
+        { cause: await response.text() },
+      );
+    }
+
+    const json = await response.json();
+    const data = await currentObservationsSchema(imperial).parseAsync(json);
```
alok0: Not a huge deal, but,

if any line throws an error during this query and parsing, nothing is saved to redis, and the lock is immediately released. This would allow for more queries than desired.

MattIPv4: Hm, are you suggesting it might be preferable to always cache the response, even if it is invalid? As to avoid sending excess requests if get invalid responses back?

alok0: Or hold the lock for the full time.

Though always caching the response makes more sense.

pjeweb (approved):

changesbyjames (approved):
Tested this locally w/ oauth and no impacts to that behaviour. Looks good to me!

PR #1861, opened 2026-02-13
Add Sentry to donation worker

flakey5:
## Describe your changes

Re https://discord.com/channels/814038676195115029/977075504760893470/1465871574685974601

Adds Sentry integration to the donation worker following the [docs](https://docs.sentry.io/platforms/javascript/guides/cloudflare/frameworks/hono/) on it.

Currently, this doesn't add source map support which kinda hinders the helpfulness of this. I wanted to hold off on this to see how y'all would prefer this to be done (if at all). At the very least, the actual release id for the worker is set as the release id in Sentry via the CF_VERSION_METADATA` binding.

If the worker is deployed via the actual `deploy` command in the worker's package.json, source map support could be done with how Sentry's wizard ([docs](https://docs.sentry.io/platforms/javascript/guides/cloudflare/#step-3-add-readable-stack-traces-with-source-maps-optional)) sets it up, where it creates some post-deploy commands that create a release and upload the source map files to Sentry. This is how [one of Node.js's workers](https://github.com/nodejs/release-cloudflare-worker) workers does it and it works so far so well ([deploy commands](https://github.com/nodejs/release-cloudflare-worker/blob/9d0d51b652b1017fc0abc163cc84378c0500b6a4/package.json#L15-L16), [part that uploads source maps](https://github.com/nodejs/release-cloudflare-worker/blob/9d0d51b652b1017fc0abc163cc84378c0500b6a4/scripts/upload-sourcemaps.sh)). Though I'm not entirely sure if the release id in Sentry can be the worker's actual release id in the Cloudflare dash with this approach.

- [ ] `SENTRY_DSN` needs to be added as a secret to the worker in the Cloudflare dash

### DIFF
--- a/.github/workflows/update-donations-worker.yml
+++ b/.github/workflows/update-donations-worker.yml
@@ -1,0 +1,64 @@
+name: "Update Donations Worker"
+on:
+  push:
+    branches:
+      - "main"
+      - "preview"
+    paths:
+      - "apps/donations/**"
+      - "package.json"
+      - "pnpm-lock.yaml"
+      - "pnpm-workspace.yaml"
+      - ".github/workflows/deploy.yml"
+
+jobs:
+  deploy:
+    name: Deploy (${{ env.ALVEUS_ENVIROMENT }})
+    runs-on: ubuntu-latest
+    environment: ${{ env.ALVEUS_ENVIRONMENT }}
+    if: github.repository == 'alveusgg/alveusgg`
+    timeout-minutes: 10
+    env:
+      ALVEUS_ENVIRONMENT: ${{ (github.ref == 'refs/heads/main' && 'production') || 'preview' }}
+    steps:
+      - name: Checkout
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+
+      - name: Install pnpm
+        uses: pnpm/action-setup@41ff72655975bd51cab0327fa583b6e92b6d3061 # v4.2.0
+        with:
+          run_install: false
+
+      - name: Install Node.js
+        uses: actions/setup-node@6044e13b5dc448c55e2357c09f80417699197238 # v6.2.0
+        with:
+          node-version-file: .nvmrc
+          cache: pnpm
+
+      - name: Authenticate GitHub Package Registry
+        run: echo "//npm.pkg.github.com/:_authToken=${{ secrets.GITHUB_TOKEN }}" >> ~/.npmrc
+
+      - name: Install dependencies
+        run: pnpm install --frozen-lockfile
+
+      - name: Deploy
+        working-directory: apps/donations/worker
+        env:
+          CLOUDFLARE_API_TOKEN: ${{ secrets.CLOUDFLARE_API_TOKEN }}
+          SENTRY_DSN: ${{ secrets.SENTRY_DSN }}
+        run: pnpm exec wrangler deploy --env=$ALVEUS_ENVIRONMENT --outdir dist --var SENTRY_DSN:$SENTRY_DSN --var SENTRY_RELEASE:${{ github.sha }}
+
+      - name: Create Sentry Release
+        uses: getsentry/action-release@dab6548b3c03c4717878099e43782cf5be654289 # v3.5.0
+        env:
+          SENTRY_AUTH_TOKEN: ${{ secrets.SENTRY_AUTH_TOKEN }}
+          SENTRY_ORG: alveus-sanctuary
+          SENTRY_PROJECT: donations-worker
+        with:
+          ignore_missing: true
+          ignore_empty: true
+          inject: false
+          release: ${{ github.sha }}
+          environment: ${{ env.ALVEUS_ENVIRONMENT }}
+          sourcemaps: apps/donations/worker/dist
+          strip_common_prefix: true
--- a/.github/workflows/update-donations-worker.yml
+++ b/.github/workflows/update-donations-worker.yml
@@ -9,14 +9,18 @@ on:
       - "package.json"
       - "pnpm-lock.yaml"
       - "pnpm-workspace.yaml"
-      - ".github/workflows/deploy.yml"
+      - ".github/workflows/update-donations-worker.yml"
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
 
 jobs:
   deploy:
-    name: Deploy (${{ env.ALVEUS_ENVIROMENT }})
+    name: Deploy (${{ (github.ref == 'refs/heads/main' && 'production') || 'preview' }})
+    environment: ${{ (github.ref == 'refs/heads/main' && 'production') || 'preview' }}
     runs-on: ubuntu-latest
-    environment: ${{ env.ALVEUS_ENVIRONMENT }}
-    if: github.repository == 'alveusgg/alveusgg`
+    if: github.repository == 'alveusgg/alveusgg'
     timeout-minutes: 10
     env:
       ALVEUS_ENVIRONMENT: ${{ (github.ref == 'refs/heads/main' && 'production') || 'preview' }}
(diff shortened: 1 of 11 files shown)


MattIPv4:
👀 @changesbyjames did you ever get anywhere with CI for the worker? Would you be opposed to @flakey5 getting that done as part of this (assuming they're willing to), so that we can manage version + sourcemaps as part of that for Sentry?

changesbyjames:
Hello :) absolutely no problem with that. We decided a while ago it would be best to use Cloudflare's build system instead of Github actions so we can keep it simple and similar to the normal Vercel process. https://developers.cloudflare.com/workers/ci-cd/builds/git-integration/github-integration/

The main issue with trying to set it up, at least in preview first, was a rogue permissions error that we couldn't immediately resolve. 
>Failed common permission check against resources. (Permission group: "Account Settings Read")

I think Matt raised it so there may be a solution now? Let me know if there's anything I can do to help!

flakey5:
> and it works so far so well

I say this then we immediately connect the Sentry package to some issues the next day :sob:. They're not _super_ critical and Sentry is investigating them, but still of note: notably increased cpu time and increased memory usage (so much so that one of Node.js's workers is hitting exceeded memory limit errors)

Re setting CI things up, more than willing to setup something using GitHub actions if y'all would like. I haven't really used Workers Builds before, but afaict it should also still allow uploading the source maps (though one of y'all that has access to your Cloudflare account needs to configure it)

MattIPv4:
👀 I think I'd be fine with sticking to GitHub Actions to do this for now, for simplicity (similar to how the chatbot is deployed). I think the setup from https://github.com/nodejs/release-cloudflare-worker + https://github.com/cdnjs/api-server seems like a fine pattern to follow (deployed to prod on `main`, and deploying to staging/preview on `preview`).

I think it'd also be worth at least checking that the worker type-checks pushes/PRs as part of our regular checks. The existing `types` check should do type-checking once a `types` command is added to the worker package.json.

.github/workflows/pull-request.yml
```diff
@@ -92,6 +92,9 @@ jobs:
       - name: Prepare .env
         run: pnpm -rc exec "cp .env.example .env || true"
 
+      - name: Generate Worker Types
+        run: pnpm -rc exec "pnpm cf-typegen || true"
```
MattIPv4: Can we make this part of the `types` script, similar to what is done for the website with `next typegen`?

apps/donations/worker/src/managers/donations.ts
```diff
@@ -74,3 +75,11 @@ export class DonationsManagerDurableObject extends DurableObject<Env> {
     return this.router.fetch(request);
   }
 }
+
+export const DonationsManagerDurableObject =
+  Sentry.instrumentDurableObjectWithSentry(
+    (env: Env) => ({
+      dsn: env.SENTRY_DSN,
```
MattIPv4: I think we'll want to explicitly set the release here?

apps/donations/worker/src/managers/pixels.ts
```diff
@@ -250,6 +252,14 @@ export class PixelsManagerDurableObject extends DurableObject<Env> {
   }
 }
 
+export const PixelsManagerDurableObject =
+  Sentry.instrumentDurableObjectWithSentry(
+    (env: Env) => ({
+      dsn: env.SENTRY_DSN,
```
MattIPv4: I think we'll want to explicitly set the release here?

apps/donations/worker/src/index.ts
```diff
@@ -24,34 +38,40 @@ app.all("/pixels/*", async (c) => {
...
-    );
-    await api.donations.createDonations.mutate({ donations });
-
-    const manager = env.PIXELS_MANAGER.getByName(`alveus-${env.MURAL_ID}`);
-    await manager.process(donations);
-  },
-} satisfies ExportedHandler<Env, string>;
+export default Sentry.withSentry<Env>(
+  (env) => ({
+    dsn: env.SENTRY_DSN,
```
MattIPv4: I think we'll want to explicitly set the release here?

apps/donations/worker/wrangler.jsonc
```diff
@@ -170,4 +170,9 @@
   },
 
   "placement": { "mode": "smart" },
+
+  // Expose the worker's version for the Sentry SDK
+  "version_metadata": {
+    "binding": "CF_VERSION_METADATA",
+  },
```
MattIPv4: Do we need this given we're going to be setting the release to the commit SHA?

.github/workflows/deploy.yml
```diff
@@ -0,0 +1,63 @@
...
+          CLOUDFLARE_API_TOKEN: ${{ secrets.CLOUDFLARE_API_TOKEN }}
+          SENTRY_DSN: ${{ secrets.SENTRY_DSN }}
+        run: cd apps/donations/worker && pnpm exec wrangler deploy --env=$ALVEUS_ENVIRONMENT --outdir dist --var SENTRY_DSN:$SENTRY_DSN --var SENTRY_RELEASE:${{ github.sha }}
+
+      - name: Create Sentry release for Donations Worker
+        uses: getsentry/action-release@dab6548b3c03c4717878099e43782cf5be654289 # v3.5.0
+        env:
+          SENTRY_AUTH_TOKEN: ${{ secrets.SENTRY_AUTH_TOKEN }}
+          SENTRY_ORG: TODO
+          SENTRY_PROJECT: TODO
```
MattIPv4:
```suggestion
          SENTRY_ORG: alveus-sanctuary
          SENTRY_PROJECT: donations-worker
```

.github/workflows/update-donations-worker.yml
MattIPv4, line 1: Can we name this `update-donations-worker.yml`?

.github/workflows/deploy.yml
```diff
@@ -0,0 +1,63 @@
...
+        run: echo "//npm.pkg.github.com/:_authToken=${{ secrets.GITHUB_TOKEN }}" >> ~/.npmrc
+
+      - name: Install dependencies
+        run: pnpm install --frozen-lockfile
+
+      - name: Deploy Donations Worker
+        env:
+          CLOUDFLARE_API_TOKEN: ${{ secrets.CLOUDFLARE_API_TOKEN }}
+          SENTRY_DSN: ${{ secrets.SENTRY_DSN }}
+        run: cd apps/donations/worker && pnpm exec wrangler deploy --env=$ALVEUS_ENVIRONMENT --outdir dist --var SENTRY_DSN:$SENTRY_DSN --var SENTRY_RELEASE:${{ github.sha }}
```
MattIPv4: Use working-directory rather than cd?

apps/donations/worker/src/managers/pixels.ts
```diff
@@ -112,13 +119,14 @@ export class PixelsManagerDurableObject extends DurableObject<Env> {
 
   private async getStateFromAPI() {
     if (!this.api) throw new Error("API not initialized");
-    const pixels: Omit<Pixel, "data">[] =
+    const pixels: Omit<Pixel, "data" | "muralId">[] =
```
flakey5: this is just to fix the type error, here `muralId` is required but the response from the trpc endpoint has it as optional

apps/donations/worker/src/utils/middleware.ts
```diff
@@ -10,3 +11,18 @@ export const createSharedKeyMiddleware = (sharedKey: string) => {
     return next();
   };
 };
+
+export const setSentryTagsMiddleware = (
+  ctx: Context,
+  next: () => Promise<void>,
+) => {
+  Sentry.setTags({
+    requestId: crypto.randomUUID(),
```
flakey5, line 20: ideally we'd be able to pass this through to the durable objects so there's only one id per request, but I can't find a way to get tags from Sentry that's not a bit hacky

changesbyjames (approved):
@flakey5 I made some changes to fix the queue type issue you were running into and some issues in the workflow file - env isn't available in the top level name & environment properties so I pulled it back to the raw expression. I also added a concurrency rule just to stop any deployments stepping on each others toes!

This is good to go now!

PR #2002, opened 2026-04-17
Update presets data

AlveusGGBot:
Generated via `pnpm build:presets` in `apps/website`.

Check this workflow's logs at https://github.com/alveusgg/alveusgg/actions/runs/24572601320.


pnpm-lock.yaml
pjeweb, line 1: What happened here? Was there a workflow change?

MattIPv4: I think this is new metadata that the latest pnpm release includes, prepping for 11.x. As to why the pnpm update workflow didn't include it, not sure.

pjeweb: Pretty sure its just broken. Its the same structure a normal pnpm has just only pnpm stuff. If you look at the rich diff on gh it removes all the normal deps. Also Vercel's pnpm cant parse it, so its installing from package.json…
Bildschirmfoto 2026-04-17 um 15 55 23

MattIPv4: No, the diff is only adding frontmatter:

Image

But yeah, Vercel not being able to parse it is concerning.

pjeweb: Not sure there is such a concept as a frontmatter in pnpm lockfiles. Its just the same structure as the normal lockfile. And this is what I meant

Image

MattIPv4: Ah I see, I think you're seeing that diff again because of the frontmatter at the top -- it is parsing that as the lockfile rather than the rest of the document.

MattIPv4 (approved):

PR #2006, opened 2026-04-17
Migrate from tsup to tsdown

MattIPv4:
## Describe your changes

Unblocks #1996 as tsup is no longer maintained and is not compatible with TS 6

...


apps/website/sw.tsdown.config.ts
```diff
@@ -0,0 +1,25 @@
...
+import { env } from "./src/env/index.js";
+
+const config = defineConfig((options) => ({
+  entry: { PushServiceWorker: "src/sw/PushServiceWorker.ts" },
+  outDir: "public/push/alveus",
+  dts: false,
+  outputOptions: {
+    codeSplitting: false,
+  },
+  sourcemap: true,
```
pjeweb, line 16:
```suggestion
  sourcemap: true,
  deps: {
    alwaysBundle: [/.*/],
  },
```

We need something like this so its bundling the zod things we need in the service worker.

pjeweb (changes requested):

apps/website/sw.tsdown.config.ts
MattIPv4: Ah yep, had completely assumed setting the target to browser would do that automatically

pjeweb (approved):

PR #2007, opened 2026-04-18
Added sign out & clearer oauth sign in

changesbyjames:
### Sign in
The sign in page, while functional, was designed for a world where you'd only really be on it directly if something went wrong during the process after clicking the sign in button on the homepage. As the census is landing on it directly, I wanted it to be a bit clearer about what was going on and to not look like an error.

Before:
Google Chrome 2026-04-18 14 19 09

After:
Google Chrome 2026-04-18 14 03 34

This pulls the name from the list of clients in the env variables - otherwise the process remains the same. When not in the oauth flow, e.g. `mode=oauth` is missing, it shows the original error messages.

### Sign out
When the user logs out of the census, it forgets their tokens. But then it just redirects back to the alveus site, which is already logged in and signs right back in! This essentially means they can never swap accounts unless they go to the alveus site and explicitly log out. 

This adopts the standard oauth property of `end_session_endpoint` which allows the census to redirect to the site to end a session as prompted by the user. There hasn't been a sign out page in the site before so this introduces one and wires it up to the oauth plumbing.

...

...


pjeweb (approved):
very nice

PR #2054, opened 2026-04-30
Support Neon CRM Donations in Donation Manager and Pixels

pjeweb:
## Describe your changes

This PR adds support for Neon CRM donations to the Alveus donation system. It introduces a new `@alveusgg/neon-crm-api` package that handles API interactions with Neon CRM, integrates a new `NeonDonationProvider` into the donation manager, and enables filtering of Neon donations for the pixel system. The implementation uses webhook-based donation notifications and extracts donor information from Neon CRM accounts.

...

### DIFF
--- a/apps/donations/neon-crm-api/api.ts
+++ b/apps/donations/neon-crm-api/api.ts
@@ -1,0 +1,87 @@
+import type { ZodType } from "zod";
+import type { Options as AllOptions } from "./env.js";
+
+export type Options = Pick<AllOptions, "organizationId" | "apiKey" | "baseUrl">;
+
+type ApiMethod = "GET" | "POST" | "PUT" | "DELETE";
+
+const apiHeaders = ({
+  organizationId,
+  apiKey,
+}: Pick<Options, "organizationId" | "apiKey">) => ({
+  authorization: `Basic ${btoa(`${organizationId}:${apiKey}`)}`,
+  accept: "application/json",
+});
+
+export const url = (strings: TemplateStringsArray, ...values: Array<string>) =>
+  strings.reduce(
+    (acc, str, i) =>
+      acc + str + (i < values.length ? encodeURIComponent(values[i]) : ""),
+    "",
+  );
+
+const fetch = async (
+  options: Options,
+  path: string,
+  method: ApiMethod = "GET",
+  body?: unknown,
+) => {
+  const request: RequestInit & { headers: Record<string, string> } = {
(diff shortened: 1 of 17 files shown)


apps/donations/neon-crm-api/api.ts
pjeweb, line 1: I moved the Neon CRM API stuff into its own module, so we can build with it in the site too

apps/donations/worker/src/providers/index.ts
```diff
@@ -1,7 +1,9 @@
+import type { MiddlewareHandler } from "hono";
 import type { Providers } from "@alveusgg/donations-core";
 
 export abstract class DonationProvider {
   abstract name: Providers;
+  middleware?: MiddlewareHandler = undefined;
```
changesbyjames: Is there something about Neon's request lifecycle that requires middleware? In the other providers the authentication check is done in the main handle function because it just needs to be done for the one request?

apps/donations/worker/src/providers/neon/neon.ts
```diff
@@ -0,0 +1,178 @@
...
+      id: crypto.randomUUID(),
+      provider: "neon",
+      providerUniqueId: `${payload.organizationId}-${data.id}`,
+      providerMetadata: {
+        neonCampaignId: data.campaign.id ?? undefined,
+        neonCampaignName: data.campaign.name ?? undefined,
+        neonFundId: data.fund.id ?? undefined,
+        neonFundName: data.fund.name ?? undefined,
+      },
+      amount: data.amount * 100,
```
changesbyjames: This could give us a weird floating point number - Paypal is string so it's slightly different but it might be worth [handing dollars & cents separately](https://github.com/alveusgg/alveusgg/blob/15bd6bd0ff11bc0e9818ad0015c9765049f33595/apps/donations/worker/src/providers/paypal/paypal.ts#L118)?

apps/donations/worker/src/providers/index.ts
pjeweb: I did it like this cause we can use the standard hono basic auth middleware. Can probably also check auth in the handle function

changesbyjames: If you don't need realm stuff it's probably worth doing it in the handle and then you don't need all the additional optional middleware plumbing - if you do, then this works!

changesbyjames (approved):

PR #2057, opened 2026-05-02
Bugfix Donation Worker / Neon Provider: Fix webhook subscription

pjeweb:
## Describe your changes

This fixes the webhook subscription of the neon donation provider registering a wrong URL.

Webhook now registers correctly


pjeweb:
Waiting on Neon support to see why web hooks are so delayed for us.

pjeweb:
I think this is ready to be reviewed and merged regardless of the intermittent delays. I don't think its something we can fix (in code).

changesbyjames (approved):

PR #2058, opened 2026-05-02
Add Twitch charity campaign linking flow

changesbyjames:
## Summary
- add a user-facing Twitch charity campaign connection flow that requests `channel:read:charity` and stores connected campaign IDs
- add website/server support for fetching the active Twitch charity campaign and keeping per-broadcaster campaign history
- update the donations worker and webhook bootstrap flow so Twitch charity subscriptions can be resubscribed for connected broadcaster accounts

To test - update the environment variables to include
`TWITCH_CHARITY_NAME=Alveus Sanctuary` then log in as alveusgg and go to /charity/twitch`


pjeweb:
This already looks amazing and worked when testing locally!

PR #2069, opened 2026-05-06
Render stream overlay !disclaimer and !text without confliciting

alok0:
## Describe your changes

Render wolftext and text not positioned absolutely so that they don't end up overlapping.

### DIFF
--- a/apps/website/src/pages/stream/overlay.tsx
+++ b/apps/website/src/pages/stream/overlay.tsx
@@ -140,6 +140,68 @@ const grid: Record<Layout, Grid> = {
   },
 };
 
+const SlotOneTextOverlay: React.FC<{
+  showDisclaimer: boolean;
+  showText: boolean;
+  text: string;
+  layout: Layout;
+}> = ({ showDisclaimer, showText, text, layout }) => {
+  const deconflictingPadding = classes(
+    // Don't overlap socials info with text
+    (layout === "fullscreen" ||
+      layout === "pipbr" ||
+      layout.startsWith("pipt")) &&
+      "pl-80",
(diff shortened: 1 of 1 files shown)


apps/website/src/pages/stream/overlay.tsx
```diff
@@ -316,53 +316,58 @@ const OverlayPage: NextPage = () => {
...
+                        // Don't overlap socials info with text
+                        (layout === "fullscreen" ||
+                          layout === "pipbr" ||
+                          layout.startsWith("pipt")) &&
+                          "pl-80",
+                        // Don't overlap sub count with text
+                        (layout === "fullscreen" ||
+                          layout === "pipbl" ||
+                          layout.startsWith("pipt")) &&
+                          "pr-80",
```
MattIPv4: If we are also showing text, we don't need to account for these two overlaps


```suggestion
                        // Don't overlap socials info with text
                        (hide.has("text") || !text) &&
                        (layout === "fullscreen" ||
                          layout === "pipbr" ||
                          layout.startsWith("pipt")) &&
                          "pl-80",
                        // Don't overlap sub count with text
                        (hide.has("text") || !text) &&
                        (layout === "fullscreen" ||
                          layout === "pipbl" ||
                          layout.startsWith("pipt")) &&
                          "pr-80",
```

MattIPv4 (dismissed):
Otherwise, LGTM, ty

alok0:
The logic was starting to get kinda convoluted, I tried to make it clearer what is happening.

MattIPv4 (approved):
Ty again

PR #2077, opened 2026-05-09
"Scroll up header" and "Scroll to top button"

### LINKED ISSUE #1817: [Idea]: "scroll up bar" and "Scroll to top" button

This idea mainly sprung to mind as I was momentarily using the Alveus website on mobile, and while I was scrolling through the list of ambassadors I found myself instinctively trying to scroll up a little bit to make the sites header navigation bar appear (from what I could find this is broadly called a "scroll up bar") which a lot of websites appear to do these days, but quickly realized that's not a feature of the Alveus website.

Secondary to that, I quickly looked down to the bottom right portion of the page view to see if there was a "Scroll to top"-button, but I couldn't find that either. So that's another request.

The "scroll up bar" would make the header navigation bar of the site appear if a user scrolls up just a little bit, giving quick access to the menus and navigation without having to actually scroll all the way to the top of the page. This is useful on long pages. If a user wants the bar to disappear again, they simply scroll down just a little again.

The "scroll to top" button would be useful if a user wants to quickly return to the top of a long page. Especially useful for users on mobile or users who may not be familiar with the "Home" button on their keyboards.

baitrixdev:
## Describe your changes

Closes #1817

- Added a **scroll-up header** on the public site: the main nav hides when you scroll down a bit and comes back when you scroll up, so you’re not stuck scrolling all the way to the top on long pages (matches the behavior described in the issue).
- Added a **“back to top”** control (bottom-right) on public pages, with a smooth scroll animation so it doesn’t just snap to the top.
- On the **homepage**, the header stays **transparent over the hero** and switches to the **same solid green bar as the rest of the site** once you scroll past the **“What is Alveus?”** block, with a normal transition between the two.
- Updated **sticky** bits (e.g. SubNav + bingo play bar) so they sit correctly under the fixed header using the shared header inset.
- **Intentionally left `/admin` alone** — no auto-hide header and no scroll-to-top button there. I didn’t want to touch admin layout/UX or risk breaking anything in the dashboard.

Stream routes and admin popouts were already outside this layout, so unchanged.

- **Public pages (e.g. `/`, `/ambassadors`, any page with SubNav):** scroll down until the header hides, scroll up a little — bar should return. Sticky subnav should sit below the bar when it’s visible.
- **Homepage:** in the hero, header should look like before (transparent on large screens); after “What is Alveus?” / when that section hits the top, bar should go solid green like other pages.
- **Scroll-to-top:** scroll past the threshold, tap the button — page should animate to the top; stays hidden on admin.
- **Mobile:** same scroll-up behavior; open the hamburger while scrolled — header should stay usable (doesn’t hide under you).
- **Admin:** confirm the old static header and no extra FAB.


MattIPv4 (commented):
Thanks for working on this.

My initial reaction is that this is a _lot_ of complexity being introduced for a non-critical feature, and a lot of page-specific logic as well. This hasn't addressed that a similar transparent navbar is used on the institute page as well, nor that every single page currently defines what background the navbar has, which is now redudant due to fixed background you're applying to the navbar itself...

Can this be simplified down massively?

baitrixdev:
You are completely correct looking at this now I must have mentally checked out lmao ill do some fixes

baitrixdev:
Pushed an update that hopefully is okay with you

The fixed shell in Layout is now the only place that decides solid vs transparent on large screens, and the navbar is just lg:bg-transparent so it isn’t fighting that anymore. The old home-only hook is gone in favour of a tiny pathname → anchor id map, and /institute is on there too so it behaves like home instead of being a special case in TS.

I also removed the repeated -mt-40 … bg-alveus-green-900 strip divs across pages. They were only there to paint behind the old transparent desktop bar; with the shell owning the paint they’re redundant (and were half the “every page defines the navbar background” feeling).

There’s still some inherent complexity because of hide-on-scroll and the hero cutover, but the page-specific surface area is basically “add one map entry if you introduce another full-bleed hero” instead of copying hooks or strips.

Happy to change anything else if you want I just personally like this.

baitrixdev:
I was thinking of making it switch to transparent slowly from the specific point like on the home screen at "What is Alveus?"

what it is now it just switches not instantly but pretty quick

something that might look better is while scrolling up it changes so from the point of where is alveus it doesnt just switch but slowly goes transparent while scrolling up.

if thats something you would like me to add let me know I just thought of it now

apps/website/src/components/layout/ScrollToTopButton.tsx
MattIPv4, line 1: I don't think this should be doing animation by itself, just rely on the browser's native smooth scrolling

apps/website/src/data/site-header-hero.ts
MattIPv4, line 1: Not sure I like needing to maintain a map of these pages, feels like a maintenance burden. Why does this need an id, why not just apply the background as soon as the user starts scrolling? Keeping it transparent until it reaches a specific element feels problematic anyway as you'd have other text under the nav?

apps/website/src/hooks/use-site-header.ts
MattIPv4, line 1: The logic in here feels very complex and very hard to follow what it is doing. Can this be simplified at all, and can comments be added explaining the flow of logic?

MattIPv4 (changes requested):
Trying this out on the preview, I notice a few issues:


https://github.com/user-attachments/assets/e547a260-ea93-4b14-85e4-689fafbce347


- It seems very inconsistent when it hides vs. remains sticky. Can we have consistent behaviour, any downward movement causes it to hide, any upward movement brings it back?

- As I alluded to in one of my comments, on the homepage with the transparent background, the nav is unreadable without the background, so I think the background needs to be applied immediately when sticky?


https://github.com/user-attachments/assets/8e8d9784-640e-4568-92a9-d539f216ecc2


- There is an awful flash of light colored background behind the navbar when moving from the homepage to any other page, this needs to be fixed.

- When there is a subnav on the page, there is a very visually unpleasant snapping that happens as the main navbar appears to animate its height/position, but the subnav does not.


Also, where you've removed the nav background color from all pages, you've not removed the associated comment.

Have you tested this with the global promotion, does it account for that?

apps/website/src/components/layout/ScrollToTopButton.tsx
baitrixdev: When testing I personally had the issue that it would instantly snap to the top instead of using smooth scrolling which is why I opted for an animation instead.

apps/website/src/data/site-header-hero.ts
baitrixdev: I do not think I understood your first comment the way you meant it then I thought you did not want the fixed color background in the home which is why I wanted to add the transparent switch

when I tested it to me it looked very ugly the moment I started scrolling it changes to the fixed background but perhaps I should just make that smoother

apps/website/src/hooks/use-site-header.ts
baitrixdev: True I am not very good with commenting things ill make sure to do that from now on
and try to simplify things more in general I normally just do whats in my head and not look at it anymore after it works which is not a good habit ill check the code when I get some time i am sorry for these issues and thank you for taking the time to look over them and tell me these issues

apps/website/src/components/layout/ScrollToTopButton.tsx
MattIPv4: As long as you use https://developer.mozilla.org/en-US/docs/Web/API/Element/scrollTo#syntax w/ smooth it should be fine?

apps/website/src/data/site-header-hero.ts
MattIPv4: Ah I see. I think it should be fine with a transition set on it. Perhaps you could get away with a fixed pixel offset before it does the transition, based on the height of the navbar, but I'm hoping this can be simplified down to just setting the background as soon as scrolling starts.

baitrixdev:
Weird in the videos I see those issues but when I test on my side I do not have any of those issues at all

apps/website/src/data/site-header-hero.ts
baitrixdev: yea I think thats smarter anyway

I was wondering have you thought about a slightly blurred navbar? it wouldnt matter what color is behind it as it would just make that the background

would that be okay? I didnt feel like it would fit the website and would kind of complicate some parts possibly

MattIPv4:
> Weird in the videos I see those issues but when I test on my side I do not have any of those issues at all

I'm testing on macOS with Firefox if that helps.

apps/website/src/data/site-header-hero.ts
MattIPv4: You can set a backdrop blur if you want, but not on the pages where it is currently transparent, though I think I'd prefer keeping the logic as simple as possible and just applying the fixed background color as soon as scrolling happens (or with the offset for the height of the nav itself)

pjeweb:
This is feeling really nice already! I noticed on mobile the scrolling might need some work when you open the menu. It does not scroll the menu, just the page with the menu staying fixed.

baitrixdev:
Hey sorry i received some unfortunate news and was unable to make a comment explaining the changes I have done thank you for this comment I will be working on that as soon as I get some time.

MattIPv4:
👋 Hey @baitrixdev, checking in here if there's anything we can do to help you move this forward?

baitrixdev:
Hey there sorry life has really been putting some pressure on my shoulders
as soon as I get some free time ill move it forward so sorry for the wait

PR #1951, opened 2026-03-26
Replace Giving Block embed with Neon CRM

MattIPv4:
## Describe your changes

Replaces the Giving Block embed on the donate page with a new embed from Neon CRM. Keeps the Giving Block donation option as a regular button link for cryptocurrency, and removes the PayPal link as the Neon CRM form supports PayPal.

Adding Neon to the Pixel Project will be done separately as a follow-up.

## Notes for testing your change

This is a bit of a pain to test locally, see notes added to README. We're not able to allow this to be tested in our Vercel PR previews.

PR #2098, opened 2026-05-15
Prevent layout shift on donate page from Giving Block widget

Maxkrvo:
## Describe your changes

- Reserve min-h-[820px] on the embed column (and its sibling) to stop the Giving Block widget from pushing content down when it loads
- Gate the fallback Giving Block card on loaded from useConsent so it doesn't flash before consent is read from localStorage

...

- [ ] go to /donate with no prior consent: no layout shift when granting consent and the widget loads
- [ ] go to /donate with prior consent: no fallback card flash

...

### DIFF
--- a/apps/website/src/components/content/TheGivingBlockEmbed.tsx
+++ b/apps/website/src/components/content/TheGivingBlockEmbed.tsx
@@ -37,12 +37,14 @@ const TheGivingBlockEmbed = ({
     div.appendChild(script);
 
     return () => {
-      div.innerHTML = "";
+      div.replaceChildren();
     };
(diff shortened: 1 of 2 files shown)


MattIPv4 (commented):
👋 Thanks for working on this. That being said, I'm not sure this is the right move for a few reasons:

1. #1951 will soon land, which removes this embed completely
2. Gating the button behind waiting for consent to load breaks the page for non-JS or slow-connection users
3. Reserving this height creates a maintenance burden, if the embedded content were to change size as we rely on dynamic sizing of it

Maxkrvo:
Fair enough, I wasn't aware of the embed being phased out. Glad to see that change coming in!

Issue #2100, opened 2026-05-15
[Idea]: Replace classes util + RegEx w/ tailwind-merge + clsx


MattIPv4:
### Describe your idea / enhancement

We have a custom conditional `classes` util, which we then combine with a lot of regex pattern testing in components. We should replace the logic in the util w/ https://www.npmjs.com/package/tailwind-merge + https://www.npmjs.com/package/clsx, and then remove the regex testing in components.

### How would this idea / enhancement be beneficial?

Less convoluted custom logic.

pjeweb:
Would maybe use the shadcn version:

```typescript
import { clsx, type ClassValue } from "clsx"
import { twMerge } from "tailwind-merge"
 
export function cn(...inputs: ClassValue[]) {
  return twMerge(clsx(inputs))
}
```

MattIPv4:
Ah, yes, I had forgotten that tailwind-merge doesn't support conditions -- using clsx makes sense 👍

MattIPv4:
#2107

PR #2107, opened 2026-05-17
Replace classes util + RegEx w/ tailwind-merge

Maxkrvo:
#2100
 
## Describe your changes

Swaps the custom classes util for twMerge(clsx(...)) and removes the regex based default suppression in 10 components. Same export name kept to avoid touching 121 import sites. text-stroke and animation-delay registered via extendTailwindMerge so the custom utilities still work well

Click through pages that use the touched components (Box, Section, Button, Heading, Stream, Streamable, YouTube, Progress, ActionButton, overlay Text) like homepage, /about, /show-and-tell, /institute/pixels and confirm nothing shifts visually.


apps/website/src/components/content/Box.tsx
```diff
@@ -35,9 +26,7 @@ const Box = ({ dark, className, ringClassName, children }: BoxProps) => (
         // mirror the rounding from the main className
         (className || "").match(/(?:^| )(rounded(?:-[^ ]+)?)(?: |$)/)?.[1] ||
```
MattIPv4, line 28: This can be removed?

apps/website/src/utils/classes.ts
MattIPv4, line 1: Do we actually need clsx here? Looking at the interface we had defined before (`Array<string | boolean | undefined | null>`), it seems like that is compatible with the interface that `twMerge` accepts?

apps/website/src/utils/classes.ts
```diff
@@ -1,11 +1,24 @@
...
 
-export function classes(
-  ...classes: Array<string | boolean | undefined | null>
-) {
-  return classes.filter(Boolean).join(" ");
+const twMerge = extendTailwindMerge<"text-stroke" | "animation-delay">({
+  extend: {
+    classGroups: {
+      // src/styles/text-stroke.ts: `text-stroke`, `text-stroke-{1..4}`, `text-stroke-{color}`
+      "text-stroke": [{ "text-stroke": ["", (v: string) => /^\d+$/.test(v)] }],
```
MattIPv4, line 16: Should this also have a `fromTheme` to accept any color?

apps/website/src/utils/classes.ts
Maxkrvo: Good catch!

apps/website/src/components/content/Box.tsx
MattIPv4, line 1: I think we're probably fine to keep the logic that mirrors the rounded- className value, to avoid needing to introduce a new prop that needs to be set on every usage?

MattIPv4: 👀 @Maxkrvo checking in on this, I think this is the only thing to address before we can land this?

MattIPv4 (approved):

Issue #2117, opened 2026-05-20
[Idea]: Use Youtube API instead of RSS feed


alok0:
### Describe your idea / enhancement

`fetchYouTubeVideosSafe` returns 

- `id`
- `title`
- `description`
- `author.id`
- `author.name`
- `author.uri`
- `published`

The [playlistItems.list](https://developers.google.com/youtube/v3/docs/playlistItems/list) endpoint can return all of this data in a single GET request

- `id`: [snippet.resourceId.videoId](https://developers.google.com/youtube/v3/docs/playlistItems#snippet.resourceId.videoId)
- `title`: [snippet.title](https://developers.google.com/youtube/v3/docs/playlistItems#snippet.title)
- `description`: [snippet.description](https://developers.google.com/youtube/v3/docs/playlistItems#snippet.description)
- `author.id`: [snippet.channelId](https://developers.google.com/youtube/v3/docs/playlistItems#snippet.channelId)
- `author.name`: [snippet.channelTitle](https://developers.google.com/youtube/v3/docs/playlistItems#snippet.channelTitle)
- `author.uri`: `https://youtube.com/channel/` + channelId
- `published`: [snippet.publishedAt](https://developers.google.com/youtube/v3/docs/playlistItems#snippet.publishedAt)

The only "downside" is that the channel url is not pretty. Getting the pretty url would require querying another endpoint, though for most video we already just use the hardcoded known urls.

### How would this idea / enhancement be beneficial?

Since the RSS endpoint appears to flake out everyday, this should be much more stable.

I've had no problems querying this endpoint continuously for years about once per minute.

`https://www.googleapis.com/youtube/v3/playlistItems?part=snippet&playlistId=<playlistId>&maxResults=50&key=<apiKey>`

example of a single item

```json

    {
      "kind": "youtube#playlistItem",
      "etag": "AIl6Ew__tYK7xR-25ZSQwBC0zLc",
      "id": "VVVMRmZpc2Y2SHhpUXI4XzRtY3ROQm05Y1EuU1NKLS1JOUxpTEU",
      "snippet": {
        "publishedAt": "2026-05-19T22:08:40Z",
        "channelId": "UCfisf6HxiQr8_4mctNBm9cQ",
        "title": "Jacksepticeye Donates $100k to Alveus Sanctuary!",
[...]

PR #2126, opened 2026-05-20
Use Youtube API instead of RSS feed

alok0:
## Describe your changes

https://github.com/alveusgg/alveusgg/issues/2117

I've not tested any of this code

### DIFF
--- a/apps/website/package.json
+++ b/apps/website/package.json
@@ -67,7 +67,6 @@
     "country-list": "^2.4.1",
     "csv-stringify": "^6.7.0",
     "dompurify": "^3.4.3",
-    "fast-xml-parser": "^5.8.0",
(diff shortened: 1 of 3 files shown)


alok0:
Looks good to me. And it looks like all the uses of this are cached so we shouldn't have any api quota limit issues.

MattIPv4:
🤔 I wonder what API key this is using

alok0:
it is the same one used further down in the same file

MattIPv4:
Yeh, I was trying to figure out where that key actually came from. Looks like it is tied to my personal account, will want to swap to an Alveus-owned key before we start using this for more well-trafficked usage.

MattIPv4 (approved):

PR #2134, opened 2026-05-24
Allow swapping Twitch presets embed w/ YouTube

MattIPv4:
## Describe your changes

With the subathon currently happening on Twitch, the livecams are only on YouTube, but can still be controlled from the AlveusGG chat. So, allowing folks to switch the presets page over to using the YouTube livestream instead.

Sidebar width can still be changed, sidebar can be hidden or swapped between Twitch + YouTube.


pjeweb (approved):
cool

PR #2158, opened 2026-06-03
Store Rounds state in localStorage

alok0:
## Describe your changes

Store rounds enabled and checkmarks state in local storage.
This should allow rounds to survive crashes.

Checkmarks reset on `!rounds stop`

...


MattIPv4 (commented):
Hm, I think I would prefer to keep the logic for toggling the component as a whole in the main overlay page -- you could instead just rely on a `useEffect` returning a cleanup to reset the statuses when the component unmounts (I think you could just set this back to an empty object instead of setting values to false)?

alok0:
Relying on mount/unmount to set state is not reliable since it isn't tied to a user action-- this is the same problem that already exists.

The only clean solution that accounts for all edge cases is to handle the reset in the `!rounds` command.

So the other way to solve this would be to promote the rounds checkmark state to the overlay.

MattIPv4:
> Relying on mount/unmount to set state is not reliable since it isn't tied to a user action-- this is the same problem that already exists.

No, the only reason it would unmount would be if the overlay stops rendering it, which would be when the command runs? There is no other reason it would unmount.

alok0:
if the browser source resets, if obs happens to be able to cleanly shutdown the browser source on the way out.

if layout change causes the browser source to change url and reload the overlay.

basically anything that isn't an immediate crash.

and finally if future changes cause the component to get unmounted/remounted for a different reason.

MattIPv4:
> if the browser source resets, if obs happens to be able to cleanly shutdown the browser source on the way out.
>
> if layout change causes the browser source to change url and reload the overlay.
>
> basically anything that isn't an immediate crash.

That's not how React works? A useEffect cleanup will only run when a component is explicitly unmounted.

If you are set on this pattern though, please split the component itself into two parts, there is no reason for all the CSS logic to be evaluating when the component is not actually rendering.

alok0:
I have one alternative solution, but it would require changing `useLocalStorage` to sync multiple instances of itself as a global state rather than a state backup/restore that it currently is.

MattIPv4:
Let's go with this then, but please split the component up into an outer and an inner so we're not evaling the CSS logic when we don't need to 👍

apps/website/src/components/overlay/Rounds.tsx
```diff
@@ -64,12 +66,23 @@ const transformChecks = (
     })
     .filter(isNotNull);
 
-const useChecks = (channels: string[], users?: string[]) => {
+const useRoundsState = (channels: string[], users?: string[]) => {
+  const [roundsEnabled, setRoundsEnabled] = useLocalStorage(
```
MattIPv4:
```suggestion
  const [enabled, setEnabled] = useLocalStorage(
```

apps/website/src/components/overlay/Rounds.tsx
```diff
@@ -64,12 +66,23 @@ const transformChecks = (
...
+    "stream/overlay:rounds-enabled",
+    useMemo(() => z.boolean(), []),
+    false,
+  );
+
   const checks = trpc.stream.getRoundsChecks.useQuery(undefined, {
+    enabled: roundsEnabled,
```
MattIPv4:
```suggestion
    enabled,
```

apps/website/src/components/overlay/Rounds.tsx
```diff
@@ -106,15 +119,33 @@ const useChecks = (channels: string[], users?: string[]) => {
...
+
+          if (command === "!rounds") {
+            if (keys[0] === "off" || keys[0] === "stop") {
+              // reset checkmarks when rounds stop
+              setStatuses((prev) =>
+                typeSafeObjectFromEntries(
+                  typeSafeObjectEntries(prev).map(([key]) => [key, false]),
+                ),
+              );
+              setRoundsEnabled(false);
```
MattIPv4:
```suggestion
              setEnabled(false);
```

apps/website/src/components/overlay/Rounds.tsx
```diff
@@ -106,15 +119,33 @@ const useChecks = (channels: string[], users?: string[]) => {
...
+            } else {
+              setRoundsEnabled(true);
```
MattIPv4:
```suggestion
              setEnabled(true);
```

apps/website/src/components/overlay/Rounds.tsx
```diff
@@ -106,15 +119,33 @@ const useChecks = (channels: string[], users?: string[]) => {
...
+            }
+            return;
+          }
         }
       },
-      [users],
+      [setRoundsEnabled, setStatuses, users],
```
MattIPv4:
```suggestion
      [setEnabled, setStatuses, users],
```

apps/website/src/components/overlay/Rounds.tsx
```diff
@@ -106,15 +119,33 @@ const useChecks = (channels: string[], users?: string[]) => {
...
     ),
   );
 
   return useMemo(
-    () => transformChecks(checks.data ?? [], statuses),
-    [checks.data, statuses],
+    () => ({
+      checks: transformChecks(checks.data ?? [], statuses),
+      roundsEnabled,
```
MattIPv4:
```suggestion
      enabled,
```

apps/website/src/components/overlay/Rounds.tsx
```diff
@@ -106,15 +119,33 @@ const useChecks = (channels: string[], users?: string[]) => {
...
+    }),
+    [checks.data, roundsEnabled, statuses],
```
MattIPv4:
```suggestion
    [checks.data, enabled, statuses],
```

apps/website/src/components/overlay/Rounds.tsx
```diff
@@ -283,4 +317,21 @@ const Rounds = ({
...
 
+const Rounds = ({
+  channels,
+  users,
+}: {
+  channels: string[];
+  users?: string[];
+}) => {
+  // Get the checks with chat command controls
+  const { roundsEnabled, checks } = useRoundsState(channels, users);
```
MattIPv4:
```suggestion
  const { enabled, checks } = useRoundsState(channels, users);
```

apps/website/src/components/overlay/Rounds.tsx
```diff
@@ -283,4 +317,21 @@ const Rounds = ({
...
+
+  if (!roundsEnabled) {
```
MattIPv4:
```suggestion
  if (!enabled) {
```

MattIPv4 (commented):
Nit on naming

apps/website/src/components/overlay/Rounds.tsx
```diff
@@ -106,15 +119,33 @@ const useChecks = (channels: string[], users?: string[]) => {
...
           }
+
+          if (command === "!rounds") {
+            if (keys[0] === "off" || keys[0] === "stop") {
+              // reset checkmarks when rounds stop
+              setStatuses((prev) =>
+                typeSafeObjectFromEntries(
+                  typeSafeObjectEntries(prev).map(([key]) => [key, false]),
+                ),
+              );
```
MattIPv4:
```suggestion
              setStatuses({});
```
Can't we just reset this to an empty object? Is there any reason we'd want to persist the checks as `false` (that's the same as them not being present, isn't it)?

alok0: The trpc query is long lived if it isn't being unmounted, may not necessarily retrigger the useEffect when needed.

MattIPv4: Not sure I follow?

alok0: Nevermind, you're right it is probably fine.

MattIPv4 (commented):
Hmm, this doesn't seem to work in my testing? I start rounds, check an ambassador, reload the page, and the check is cleared? Feels like a race somewhere?

alok0:
There could be a race for the useEffect and useLocalStorage, i think the way useLocalStorage currently works, the initial value is always there in the first render. So the useEffect could be triggering a set state before the useLocalStorage settles.

Can the useEffect be removed, this is what confused me before, because I think transformChecks does most of the heavy lifting.

alok0:
Oops, made a mistake

alok0:
Ok I think I've removed the useEffect, if that is what was causing the race.

MattIPv4 (approved):

PR #2170, opened 2026-06-05
Treat community map locations as plain-text

MattIPv4:
## Describe your changes

We were previously rendering the location from the DB as HTML on the map page. This allowed for a potential stored XSS if a moderator approved a post with HTML in the location (this would be unlikely as it renders in plain text in preview, so any HTML would be very obvious there).

Any HTML included in the location DB payload is rendered in plain text.


MattIPv4:
image

Confirmed this is no longer vulnerable.

pjeweb (approved):

PR #2185, opened 2026-06-10
Fix calender event title wrapping with long words

alok0:
## Describe your changes

fix this text wrapping
screenshot showing overflowed text

...

...


MattIPv4 (approved):
Joe should get a shorter channel name

Issue #64, opened 2023-02-14
Images of the Dono Trees


pjeweb:
Flip said he could make pictures of the dono trees, so maybe we can get him to. With those images I would love to make them easily zoomable. Might need some stitching together first. Maybe we can even OCR them and make them text searchable.

MattIPv4:
If we get decent pictures, would definitely make sense to stitch them and get them as seamless as possible, but probably then re-split them to make loading a bit more efficient on the site.

Major +1 to doing some OCR here, so we have a pre-computed list of names and locations, so a user can search for their name and see it highlighted on the page etc.

MattIPv4:
We have high-res images of the trees from Alveus at last:

- https://files.alveus.site/dono-trees/1.jpg
- https://files.alveus.site/dono-trees/2.jpg
- https://files.alveus.site/dono-trees/3.jpg
- https://files.alveus.site/dono-trees/4.jpg
- https://files.alveus.site/dono-trees/5.jpg
- https://files.alveus.site/dono-trees/6.jpg

We've also been given a list of all the names:

- https://files.alveus.site/dono-trees/names.txt

I think the idea now would be to run some OCR against them one time to get all the names and their locations, and then use that data alongside the images themselves to provide a search experience on the site. When it comes to doing the OCR, we'll probably want to have some smart logic that makes use of the actual list of names + levenshtien distances or similar, as the OCR may not be perfect.

ExperimentalCyborg:
Unfiltered first attempt (ocr only)

[6.jpg.json](https://github.com/user-attachments/files/18133079/6.jpg.json)
[5.jpg.json](https://github.com/user-attachments/files/18133080/5.jpg.json)
[4.jpg.json](https://github.com/user-attachments/files/18133081/4.jpg.json)
[3.jpg.json](https://github.com/user-attachments/files/18133082/3.jpg.json)
[2.jpg.json](https://github.com/user-attachments/files/18133083/2.jpg.json)
[1.jpg.json](https://github.com/user-attachments/files/18133084/1.jpg.json)

ExperimentalCyborg:
Quick script i used to generate these, in case someone wants to replicate or provide more parameters to readtext to get a cleaner result

```py
import os.path
import json
import easyocr


class TextDetector:
    def __init__(self, languages=None):
        if languages is None:
            languages = ['en']
        self.reader = easyocr.Reader(languages)

    def _parse_result(self, raw: list[tuple]) -> list[dict]:
[...]

PR #2186, opened 2026-06-11
Add donor trees for Fund-a-thon with name search

kimen:
## Describe your changes

This is the implementation of a page that allows viewing the donor trees, resolving #64.
It splits the images into smaller chunks to allow loading on slow internet connections, allows username lookup, zoom, drag to move.
It has been built re-using the json files that were in the issue to locate the leafs and have their name.
There are 2 names that are in names.txt that are not in the json files, but I also didn't see any leaf missing a tag:

- kzrl
- mattyBFAPSss

I'm also very aware that the trees have been moved, and are not in the same order anymore, nor are they on the same background, but I think it's fine to still use the old pictures, assuming the leafs themselves haven't been moved around.


MattIPv4:
👀 Thanks for working on this. A few initial thoughts:

1. I do think we'll want to update the images to the new location for the trees before we put this live, they look much better now!

2. I'm not sure the tiling is really necessary, it adds a _lot_ of additional checked-in files + data into the repo. I think we can just let the user wait for the full res image to load?

3. Can we make sure the images are placed in our regular assets folder and imported, rather than being stored as static public assets, so that they go through Next.js' image optimization? You won't need thumb + full res versions then, Next.js will do that for you if you set appropriate sizes on image tags.

4. Do we need to roll our own pan/zoom logic? Could we perhaps rely on xyflow instead, which we use in a few other places in the app (with nodes being leaf names)?

kimen:
> 👀 Thanks for working on this. A few initial thoughts:
> 
> 1. I do think we'll want to update the images to the new location for the trees before we put this live, they look much better now!
> 2. I'm not sure the tiling is really necessary, it adds a _lot_ of additional checked-in files + data into the repo. I think we can just let the user wait for the full res image to load?
> 3. Can we make sure the images are placed in our regular assets folder and imported, rather than being stored as static public assets, so that they go through Next.js' image optimization? You won't need thumb + full res versions then, Next.js will do that for you if you set appropriate sizes on image tags.
> 4. Do we need to roll our own pan/zoom logic? Could we perhaps rely on xyflow instead, which we use in a few other places in the app (with nodes being leaf names)?

1. Yeah that makes sense, the page logic would be the same, we just would need the images, and re-make the OCR'd json files.
(2. 3. 4. 5.)  I've remade it with xyflow, without tiling, using the asset folder, and it's loading a next.js image for the zoomed out version and then it loads the full webp when zoomed in enough

Note: I've forced pushed to rewrite the commit history, to not have the tiles there.

apps/website/src/components/content/DonorTreePanel.tsx
```diff
@@ -0,0 +1,271 @@
...
+
+const TreeImageNode = ({ data }: NodeProps<Node<ImageData>>) => {
+  // Boolean selector so this only re-renders when crossing the threshold, not
+  // on every pan/zoom frame. The panZoom guard matters: for the first frames
+  // after mount the store viewport is a hardcoded zoom 1 (above
+  // FULL_RES_ZOOM), which would fire the multi-MB full-res request on every
+  // mount before the initial fit applies.
+  const showFullRes = useStore(
+    (s) => s.panZoom !== null && s.transform[2] >= FULL_RES_ZOOM,
+  );
```
MattIPv4: Once we show full res once, we might as well always show full res?

apps/website/src/components/content/DonorTreePanel.tsx
```diff
@@ -0,0 +1,271 @@
...
+      {/* Full-res — fetched on demand, overlays thumbnail when zoomed in */}
+      {showFullRes && (
+        // eslint-disable-next-line @next/next/no-img-element
+        <img
+          src={data.image.src}
+          alt=""
+          draggable={false}
+          className="absolute inset-0 size-full"
+        />
+      )}
```
MattIPv4: Can we avoid using a raw `img` here, you should be able to pass through `Image` still, setting `quality` to either `90` or `100`?

apps/website/src/components/content/DonorTreePanel.tsx
```diff
@@ -0,0 +1,271 @@
...
+        selectable: false,
+        focusable: false,
+      };
+    });
+
+    return [imageNode, ...annNodes];
+  }, [tree, activeAnnotation]);
+
+  return (
+    <div className="relative aspect-video w-full overflow-hidden rounded-2xl border border-alveus-green-300/20 shadow-lg">
```
MattIPv4: Can we set a solid background on this?

apps/website/src/components/content/DonorTreeSearch.tsx
```diff
@@ -0,0 +1,77 @@
...
+    <Combobox
+      value={selected}
+      onChange={(value) => {
+        setSelected(value);
+        if (value) onSelect(value);
+      }}
+    >
+      <Label className="mb-1 block text-sm font-medium text-alveus-tan">
+        Search for your name
+      </Label>
```
MattIPv4: I think this could probably be `sr-only`, given the placeholder on the input?

apps/website/src/components/content/DonorTreeSearch.tsx
```diff
@@ -0,0 +1,77 @@
...
+              <li className="px-4 py-2 text-sm text-white/60">
+                No results found
+              </li>
+            ) : (
+              filtered.map((name) => (
+                <ComboboxOption
+                  key={name}
+                  value={name}
+                  as="li"
+                  className="cursor-pointer px-4 py-2 text-sm text-white data-focus:bg-alveus-green-700"
```
MattIPv4: Nit, this focus color is exactly the same as the hero background, which I'm not sure is great?

apps/website/src/data/donor-trees.ts
MattIPv4, line 1: Instead of using an `as` type-assertion, can we rely on Zod to properly type-guard these JSON imports, ensuring the data is as we expect?

apps/website/src/pages/about/events/donor-trees.tsx
MattIPv4, line 1: Would it make sense to use our existing `Carousel` component here, instead of a custom implementation?

apps/website/src/components/content/DonorTreePanel.tsx
MattIPv4, line 1: Thoughts on whether it'd make sense to constrain the xyflow viewport to the image? Currently, you can zoom out + drag beyond the image.

MattIPv4 (commented):
@kimen thanks for the updates, this is looking much better!

MattIPv4:
👋 Hey @kimen, checking in here if you have the time to resolve these remaining comments? I'm not sure when we'll get updated photos from Alveus, so I'm thinking we might be better to merge this once it is ready and update it with newer photos later on.

apps/website/src/components/content/DonorTreePanel.tsx
kimen: I've switched to using Image, but using `quality` makes Next shrink the resolution of the image making it unreadable when zoomed in. So instead I use `unoptimized`

apps/website/src/pages/about/events/donor-trees.tsx
kimen: The Carousel component doesn't work here because it wants to load all the panels at once, instead of doing just in time loading

kimen:
> 👋 Hey @kimen, checking in here if you have the time to resolve these remaining comments? I'm not sure when we'll get updated photos from Alveus, so I'm thinking we might be better to merge this once it is ready and update it with newer photos later on.

Yeah I saw that taking the photos was more involved than Jacob initially thought.
I've updated the page to have the tree a bit bigger, I've added the "fullscreen" option, and addressed your comments either in commits, or as an answer.

apps/website/src/pages/about/events/donor-trees.tsx
MattIPv4: Hm, how so? Why would that be a problem? Wouldn't we just make sure the images are marked as lazy?

kimen: It's not really the image loading itself, but all the annotations done on the tree, and their tooltip hooks.
That in itself is just a slight performance drag, but the bigger issue is actually that the Carousel implements drag, which would conflict with the xyflow drag. So we would need to disable the drag of the Carousel.

apps/website/src/components/content/DonorTreePanel.tsx
```diff
@@ -0,0 +1,339 @@
...
+// after the panel resizes (window resize, phone rotation). Must render inside
+// <ReactFlow> to read the store.
+const PanelMinZoom = ({
+  image,
+  flow,
+  onChange,
+  active,
+}: {
+  image: DonorTree["image"];
+  flow: ReactFlowInstance | null;
```
MattIPv4: Is there not a hook that gives access to this, rather than needing it as a prop?

apps/website/src/components/content/DonorTreePanel.tsx
```diff
@@ -0,0 +1,339 @@
...
+        tree.image.height * (1 + EXTENT_PADDING),
+      ],
+    ],
+    [tree.image.width, tree.image.height],
+  );
+
+  return (
+    <div
+      className={classes(
+        "relative overflow-hidden bg-alveus-green-900",
```
MattIPv4:
```suggestion
        "relative overflow-hidden bg-alveus-tan",
```
I think a lighter bakground might look nicer?

apps/website/src/components/content/DonorTreePanel.tsx
```diff
@@ -0,0 +1,339 @@
...
+  const translateExtent = useMemo<CoordinateExtent>(
+    () => [
+      [-tree.image.width * EXTENT_PADDING, -tree.image.height * EXTENT_PADDING],
+      [
+        tree.image.width * (1 + EXTENT_PADDING),
+        tree.image.height * (1 + EXTENT_PADDING),
+      ],
+    ],
+    [tree.image.width, tree.image.height],
+  );
```
MattIPv4, line 317: Something doesn't seem right here? At the minimum zoom, I can't move the image to the edges of the viewport, and then once I zoom in, I can move the image entirely out of the viewport? I think ideally we'd want to ensure the image edges can never leave the viewport edges?

https://github.com/user-attachments/assets/5e97d1a3-2e27-47ce-b98b-9c67942db7ee

apps/website/src/pages/about/events/donor-trees.tsx
MattIPv4: Ah okay, yeh that makes sense. Might be worth looking at how difficult that'd be, as it would be nice to have one consistent UI for our carousels, but if it seems too complex then don't worry.

kimen: I've implemented a version with the Carousel, modifying the carousel to have a prop to make it draggable or not

apps/website/src/components/content/DonorTreePanel.tsx
kimen: I've changed it so that the image always is on the viewport

MattIPv4 (dismissed):
Ty for the work on this 💚

MattIPv4 (approved):

PR #2190, opened 2026-06-13
Ensure preset cards show on map in zen mode

### LINKED ISSUE #2167: [Bug]: Camera Presets - Tooltip on the Map View are not visible in Zen Mode
### Zen Mode Disabled
Image

### Zen Mode enabled
Image

To see the Preview regardless of the Zen Mode State, because its linked to camera control.

### Steps to reproduce

1. Go to https://www.alveussanctuary.org/about/tech/presets
2. Select a Camera location 
3. Switch to "map mode"
4. Hover over a Camera
5. Toggle Zen Mode and repeat 4.

### Browser specs

https://www.whatsmybrowser.org/b/YKB87

### Anything else?

Also tested in https://www.whatsmybrowser.org/b/GDORF, same result

Moolash3:
Removes the z-index in PresetMapNode and adds z-index of 110 to CopyClipboardButton and RunCommandButton to ensure that the preset cards show on the map in zen mode and all tooltips show over the card.

resolves #2167


MattIPv4 (changes requested):
👀 Ty for this. I don't think we need to pass zen mode down, we can just change the z-index always.

That being said, the z-index of the run command tooltip needs to be higher than the preset card always, to handle this situation:

Image

So we probably just want to remove the z-90 for the preset card, and then set the run command tooltip to z-110?

Moolash3:
> 👀 Ty for this. I don't think we need to pass zen mode down, we can just change the z-index always.
> 
> That being said, the z-index of the run command tooltip needs to be higher than the preset card always, to handle this situation:
> Image
> 
> So we probably just want to remove the z-90 for the preset card, and then set the run command tooltip to z-110?

Sounds good. Do you also want me to implement the copilot suggestions?

apps/website/src/components/ptz/PresetMap.tsx
```diff
@@ -58,7 +58,7 @@ const PresetMapNode = ({ data }: NodeProps<Node<PresetMapNodeData>>) => {
       </PresetCard>
     ),
     aria: `${data.name}: ${data.preset.description}`,
-    className: "z-90",
+    className: "z-110",
```
MattIPv4: We don't need to set a z-index here, we can let it use the default of z-100?

apps/website/src/components/shared/actions/CopyToClipboardButton.tsx
```diff
@@ -45,6 +45,7 @@ const CopyToClipboardButton = ({
         content,
         aria: statusText,
         force: status !== undefined,
+        className: "z-111",
```
MattIPv4: I think this can just be z-110?

MattIPv4: @Moolash3 looks like this got lost?

Moolash3: Its not needed as the default in tooltip is 100. So everything is 100 now that we dont send 90 in with the preset.

MattIPv4: Right, but these are now both going to be at z-100. We want to guarantee that the run/copy tooltip sits above the preset card tooltip

Moolash3: Sounds good

Moolash3:
@MattIPv4 Do you all rebase or merge commit?

MattIPv4:
image

Hm, looks like there's another tooltip that needs the z-110 applied, the run command tooltip is sitting behind the preset card.

> @MattIPv4 Do you all rebase or merge commit?

I generally prefer rebasing, but either works.

MattIPv4:
@Moolash3 the stacking issue hasn't been fixed still: https://github.com/alveusgg/alveusgg/pull/2190#issuecomment-4707048802

I think you want to be setting a higher z-index on RunCommandButton, not the ClipboardButton?

Moolash3:
> @Moolash3 the stacking issue hasn't been fixed still: [#2190 (comment)](https://github.com/alveusgg/alveusgg/pull/2190#issuecomment-4707048802)
> 
> I think you want to be setting a higher z-index on RunCommandButton, not the ClipboardButton?

Yeah you are right, I totally missed the first part of the message and just read the rebase part. My bad!

MattIPv4 (approved):
Ty!

PR #2197, opened 2026-06-16
Add Neon donation sync safeguard

pjeweb:
## Describe your changes

- Add an hourly Neon donation sync that searches for recent successful, non-self-import donations and queues unseen donations through the existing worker flow
- Add Neon CRM API helpers for donation search, direct donation fetches, and parse-error reporting
- Harden Neon webhook setup so startup failures do not create duplicate webhooks, and duplicate matching webhooks are pruned safely

The sync runs from the donations worker scheduled handler and routes through the existing `DonationsManagerDurableObject`, so dedupe and downstream pixel processing use the same path as live webhooks.

The Neon sync intentionally imports all successful non-self-import donations, not only pixel campaign donations. Pixel eligibility remains handled by the existing pixel worker rules.

- Temporarily disable or ignore the Neon webhook path, then create a small test donation in Neon that is `Succeeded` and not `Self-Import` (for example, through a normal donation form/admin donation flow)
- Trigger the sync manually with:
  `POST https://donation-manager.preview.workers.alveussanctuary.org/donations/neon/sync`
  using `Authorization: ApiKey <SHARED_KEY>`
- Confirm the response reports the donation under `imported`, and a second sync reports it as skipped/deduped
- Confirm the donation appears in the website DB with `provider = "neon"` and `providerUniqueId = "alveussanctuary-<donation id>"`
- For pixel campaign/fund donations, confirm pixels are created by the existing pixel worker; for non-pixel donations, confirm no pixels are created


apps/donations/neon-crm-api/schema.ts
```diff
@@ -150,6 +150,15 @@ const CustomFieldData = z.union([
   IdNamePairBase.extend({ value: z.string() }),
 ]);
 
+const normalizeStatus = (status: string) =>
+  status.toUpperCase().replace(/[^A-Z]/g, "_");
```
MattIPv4: Why not define this as a custom schema like everything else?

MattIPv4 (commented):
This looks good, but definitely would appreciate @changesbyjames reviewing this

changesbyjames (dismissed):

pjeweb:
@changesbyjames sorry if you could re-review after merge and the one fix for the status schema please

MattIPv4 (approved):

PR #2221, opened 2026-06-25
Update ptzlist command + add scenecams command

### LINKED ISSUE #2207: Rewording !ptzlist command description
Since !ptzlist no longer returns the list of cameras on stream anymore, the "or get the cameras on stream currently" should be removed from the commands description under the Live Cams Control section and on the all commands page. 
Could instead add the !scenecams command and add a description such as "Get names and slots of all cameras on stream currently"

MillerSpil:
## Describe your changes
Updates the documented description of `!ptzlist` and adds the `!scenecams` command, per #2207.

**What `!ptzlist` does:**
Given a camera, it returns that camera's saved preset positions.
`!ptzlist bb` -> `PTZ Presets: cave, down, downleft, home, leftcorner`

The old description also claimed it could "get the cameras on stream currently," but `!ptzlist` with no argument no longer returns anything (Verified live that this behavior is no longer functional). Removing that clause keeps the description accurate to what the command actually does.

**`!scenecams` (new):**
This is the command that now reliably reports the cameras currently on stream with their slot numbers. It's now documented as the proper replacement for the removed clause.

Resolves #2207

- Visit `/about/tech` (Live Cam Controls section) and `/about/tech/commands` (Scenes section).
- The `!ptzlist` description should read "Get all preset positions for a camera", with no mention of cameras on stream.
- `!scenecams` should appear under the Scenes category with the description "Get names and slots of all cameras on stream currently".


apps/website/src/data/tech/commands.ts
MattIPv4:
```suggestion
        required: true,
```
This should now be required, the command cannot be run without a camera name?

MillerSpil: My bad 🤣 should have caught that. All changed now.

MattIPv4 (approved):

PR #2226, opened 2026-06-28
Save twitch subs as donations

### LINKED ISSUE #1588: Extend donation manager and feed to support twitch subscriptions #1585

At this point are they even donations anymore? Anyways, to replace the SE activity feed we want to also support (gifted) subs I guess.

See #1584 

better feed more data

Moolash3:
## Describe your changes

Adds twitch webhook subscriptions for new, gift, and resubscriptions. Added a filter to the donation list page to filter out subscriptions which is off by default.

Updating the `received` donations index to be a multicolumn index to support filtering on provider. This allows us to be more granular with filtering with less of a performance cost. I seeded 1m donations locally and ran queries of recievedAt desc and filtering by neoncrm and paypal providers and it was performant.

image

The `channel.chat.notification` (https://dev.twitch.tv/docs/eventsub/eventsub-subscription-types/#channelchatnotification) event should be the one used for all twitch related webhooks in the future as it has all subscription, charity donation, and bit information. The reasons these changes do NOT implement this webhook are the following,

- It requires a higher level of permissions and that would involve changing the oauth flow and reauthorizing the app with the channels.
- The amounts of events sent would be significantly higher. Especially if subscribed across all three channels.

Using the `channel.chat.notification` webhook would allow us to know the following,

- If a user upgraded from a prime
- Was a prime sub
- Continued after a gift
- Bits
- Charity donations

Resolves #1588 
...

I was able to test using the Twitch CLI (https://dev.twitch.tv/docs/cli/). Generate your payload using the cli and then add `twitch-eventsub-message-id` as a UUID and `twitch-eventsub-message-timestamp` as a datetime as headers.

- Must turn the filter on the donations page for subs to show up
- If you dont have paypal/neoncrm env keys, I had to comment out those providers in the manager init

...

### DIFF
--- a/apps/donations/worker/src/providers/twitch/schema.ts
+++ b/apps/donations/worker/src/providers/twitch/schema.ts
@@ -6,8 +6,15 @@ export const TwitchChallengePayload = z.object({
 export const TwitchSubscriptionPayload = z.object({
   id: z.string(),
   type: z
-    .literal("channel.charity_campaign.donate")
-    .describe("Only channel.charity_campaign.donate events are supported."),
+    .literal([
+      "channel.charity_campaign.donate",
+      "channel.subscribe",
+      "channel.subscription.gift",
+      "channel.subscription.message",
+    ])
+    .describe(
+      "Only channel.charity_campaign.donate, channel.subscribe, channel.subscription.gift, and channel.subscription.message events are supported.",
+    ),
   version: z.string(),
   status: z.string(),
   condition: z.object({
@@ -17,6 +24,7 @@ export const TwitchSubscriptionPayload = z.object({
     method: z.string(),
     callback: z.string(),
   }),
+  created_at: z.iso.datetime(),
 });
 
 export const TwitchAmountPayload = z.object({
@@ -31,7 +39,7 @@ export const TwitchAmountPayload = z.object({
     ),
 });
 
-export const TwitchCharityDonationEventPayload = z.object({
+export const TwitchDonationEventPayload = z.object({
   id: z.string(),
   campaign_id: z.string(),
   broadcaster_user_id: z.string(),
@@ -42,52 +50,28 @@ export const TwitchCharityDonationEventPayload = z.object({
   amount: TwitchAmountPayload,
 });
 
-export const TwitchNewSubscriptionEventPayload = z.object({
-  id: z.string(),
-  campaign_id: z.string(),
+export const TwitchSubscriptionDonationEventPayload = z.object({
   broadcaster_user_id: z.string(),
   user_id: z.string(),
   user_login: z.string(),
   user_name: z.string(),
-  tier: z.number(),
-  is_gift: z.boolean(),
-});
-
-export const TwitchReSubscriptionEventPayload = z.object({
-  id: z.string(),
-  campaign_id: z.string(),
-  broadcaster_user_id: z.string(),
-  user_id: z.string(),
-  user_login: z.string(),
-  user_name: z.string(),
-  tier: z.number(),
+  tier: z.literal(["1000", "2000", "3000"]),
+  is_gift: z.boolean().optional(),
+  total: z.number().optional(),
+  cumulative_total: z.number().optional(),
+  cumulative_months: z.number().optional(),
+  is_anonymous: z.boolean().optional(),
   message: z.object().optional(),
-  cumulative_months: z.number(),
-  streak_months: z.number(),
-  duration_months: z.number(),
+  streak_months: z.number().optional(),
+  duration_months: z.number().optional(),
 });
-
-export const TwitchGiftSubscriptionEventPayload = z.object({
-  id: z.string(),
-  campaign_id: z.string(),
-  broadcaster_user_id: z.string(),
-  user_id: z.string(),
-  user_login: z.string(),
-  user_name: z.string(),
-  tier: z.number(),
-  total: z.number(),
-  cumulative_total: z.number(),
-  is_anonymous: z.boolean(),
-});
-
-export const TwitchEventPayload = z.xor([
-  TwitchCharityDonationEventPayload,
-  TwitchNewSubscriptionEventPayload,
-  TwitchReSubscriptionEventPayload,
-  TwitchGiftSubscriptionEventPayload,
-]);
 
 export const TwitchNotificationPayload = z.object({
   subscription: TwitchSubscriptionPayload,
-  event: TwitchEventPayload,
+  event: TwitchDonationEventPayload,
+});
+
+export const TwitchSubscriptionNotificationPayload = z.object({
+  subscription: TwitchSubscriptionPayload,
+  event: TwitchSubscriptionDonationEventPayload,
 });
--- a/apps/donations/worker/src/providers/twitch/schema.ts
+++ b/apps/donations/worker/src/providers/twitch/schema.ts
@@ -31,7 +31,7 @@ export const TwitchAmountPayload = z.object({
     ),
 });
 
-export const TwitchEventPayload = z.object({
+export const TwitchCharityDonationEventPayload = z.object({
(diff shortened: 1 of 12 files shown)


MattIPv4:
@changesbyjames @pjeweb, could I leave this one to y'all to review, as I think you both have a lot more context on this part of the site?

pjeweb:
@Moolash3 excited to review this! i will try to make time soon.

I assume you considered having a separate entity/table for subscription events, right?

What additional scopes/perms would we need? Is it the broadcaster or a moderator or is it the app token itself?

Moolash3:
> @Moolash3 excited to review this! i will try to make time soon.
> 
> I assume you considered having a separate entity/table for subscription events, right?
> 
> What additional scopes/perms would we need? Is it the broadcaster or a moderator or is it the app token itself?

First off, no worries, im just happy to try and help out! Forgive my amateurish typescript, my primary language is elixir so im not that familiar with prisma and next but im excited to learn more.

I considered adding a `twitch_subscriptions` table or just a `contributions` table for non traditional donations (bits/subs/superchats/etc...) and I ended up with just adding to the donations table because I didnt want to over engineer the work I was trying to do. If there are wants to be able to query more granular data about subs like "get all x tier subs" then I would definitely want to split out the data so it can be indexed properly. It came down to how I saw the data being used.

For the oauth changes, either we can seed a new token with the updated scopes 
```
Requires user:read:chat scope from chatting user. If app access token used, then additionally requires user:bot scope from chatting user, and either channel:bot scope from broadcaster or moderator status.
```

but then we would have to manually update if the refresh_token ever expired or was invalid. Or create somewhere in the admin to allow new channels to connect with the new scopes.

https://github.com/BarryCarlyon/twitch_misc/blob/main/eventsub/websockets/web/chat/README.md

apps/website/src/components/admin/donations/DonationFeedItem.tsx
```diff
@@ -133,7 +158,9 @@ function DonationFeedItem({
             />
           }
         >
-          {formatDonationAmount(donation.amount)}
+          {donation.provider === "twitchsubscription"
+            ? formatTwitchSub(donation as unknown as TwitchSubscriptionDonation)
```
pjeweb: **nit (typing):** I'd avoid casting the whole feed item through `unknown`. Cast just the metadata field:

```suggestion
          {donation.provider === "twitchsubscription"
            ? formatTwitchSub(
                donation.providerMetadata as TwitchSubscriptionDonation["providerMetadata"],
              )
            : formatDonationAmount(donation.amount)}
```

And then `formatTwitchSub` should take that metadata type instead of `TwitchSubscriptionDonation`.

apps/donations/worker/src/providers/twitch/schema.ts
```diff
@@ -42,7 +50,28 @@ export const TwitchEventPayload = z.object({
...
+  user_id: z.string(),
+  user_login: z.string(),
+  user_name: z.string(),
+  tier: z.literal(["1000", "2000", "3000"]),
+  is_gift: z.boolean().optional(),
+  total: z.number().optional(),
+  cumulative_total: z.number().optional(),
+  cumulative_months: z.number().optional(),
+  is_anonymous: z.boolean().optional(),
+  message: z.object().optional(),
```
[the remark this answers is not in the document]
pjeweb: I agree with copilot here!

apps/donations/worker/src/providers/twitch/subscription.ts
```diff
@@ -0,0 +1,230 @@
...
+
+    if (type === "revocation") {
+      await this.clear();
+      return new Response(null, { status: 201 });
+    }
+
+    const uniqueMessageId = request.headers.get(TwitchEventHubMessageIdHeader);
+    if (!uniqueMessageId) {
+      await this.clear();
+      return new Response(null, { status: 400 });
```
pjeweb: On missing Twitch-Eventsub-Message-Id, return 400 but don’t wipe provider config (that’s for revocation).

apps/donations/worker/src/providers/twitch/subscription.ts
pjeweb, line 1: Did you consider anonymous gift subscriptions?

When`is_anonymous = true`, Twitch can null `user_id` / `user_login` / `user_name`. Schema + `donatedBy.username` should allow that and fall back to "Anonymous", or validation will 400 and drop the event.

pjeweb (commented):
Awesome work! I am happy with the decisions (using the donations for now) etc.

pjeweb:
**suggestion (badge copy + resub message):** The compact labels (`Gifted 5 T1`, `Resub T2 1X`) are a bit hard to scan — gift count vs tier vs months blur together. Also use `×` and spell out `Tier`.

Something like:

- Gift → `Gifted 5× Tier 1`
- New → `New · Tier 1`
- Resub → `Resub · Tier 2 · 47 mo` using `cumulativeMonths`
  - If `durationMonths > 1` (e.g. 3/6 months upfront), keep that visible too, e.g. `Resub · Tier 2 · 6× · 47 mo` or `Resub · Tier 2 · +6 mo (47 total)`
  - If `durationMonths === 1`, omit it so you don’t get a noisy `1×`

A bit longer, but should still fit on mobile unless the username is very long (username already ellipsizes).

Separate small gap: `channel.subscription.message` includes `event.message.text`, but we don’t persist it. The feed already renders `donation.note` — worth setting `note` from that text so resub messages show up!

pjeweb:
Rebased onto latest `main` (no conflicts) so we can get a fresh Vercel preview — the previous build failure (`Can't resolve 'next-superjson-plugin/tools'`) looked unrelated to these changes and more like a stale branch vs current deps.

Force-pushed to `add-twitch-sub-events` on your fork. Your commits are preserved (authors unchanged). Ping me if you'd rather have handled the rebase yourself!

apps/donations/worker/src/providers/twitch/schema.ts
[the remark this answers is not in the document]
Moolash3: Sounds good

apps/website/src/components/admin/donations/DonationFeedItem.tsx
Moolash3: Perfect, Thanks

apps/donations/worker/src/providers/twitch/subscription.ts
Moolash3: Sounds good

apps/donations/worker/src/providers/twitch/subscription.ts
Moolash3: Sounds good.

Moolash3:
> **suggestion (badge copy + resub message):** The compact labels (`Gifted 5 T1`, `Resub T2 1X`) are a bit hard to scan — gift count vs tier vs months blur together. Also use `×` and spell out `Tier`.
> 
> Something like:
> 
>     * Gift → `Gifted 5× Tier 1`
> 
>     * New → `New · Tier 1`
> 
>     * Resub → `Resub · Tier 2 · 47 mo` using `cumulativeMonths`
>       
>       * If `durationMonths > 1` (e.g. 3/6 months upfront), keep that visible too, e.g. `Resub · Tier 2 · 6× · 47 mo` or `Resub · Tier 2 · +6 mo (47 total)`
>       * If `durationMonths === 1`, omit it so you don’t get a noisy `1×`
> 
> 
> A bit longer, but should still fit on mobile unless the username is very long (username already ellipsizes).
> 
> Separate small gap: `channel.subscription.message` includes `event.message.text`, but we don’t persist it. The feed already renders `donation.note` — worth setting `note` from that text so resub messages show up!

Perfect, I will work on implementing these changes this week. Apologies for the delay in responding.

Moolash3:
image

PR #2229, opened 2026-06-29
Add pollinator census to a new "Get involved" section in nav

changesbyjames:
## Describe your changes

...

...


apps/website/src/data/navigation.ts
```diff
@@ -68,23 +68,33 @@ export const mainNavStructure: NavStructure = {
...
-            title: "Show & Tell",
-            link: "/show-and-tell",
-          },
           bookClub: {
             title: "Alveus Book Club",
             link: "/book-club",
           },
         },
       },
+
```
MattIPv4:
```suggestion
```

apps/website/src/data/navigation.ts
```diff
@@ -68,23 +68,33 @@ export const mainNavStructure: NavStructure = {
...
+        title: "Get involved",
+        links: {
+          showAndTell: {
+            title: "Show & Tell",
+            link: "/show-and-tell",
+          },
+          pollinatorCensus: {
+            title: "Pollinator Census",
+            link: "https://census.alveussanctuary.org",
+          },
```
MattIPv4, line 82: Do we want this to be `external: true,`?

changesbyjames: What do you think? I don't mind either way, just whether it opens it as a new tab or replaces the tab

MattIPv4: I'd lean toward it being external probably, same as the apparel link. Do you want to set up `/census` in next.config to redirect?

MattIPv4:
@copilot nit, can you place the census redirect after the plushy redirect, instead of at the end of the array?

pjeweb:
:omgBruh: AI cant even follow formatting rules

MattIPv4 (approved):

PR #2307, opened 2026-08-08
New Wishlist System 

DanMan149:
## What this adds

A full wishlist feature for the site, similar to an Amazon Wishlist, but built for Alveus specifically:

- Public wishlist page with category filters, urgent-needs section, and progress bars
- Admin panel to add items by pasting a product URL (auto-fetches title/price/image) or by creating a custom funding goal with no linked product
- Combined funding, donations of any amount stack toward a goal's target
- PayPal donations with an optional message from the donor
- Neon CRM logging + automatic tax receipts after each donation
- Discord webhook alerts when a donation comes in
- A generic stream overlay that pops up an animated alert on donation (can ask an artist to make a better one)
- Staff fulfillment dashboard,  fully-funded items list with a one-click pre-filled Amazon cart, plus a donor-message queue for on-stream shoutouts
- Two-stage item lifecycle: fully funded → marked as opened/unboxed

## Demo

(https://alveus-wishlist-test-sooty.vercel.app/) — click-through UI demo with mock data, no real payments.

## Payment security

Since this handles real money, wanted to flag the security model up front:

- Donations go through PayPal's Order's API, not embeddable buttons, the amount is set server-side and locked into the order object, so it can't be edited by the donor in the browser
- Every payment gets a confirmed `Completed` status check via PayPal's capture endpoint before anything is marked fulfilled
- PayPal also independently fires a webhook confirming the same payment, so there's a second server-side confirmation path
- Neon CRM is only ever called from the server after a payment is confirmed — donors never touch a Neon-hosted form

## A note on testing

I wasn't able to run a full local `pnpm build` since `@alveusgg/data` needs org package registry access I don't have. I did verify the new/changed files compile correctly against real `zod`, `@trpc/server`, and type shapes pulled directly from the actual `trpc.ts`, `permissions.ts`, `user-roles.ts`, and `donations.ts` in this repo, so the procedures follow the same patterns as `donationsRouter` (protectedProcedure + createCheckPermissionMiddleware for staff actions, prisma imported directly). Happy to fix anything CI flags that I couldn't catch without full repo access.

## What's still open

- A cache-cleanup cron for cached product metadata
- A scheduled caller for the Discord weekly-digest function (the function exists, nothing triggers it yet)
- WebSocket push for the stream overlay (currently polls every 5s)

## New files

- `apps/website/src/server/trpc/router/wishlist.ts`
- `apps/website/src/components/wishlist/*` (5 files)
- `apps/website/src/pages/wishlist/*`, `pages/admin/wishlist/*`, `pages/overlay/*`, `pages/api/webhooks/paypal.ts`
- `apps/website/src/utils/{fetchUrlMeta,paypal,neonCrm,discord,amazonCart,metaCache}.ts` + `utils/providers/*` (7 provider files for Amazon/Chewy/PetSmart/Petco/Etsy/Walmart)
- `apps/database/prisma/wishlist.prisma` (schema addition)
- Small additions to `data/user-roles.ts` and `data/permissions.ts` for a new `wishlist` staff role


apps/website/src/components/wishlist/CategoryManager.tsx
```diff
@@ -0,0 +1,168 @@
...
+}
+
+interface FormState {
+  name: string;
+  slug: string;
+  sortOrder: number;
+}
+
+const toSlug = (name: string) =>
+  name.toLowerCase().replace(/\s+/g, "-").replace(/[^a-z0-9-]/g, "");
```
MattIPv4, line 32: Can we rely on the existing `convertToSlug` util?

pjeweb:
Thanks for working on this! So excited. I think a few files didnt make it in the PR:

Missing / needs fixing for CI:
- Prisma wishlist models never committed (WishlistItem, WishlistCategory, WishlistDonationOrder, WishlistMetaCache)
- amazonCart.ts imported but missing
- discord.ts seems to overwrite the existing helpers — keep escapeLinksForDiscord / getDiscordTimestamp
- Admin pages: wrong AdminPageLayout path; use @/components/admin/AdminPageLayout + getAdminSSP + menuItems
- Import types from @alveusgg/database, not @prisma/client

apps/website/src/components/wishlist/WishlistAdminTable.tsx
```diff
@@ -0,0 +1,192 @@
...
+                        href={item.url}
+                        target="_blank"
+                        rel="noopener noreferrer"
+                        className="text-gray-400 hover:text-gray-600 transition-colors"
+                        title="Open product link"
+                      >
+                        <svg className="w-4 h-4" fill="none" viewBox="0 0 24 24" stroke="currentColor">
+                          <path strokeLinecap="round" strokeLinejoin="round" strokeWidth={2} d="M10 6H6a2 2 0 00-2 2v10a2 2 0 002 2h10a2 2 0 002-2v-4M14 4h6m0 0v6m0-6L10 14" />
+                        </svg>
+                      </a>
```
MattIPv4: Can we use our existing `Link` component here, for consistent styling?

apps/website/src/components/wishlist/WishlistDonateButton.tsx
```diff
@@ -0,0 +1,209 @@
...
+        aria-label={`Donate the value of ${itemTitle} via PayPal`}
+      >
+        <svg viewBox="0 0 24 24" width="14" height="14" fill="currentColor" aria-hidden="true">
+          <path d="M7.076 21.337H2.47a.641.641 0 0 1-.633-.74L4.944.901C5.026.382 5.474 0 5.998 0h7.46c2.57 0 4.578.543 5.69 1.81 1.01 1.15 1.304 2.42 1.012 4.287-.023.143-.047.288-.077.437-.983 5.05-4.349 6.797-8.647 6.797h-2.19c-.524 0-.968.382-1.05.9l-1.12 7.106zm14.146-14.42a3.35 3.35 0 0 0-.607-.541c-.013.076-.026.175-.041.254-.93 4.778-4.005 7.201-9.138 7.201h-2.19a.563.563 0 0 0-.556.479l-1.187 7.527h-.506l-.24 1.516a.56.56 0 0 0 .554.647h3.882c.46 0 .85-.334.922-.788.06-.26.76-4.852.816-5.09a.932.932 0 0 1 .923-.788h.58c3.76 0 6.705-1.528 7.565-5.946.36-1.847.174-3.388-.777-4.471z"/>
+        </svg>
+        Donate with PayPal
+      </button>
+
+      {open && (
+        <div
```
MattIPv4: Can we use a `@headlessui/react` dialog here, so that it handles the accessiblity requirements for us?

apps/website/src/components/wishlist/WishlistGrid.tsx
```diff
@@ -0,0 +1,197 @@
...
+            <div className="p-4 flex flex-col flex-1 gap-2">
+              <div className="flex flex-wrap items-center gap-1.5">
+                {priority && (
+                  <span className={`text-xs font-medium px-2 py-0.5 rounded-full border ${priority.cls}`}>
+                    {priority.label}
+                  </span>
+                )}
+                {isGoal && (
+                  <span className="text-xs font-medium px-2 py-0.5 rounded-full bg-amber-50 text-amber-800 border border-amber-200">
+                    🎯 Funding Goal
```
MattIPv4: Can we use a proper SVG icon here, rather than a unicode char?

apps/website/src/components/wishlist/WishlistGrid.tsx
```diff
@@ -0,0 +1,197 @@
...
+              {isGoal && item.goalAmountCents != null ? (
+                <GoalProgressBar raisedCents={item.raisedAmountCents} goalCents={item.goalAmountCents} />
+              ) : (
+                showQuantityProgress && <ProgressBar value={item.quantityFulfilled} max={item.quantity} />
+              )}
+
+              <div className="mt-auto pt-1 flex flex-col gap-1">
+                {isFulfilled ? (
+                  <button disabled className="w-full text-center text-xs font-medium py-2 px-4 rounded-lg bg-gray-100 text-gray-400 cursor-default">
+                    Fulfilled ✓
```
MattIPv4: Can we use a proper SVG icon here, rather than a unicode char?

apps/website/src/components/wishlist/WishlistGrid.tsx
```diff
@@ -0,0 +1,197 @@
...
+                  />
+                ) : (
+                  <>
+                    <a
+                      href={item.url ?? "#"}
+                      target="_blank"
+                      rel="noopener noreferrer"
+                      className="block w-full text-center text-xs font-medium py-2 px-4 rounded-lg bg-green-700 text-white hover:bg-green-800 transition-colors"
+                    >
+                      View &amp; purchase →
```
MattIPv4: Can we use a proper SVG icon here, rather than a unicode char?

apps/website/src/components/wishlist/WishlistItemForm.tsx
```diff
@@ -0,0 +1,469 @@
...
+        if (meta.price && !price) setPrice(meta.price);
+        setMetaFetched(true);
+      }
+      setFetchingMeta(false);
+    }, 800);
+
+    return () => {
+      if (urlDebounceRef.current) clearTimeout(urlDebounceRef.current);
+    };
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
MattIPv4: What deps are missing here, and why? Please include a justification for the disable, or have the correct dependencies defined.

apps/website/src/components/wishlist/WishlistGrid.tsx
```diff
@@ -0,0 +1,197 @@
...
+        <div className="h-full bg-green-500 rounded-full transition-all duration-500" style={{ width: `${pct}%` }} />
+      </div>
+    </div>
+  );
+}
+
+function GoalProgressBar({ raisedCents, goalCents }: { raisedCents: number; goalCents: number }) {
+  const pct = goalCents > 0 ? Math.min(100, Math.round((raisedCents / goalCents) * 100)) : 0;
+  const raised = (raisedCents / 100).toLocaleString(undefined, { maximumFractionDigits: 0 });
+  const goal = (goalCents / 100).toLocaleString(undefined, { maximumFractionDigits: 0 });
```
MattIPv4: Use the `useLocaleString` hook please, so this doesn't cause any hydration errors

apps/website/src/pages/wishlist/donate/success.tsx
```diff
@@ -0,0 +1,117 @@
...
+      .then((result) => {
+        setAmount(result.amount ?? null);
+        setState("success");
+      })
+      .catch((err) => {
+        setError(err instanceof Error ? err.message : "Something went wrong");
+        setState("error");
+      });
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [orderId, itemId, router.isReady]);
```
MattIPv4, line 38: Please include a justification for any dependencies that aren't included

apps/website/src/utils/metaCache.ts
MattIPv4, line 1: In-memory caching will unfortunately not work due to how our backend runs. If you need caching, it'll need to be done via Redis. Please don't thrash the DB with reads/writes for caching.

apps/website/src/utils/neonCrm.ts
MattIPv4, line 1: Please extend the existing dedicated `@alveusgg/neon-crm-api` package that we have, instead of rolling a duplicative util file

apps/website/src/utils/discord.ts
MattIPv4, line 1: ~~You appear to have wiped out the existing contents of this file. Please restore those existing util methods.~~ pje said this

apps/website/src/server/trpc/router/wishlist.ts
MattIPv4, line 1: Please split this into two separate routes, as is the established pattern in the code base; a public router and an admin router.

MattIPv4 (commented):
Thanks for working on this. Some general thoughts so I don't have to repeat them across every file:

1. ~~You seem to be missing the Prisma schema changes required to support this change?~~ pje said this
2. Can we remove the file name comments, they seem redundant?
3. ~~Can you update all the database imports to be from `@alveusgg/database`, rather than `@prisma/client`?~~ pje said this
4. Can you update all trpc utils (and any other imports) to use the `@` root alias (e.g., `@/utils/trpc`)?
5. Can we use our existing field components in `@/components/shared/form`, rather than rolling custom inputs in the forms?
6. Can we move any SVG icons into `@/icons` please, with correct attribution, rather than anonymous SVGs inline in components?
7. Can we use the existing `classes` util for computing dynamic `className` props, rather than template literals?
8. Can we use `next/image` rather than a bare `<img` tag, avoiding the need to disable eslint in a bunch of places? Are we uploading the images so that we don't run into CSP issues, or are we relying on Next.js proxying?
9. Please avoid using `as xyz` type-casts, and definitely never `as unknown as xyz` -- TypeScript ensures our logic is safe, we shouldn't need to bypass it.
10. Really not sure I'm keen on any of the provider logic, especially the Playwright-based ones, and I don't think we should be doing web scraping at all -- can we rely on APIs if they exist, and if not, manual data entry?

DanMan149:
Few things I need input on:

1. Shared form field components (@/components/shared/form): could you point to an example usage so the wishlist forms can be converted to use them instead of raw
2. Icons folder (@/icons): same, a pointer to how an existing component imports from there (and whether SVGs need a specific wrapper component or just move as raw files) would help get this right in one pass.
3. Next/image instead of : can switch, but since wishlist images come from arbitrary external URLs (product pages, or admin-pasted image links), this needs either domain allowlisting in next.config or confirmation you're relying on something else (a proxy, Next's remote patterns, etc.) What's the existing approach for other externally-sourced images in the site, if any?
4. Playwright/scraping: I'm down to removing it if there's a strong preference against scraping. The tradeoff is that without it, Amazon links won't auto-fill since Amazon blocks plain HTTP fetches, so staff would need to enter Amazon items manually. Chewy/PetSmart/Petco scraping is much lighter (its just reading public Open Graph tags), would keeping just that lighter-weight OG tag reading and dropping the Playwright/Amazon-specific piece be better or should I just make data entry entirely manual?
on relying on API's: I did a lot of research on this and from I've found out Amazon's Product Advertising API has been deprecated and replaced by a new "Creators API" but being eligible for it is tough: you need an Amazon Associates affiliate account with at least ten qualifying sales within a rolling window just to get approved, and if that account goes 30 days without maintaining sales, Amazon revokes API access and you have to reapply.

MattIPv4:
> 1\. Shared form field components (@/components/shared/form): could you point to an example usage so the wishlist forms can be converted to use them instead of raw

`apps/website/src/components/shared/form/TextField.tsx` and all its usages, for example, in `apps/website/src/components/admin/calendar-events/CalendarEventForm.tsx`.

> 2\. Icons folder (@/icons): same, a pointer to how an existing component imports from there (and whether SVGs need a specific wrapper component or just move as raw files) would help get this right in one pass.

The SVGs need a specific wrapper. Take a look at any of the files in there today and you will see how they work, they're all the same style.

For example, `apps/website/src/icons/IconShare.tsx` and its usage in `apps/website/src/components/shared/ShareButton.tsx`.

> 3\. Next/image instead of : can switch, but since wishlist images come from arbitrary external URLs (product pages, or admin-pasted image links), this needs either domain allowlisting in next.config or confirmation you're relying on something else (a proxy, Next's remote patterns, etc.) What's the existing approach for other externally-sourced images in the site, if any?

Accepting arbitrary image URLs will not be safe, nor will it work. You will either need to implement image uploads, as we do in a few other places in the codebase, or you will need to have a specific allowlist of known hosts that we accept images from that we can add to `remotePatterns` in `next.config.ts` to support proxying them.

> 4\. Playwright/scraping: I'm down to removing it if there's a strong preference against scraping. The tradeoff is that without it, Amazon links won't auto-fill since Amazon blocks plain HTTP fetches, so staff would need to enter Amazon items manually. Chewy/PetSmart/Petco scraping is much lighter (its just reading public Open Graph tags), would keeping just that lighter-weight OG tag reading and dropping the Playwright/Amazon-specific piece be better or should I just make data entry entirely manual?
on relying on API's: I did a lot of research on this and from I've found out Amazon's Product Advertising API has been deprecated and replaced by a new "Creators API" but being eligible for it is tough: you need an Amazon Associates affiliate account with at least ten qualifying sales within a rolling window just to get approved, and if that account goes 30 days without maintaining sales, Amazon revokes API access and you have to reapply.

I am strongly against the Playwright pieces. Running Playwright on-demand on the backend feels rather dangerous, and likely rather resource-intensive as well.

I am less strongly against the standard HTML scraping pieces, though given what Alveus is and stands for, I suspect we should lean into being good internet citizens and avoid scraping. If there are APIs we can use (even third-party if we can't access Amazon's first-party one), let's do that, but if not, manual data entry is likely the best/only option.

DanMan149:
- Removed playwright and scraping completely, manual entry for every retailer. Even if we use third party API's the problem is still there, instead of scraping itself, Alveus is paying someone to do it. 

- Went with image allowlist — 10 hosts added to the existing remotePatterns array. (Amazon, Chewy, Petco, Petsmart, Walmart, Target, HEB, Lowes, HomeDepot and IBB as a generic one alongside imgur which is already there)

DanMan149:
Also removed the existing /wishlist → Amazon redirect in next.config.ts since it was permanently 301-redirecting away from where the new wishlist page needs to live, so it had to go for this page to actually be reachable. Worth noting this is a permanent: true redirect, so some browsers/CDNs may have it cached, users might still get sent to Amazon after this merges and may need a hard refresh.

DanMan149:
Took a look at @alveusgg/neon-crm-api before making any changes and wanted to flag something before I take a guess at this, since it's payment-adjacent. Right now the package has no createDonation or account-creation function. Everything in schema.ts related to donations is either:

1. Reading existing data (getDonation, searchDonations), or
2. A DonationWebhookPayload shape representing something Neon sends to us i.e. Neon already knows about the donation (presumably from their own hosted checkout) and is notifying our system after the fact.

I don't see anything in the package suggesting Neon CRM's API has a direct "create a donation record" endpoint we could wrap, since it looks like the existing pattern assumes donations always originate on Neon's side, and this package's job is to read that data back out or receive their webhook, not to push new donations into Neon from an external source. That matters because the PayPal → Neon CRM sync needs the opposite direction: donor pays via PayPal on Alveus site, and then we create a record in Neon after the fact.
So before I move on with this, wanted to check that if Neon's API actually have a direct "create donation" endpoint, or is the intended pattern here different for e.g. should PayPal donations get logged through Neon's own webhook/import mechanism instead of a direct API push? 

Also there's some fixes that I still need to work on, if there is anything else that I missed out  do let me know.

pjeweb:
> wanted to check that if Neon's API actually have a direct "create donation" endpoint

Yes! I just didnt model it in the api package! https://developer.neoncrm.com/api-v2/#/Donations/createUsingPOST_1

I assume its gonna share a lot of schema with the other apis though. API documentation isnt the best, but I am happy to help or we can see how/if they appear and work our way there. They have surprises like UTC timestamps being in local time etc ;)

pjeweb:
damn copilot really created quite a lot of comments. any way to just close all and let it redo its thing? lol

DanMan149:
Co-pilot be yapping, idk if there is a close-all option lmao

MattIPv4:
> damn copilot really created quite a lot of comments. any way to just close all and let it redo its thing? lol

There isn't, and it has a memory, so it wouldn't comment on findings it has already commented on.

A lot of the comments are unfortunately duplicative of what we flagged, but I think it does also raise some valid concerns, so it probably does make sense for @DanMan149 to go through them and address any remaining?

DanMan149:
Yeah, I am going through the linting, prettier and types checks it failed, earlier.

DanMan149:
> Yes! I just didnt model it in the api package! https://developer.neoncrm.com/api-v2/#/Donations/createUsingPOST_1

I'm working on creating the CreateDonationInput Zod schema matching this (verified it compiles cleanly against zod@4, same as the package's real package.json), scoped down to just the fields the wishlist integration would actually populate. But I kinda hit a wall: accountId is the first field, and there's nothing in this schema suggesting a donation can be created without one. That means before we can call this endpoint, we need to know which Neon account the donor corresponds to.

So two questions:

1. Does Neon have an account search-by-email or create-account endpoint? The current package only has getAccount(id), for a first-time PayPal donor with no existing Neon record, we'd need either a search endpoint (to check if they already have an account) or a create endpoint (to make one).
2. What's the tenderType ID for an external/generic payment method? payments[ ].tenderType is a numeric reference, that is configured to payment methods in Neon. It's specific to Alveus' Neon instance and not something I can look up or guess. Is there already a "PayPal" or generic "external" tender type configured, or would one need to be added?

Also I went through all the reviews by Co-Pilot and resolved them, so if one of you could run it again, to check if I got everything or not.

PR #2339, opened 2026-08-29
Stop show and tell videos playing when scrolled out of view

### LINKED ISSUE #2316: [Bug]: S&T videos continue playing even when not visible

When going through multiple videos/images on a show and tell post, video continues playing in the background even when the video is no longer the current item being shown.

Either pause video, or unmount video when it is no longer visible.

### Steps to reproduce

Play video.
Go to next image/video.
Hear video still playing in the background.

### Browser specs

Both chrome and firefox

### Anything else?

_No response_

mrusse:
## Describe your changes

Video attachments in the show and tell lightbox kept playing in the background after scrolling to another item. This tracks whether a video's container is visible and unmounts the iframe when it isn't.

Resolves #2316

Tested with both YouTube and Streamable embeds, on Chrome and Firefox. Testing seen below (unmute when watching).

### Before:
- Chrome: 

https://github.com/user-attachments/assets/dba2ad73-e672-4f47-a106-d1d1c27a0fec

- Firefox:

https://github.com/user-attachments/assets/4ca13447-bbae-447e-877f-1f9544f1d24a

### After:
- Chrome:

https://github.com/user-attachments/assets/2b5428b5-4de9-440e-a97d-080ad4327324

- Firefox:

https://github.com/user-attachments/assets/5765ea17-6b6c-43d3-af15-546656c79520

### DIFF
--- a/apps/website/src/components/show-and-tell/gallery/VideoItem.tsx
+++ b/apps/website/src/components/show-and-tell/gallery/VideoItem.tsx
@@ -1,4 +1,5 @@
 import type { JSX, ReactNode } from "react";
+import { useEffect, useRef, useState } from "react";
 
 import type { LinkAttachment } from "@alveusgg/database";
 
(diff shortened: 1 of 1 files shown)


MattIPv4:
🤔 Can't we update the lightbox implementation itself to handle this, rather than needing to do it for the specific embed implementation? Doesn't this same issue affect the lightboxes on other pages that have videos (collabs, about, etc.)?

mrusse:
> 🤔 Can't we update the lightbox implementation itself to handle this, rather than needing to do it for the specific embed implementation? Doesn't this same issue affect the lightboxes on other pages that have videos (collabs, about, etc.)?

For sure, that would be a much better fix. I tested this more and as you suggested the collabs, about, and ambassadors pages also have this issue. I will close this and open a separate PR when I have a more general fix for the lightbox. Ty!

PR #2341, opened 2026-08-30
Pause lightbox videos when they scroll out of view

### LINKED ISSUE #2316: [Bug]: S&T videos continue playing even when not visible

mrusse:
## Describe your changes

Video embeds kept playing in the background after scrolling to another item in the lightbox. This fixes it in the lightbox/carousel implementation itself, so it covers every page that uses the lightbox for videos (collaborations, ambassador clips, about pages, etc.)

Carousel now reports which item is currently scrolled into view. Lightbox uses that to tell any video iframe that's no longer active to pause.

Resolves #2316

This PR replaces #2339. This addresses the feedback there to fix it at the lightbox level instead. Let me know of any other issues with this fix.

Tested with both YouTube and Streamable embeds. Verified working on both Chrome and Firefox. Videos below show before/after on Chrome.

**Before:**

https://github.com/user-attachments/assets/8a29102c-90a2-4d01-b0be-f1c42ba77350

**After:**

https://github.com/user-attachments/assets/9aed9281-2bae-4391-b232-a85b712b67a7

### DIFF
--- a/apps/website/src/components/content/Carousel.tsx
+++ b/apps/website/src/components/content/Carousel.tsx
@@ -30,6 +30,8 @@ type CarouselProps = {
   // only via the arrows or programmatic scrolling. Useful when items capture
   // their own pointer gestures (e.g. a pannable canvas).
   draggable?: boolean;
+  // Tells you which item is currently visible, by its key
+  onActiveKeyChange?: (key: string | undefined) => void;
 };
 
 const overlayButtonClassName =
@@ -47,6 +49,7 @@ const Carousel = ({
   variant = "default",
   overlayClassName = "",
   draggable = true,
+  onActiveKeyChange,
 }: CarouselProps) => {
   const reducedMotion = usePrefersReducedMotion();
 
@@ -141,7 +144,18 @@ const Carousel = ({
     else if (nearLeft) setState("start");
     else if (nearRight) setState("end");
     else setState("scrolling");
-  }, [interacted]);
+
+    // Work out which item is currently scrolled into view
+    if (onActiveKeyChange) {
+      const gap =
+        Number.parseFloat(getComputedStyle(current).columnGap) ||
+        Number.parseFloat(getComputedStyle(current).gap) ||
+        0;
+      const keys = Object.keys(items);
+      const index = Math.round(current.scrollLeft / (width + gap));
+      onActiveKeyChange(keys[Math.max(0, Math.min(index, keys.length - 1))]);
+    }
+  }, [interacted, items, onActiveKeyChange]);
 
   // Run the scroll handler on load to check our current state
   // Run it on any window resize to check our state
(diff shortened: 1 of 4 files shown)


apps/website/src/components/content/Carousel.tsx
```diff
@@ -141,7 +144,18 @@ const Carousel = ({
...
+    if (onActiveKeyChange) {
+      const gap =
+        Number.parseFloat(getComputedStyle(current).columnGap) ||
+        Number.parseFloat(getComputedStyle(current).gap) ||
+        0;
+      const keys = Object.keys(items);
+      const index = Math.round(current.scrollLeft / (width + gap));
+      onActiveKeyChange(keys[Math.max(0, Math.min(index, keys.length - 1))]);
+    }
+  }, [interacted, items, onActiveKeyChange]);
```
MattIPv4: Is `items` a stable dependency? I fear it may not be?

apps/website/src/components/content/Carousel.tsx
```diff
@@ -47,6 +49,7 @@ const Carousel = ({
   variant = "default",
   overlayClassName = "",
   draggable = true,
+  onActiveKeyChange,
```
MattIPv4: Nit, I think we'd be fine to call this `onChange` to follow convention

apps/website/src/components/content/Carousel.tsx
```diff
@@ -141,7 +144,18 @@ const Carousel = ({
...
+    // Work out which item is currently scrolled into view
+    if (onActiveKeyChange) {
+      const gap =
+        Number.parseFloat(getComputedStyle(current).columnGap) ||
+        Number.parseFloat(getComputedStyle(current).gap) ||
+        0;
+      const keys = Object.keys(items);
+      const index = Math.round(current.scrollLeft / (width + gap));
+      onActiveKeyChange(keys[Math.max(0, Math.min(index, keys.length - 1))]);
+    }
```
MattIPv4, line 169: Carousel allows for multiple items to be visible at once, how does this handle that?

MattIPv4 (commented):
Thanks for working on this!

apps/website/src/components/content/Carousel.tsx
mrusse: Moved it into a ref that's kept in sync via a useEffect so scrolled will stay stable.

apps/website/src/components/content/Carousel.tsx
mrusse: Renamed to `onChange`.

apps/website/src/components/content/Carousel.tsx
mrusse: Wasn't handled before. Changed it to report undefined when it's more than one, since there's no single "active" item in that case.

MattIPv4:
Hey @mrusse, sorry for not reviewing this yet. I've been travelling the last couple of weeks and haven't had a solid chunk of time to test this out yet. Will get to it as soon as I can, I haven't forgotten!

MattIPv4 (approved):
Thanks :)

PR #2371, opened 2026-09-14
Check `is_sent` response from  `/helix/chat/message`

### LINKED ISSUE #2369: [Bug]: cam control repeat command throttle

It appears you can't send the same exact command within 30 seconds of each other. Which is not normally a problem except for zoom and if some other problem arises and you need to repeat a command.

Clicking the same button twice should be able to send the same command twice.

### Steps to reproduce

1. Click zoom in
2. Click zoom in again
3. Command isn't sent again
4. Wait 30 seconds
5. Click zoom in again
6. Then it sends command

### Browser specs

all

### Anything else?

_No response_

alok0:
## Describe your changes

Failed messages sends for non-api failures respond with 200 and error codes in the response body.

Throw an error if not `is_sent`.

Resolves https://github.com/alveusgg/alveusgg/issues/2369 ?

The frontend `RunCommandButton` should already have status message `Error sending command` popup on errors.

...

...


apps/website/src/server/apis/twitch.ts
```diff
@@ -501,6 +501,21 @@ export async function getClips(
...
+  data: z.array(
+    z.object({
+      message_id: z.string().optional(),
+      is_sent: z.boolean().optional(),
+      drop_reason: z
+        .object({
+          code: z.string().optional(),
+          message: z.string().optional(),
+        })
+        .optional(),
```
MattIPv4:
```suggestion
      message_id: z.string(),
      is_sent: z.boolean(),
      drop_reason: z
        .object({
          code: z.string(),
          message: z.string(),
        })
        .nullable(),
```
This schema is not valid based on the response from Twitch

apps/website/src/server/apis/twitch.ts
```diff
@@ -501,6 +501,21 @@ export async function getClips(
...
+    }),
+  ),
+});
```
MattIPv4, line 517:
```suggestion
const sendChatMessageResponseSchema = z.object({
  data: z.array(
    z.object({
      message_id: z.string(),
      is_sent: z.boolean(),
      drop_reason: z
        .object({
          code: z.string(),
          message: z.string(),
        })
        .nullable(),
    }),
  ).length(1),
});
```
This schema did not match up with what Twitch returned

apps/website/src/server/apis/twitch.ts
```diff
@@ -520,10 +535,16 @@ export async function sendChatMessage(
...
+  const { data } = await sendChatMessageResponseSchema.safeParseAsync(json);
+
+  if (response.status !== 200 || !data?.data?.[0]?.is_sent) {
     console.error(json);
-    throw new Error("Failed to send chat message!");
+    throw new Error(
+      "Failed to send chat message! " +
+        data?.data.map((m) => m.drop_reason?.message || "").join(" ") || "",
+    );
   }
```
MattIPv4, line 548:
```suggestion
  const json = await response.json();
  if (response.status !== 200) {
    console.error(json);
    throw new Error("Failed to send chat message");
  }

  const { data } = await sendChatMessageResponseSchema.safeParseAsync(json);
  if (!data?.data[0]?.is_sent) {
    console.error(json);
    throw new Error(
      `Failed to send chat message: ${data.data[0]?.drop_reason?.message || ""}`,
    );
  }
```

Two things here:

1. We only need to run the body through the schema if we know we got an expected response back
2. If `data` was undefined due to a schema error, the error message would contain `undefined`

MattIPv4 (changes requested):

MattIPv4 (approved):

PR #2384, opened 2026-09-17
Update MinIO Docker image to Quay registry

RutanshS:
## Describe your changes
MinIO recently removed or restricted their core `minio/minio` image on Docker Hub. Running compose on a fresh machine currently fails with a `pull access denied` error. 

This PR updates the compose file to point to the new official registry (`quay.io/minio/minio:latest`) so the local dev environment can spin up successfully.

Refer : `https://quay.io/repository/minio/minio`

Run `docker compose up -d`. Note: If testing locally on a machine that already had the previous image, you may need to temporarily remove your cached `minio/minio:latest` image to force Docker to pull the new one from Quay.


MattIPv4 (commented):
👀 https://github.com/minio/minio it looks like minio is completey abandoned, so I suspect we need some alternative minimal S3 setup locally?

MattIPv4:
SeaweedFS might be a viable alternative? Something like:

```
s3:
    image: chrislusf/seaweedfs:latest
    ports:
      - 9000:9000
    environment:
      WEED_S3_DEFAULT_BUCKETS: alveusgg
      WEED_S3_CONFIG: |
        {
          "identities": [
            {
              "name": "alveusgg",
              "credentials": [
                {
                  "accessKey": "alveusgg_key",
                  "secretKey": "alveusgg_secret"
                }
              ],
              "actions": ["Admin", "Read", "Write"]
            }
          ]
        }
    command: 'server -s3 -s3.port=9000 -s3.allowEmptyFolder=false'
    volumes:
      - s3:/data
```

RutanshS:
Yeah, I just checked, they moved the official repo for minio to quay but they are not actively maintaining it, so switching to another option might actually be the play here. But since it is only for local development, I think SeaweedFS might actually introduce unnecessary complexity.

MattIPv4:
I don't want our local development to rely on an unmaintained image that may gain vulnerabilities over time. I'm totally open to other suggestions for alternatives, though, if you feel Seaweed isn't a good replacement!

MattIPv4:
I did some more exploring on this, and I think I've landed on S3Proxy: #2390 SeaweedFS ended up being too much, as you suspected! Thanks for flagging that the MinIO setup no longer worked 🙏

RutanshS:
Thank you for taking care of this 👍. Closing this PR.

PR #2388, opened 2026-09-19
Replace Zoom Buttons

### LINKED ISSUE #2380: [Idea]: Add a "Copy Command" Button

Add a copy command button, next to the zoom buttons. 

Makes it easier for cam ops to copy/paste in chat instead of having to type it out each time. They only need to change the zoom value since the command will be based on the cam selected.

DanMan149:
## Describe your changes

Resolves #2380

Replaces the zoom in/out buttons on PTZ camera preset pages with a single dropdown offering fixed zoom levels (10, 50, 70, 90, 125, 150, 200, 400, 600). Selecting a value runs `!ptzzoom {camera} {value}` the same way the old buttons did, just as an absolute value instead of two fixed steps (80/120).

Also while at it, I added a clear ("x") button inside the preset search box to reset the search text. heh

...

- Select a PTZ camera with zoom enabled, open the "Zoom" dropdown, confirm all 9 values are listed
- Pick a value, confirm it sends `!ptzzoom {camera} {value}` (checked on local preview via DevTools Network tab against the `stream.runCommand` mutation)
- Confirm the dropdown resets to "Zoom" after each selection
- Type in the preset search box, confirm a clear (x) button appears and resets the search on click

...


alok0:
I would suggest 25, 50, 67, 80, 125, 150, 200, and 400 since these are reciprocals.

MattIPv4:
Might it be better to frame it as +/-20% etc. rather than the underlying values which I think are less obvious to folks what they're doing?

alok0:
Does zoom in +100% make more sense than zoom 200% ? ( it does not )

DanMan149:
I was gonna do something similar to what Dansza suggested in #2380 but tbh I left it out only because it'd just confuse people further.

alok0:
Might help if there were two sections or two buttons, one for zoom out and one for zoom in.

DanMan149:
So what did we decide upon?

MattIPv4:
I think either a single dropdown with +/- percentages or two separate buttons using the icons today. No strong leaning either way, though I imagine a single dropdown is easier to implement.

DanMan149:
image
this is what Dansza suggested, so should I just do something similar to this, but in a single dropdown? 

image

also are these values correct?

alok0:
I think percentages are already relative, you can't make it relative again, the math makes no sense imo.

DanMan149:
> I think percentages are already relative, you can't make it relative again, the math makes no sense imo.

image

this is according to what you suggested

alok0:
The only other sane representation we generally use is multipliers, 200% = 2x; 50% = 0.5x

MattIPv4:
I think it'd be totally fine to represent 90% as -10%, etc.?

alok0:
I think it is absolutely not ok. There is no precedent in UI design for zoom to be represented like that.

MattIPv4:
We're not representing absolute zoom from a base reference, though; this is all relative zoom... I don't think it is intuitive or apparent that 90% each time you select it would do a different thing, nor may it be apparent to all users that that even represents zooming out. I think having a clear concept of zooming in, and zooming out, either through +/- values, or separate +/- buttons, is important there.

You have to remember, while both of you are well experienced with operating the cams, this page is specifically designed to get folks not familiar with operating cams and the underlying bot enabled to do basic operations.

DanMan149:
> You have to remember, while both of you are well experienced with operating the cams, this page is specifically designed to get folks not familiar with operating cams and the underlying bot enabled to do basic operations.

Exactly why I made it the way it is currently. Easier for other cam ops to use, since those are the values they see in chat daily and are probably used to as well. Bringing in percentage or absolute values will confuse them a lot.

alok0:
I understand what you are trying to achieve, but +/- percent doesn't work correctly here.

This is why I advocate for making it clear which values are zoom out and which ones are zoom in. 

Perhaps zoom out could be represented by its reciprocal. Zoom out 2x = 50

MattIPv4:
> since those are the values they see in chat daily and are probably used to as well

I wouldn't over-index on the values seen in chat being the driving factor here -- that is an implementation detail of how it currently works, not the ideal state.

> This is why I advocate for making it clear which values are zoom out and which ones are zoom in.

If we can't figure out a clear way to do that with a single dropdown, I think what Dansza mocked up is probably better than, two clear separate buttons + menus for each operation.

alok0:
Menu with items:

- Zoom out 4x
- Zoom out 2x
- Zoom out 1.5x
- Zoom out 1.25x
- Zoom in 1.25x
- Zoom in 1.5x
- Zoom in 2x
- Zoom in 4x

MattIPv4:
> * Zoom out 4x
> * Zoom out 2x
> * Zoom out 1.5x
> * Zoom out 1.25x
> * Zoom in 1.25x
> * Zoom in 1.5x
> * Zoom in 2x
> * Zoom in 4x

I don't think that's clear at all, unfortunately. I don't see how a 4x zoom-out operation is intuitive to understand.

DanMan149:
what if I revert back to + and - icons, and split the zoom in and zoom out values into two separate drop-downs

MattIPv4:
👍 I think that might be the easiest way to clearly convey the two seperate operations without any ambiguity

apps/website/src/components/ptz/PresetList.tsx
```diff
@@ -70,67 +91,157 @@ const PresetTools = ({
...
+      >
+        {cameras[camera].title}
+        <span className="text-sm text-alveus-green-400 italic">
+          {` (${camera.toLowerCase()})`}
+        </span>
+      </Heading>
+
+      {isCameraPTZ(cameras[camera]) && (
+        <>
+          {zoom && hasScopes && (
```
MattIPv4: We need to check the subscription here, as well as scopes, like `RunCommandButton` does.

MattIPv4: I think we do this in a few different places already, so I wouldn't be opposed to moving the scopes + sub check into a hook to make it DRY-er.

DanMan149:
Moved the scopes + subscription check into a shared `useSubscriberAccess` hook in `hooks/subscription.ts`, used by both `RunCommandButton` and the zoom dropdown now. 
Zoom controls are gated on `hasScopes && subscription.data`, same as `RunCommandButton`'s `subOnly` check.

Didn't touch `presets.tsx`'s own subscription query since it feeds a few other things on the page (zen mode, split view, the status badge).

apps/website/src/components/ptz/PresetList.tsx
```diff
@@ -70,67 +90,155 @@ const PresetTools = ({
...
+                        className="cursor-pointer rounded-sm px-2 py-1 text-sm data-focus:bg-alveus-green-100"
+                        as="li"
+                      >
+                        {level}
+                      </ListboxOption>
+                    ))}
+                  </ListboxOptions>
+                </div>
+              </Listbox>
+
```
MattIPv4: Looks like we lost the little joiner here? `<div className="pointer-events-none -ml-0.5 h-0.5 w-4 rounded-sm bg-alveus-green-400" />`

Image

Image

apps/website/src/components/ptz/PresetList.tsx
```diff
@@ -70,67 +90,155 @@ const PresetTools = ({
...
+                  >
+                    {zoomOutLevels.map((level) => (
+                      <ListboxOption
+                        key={level}
+                        value={level}
+                        className="cursor-pointer rounded-sm px-2 py-1 text-sm data-focus:bg-alveus-green-100"
+                        as="li"
+                      >
+                        {level}
+                      </ListboxOption>
```
MattIPv4: Do we want tooltips on the options similar to what the run command button does, so folks understand what selecting an option does?

Image

apps/website/src/components/ptz/PresetList.tsx
```diff
@@ -70,67 +90,155 @@ const PresetTools = ({
...
+                  >
+                    {zoomInLevels.map((level) => (
+                      <ListboxOption
+                        key={level}
+                        value={level}
+                        className="cursor-pointer rounded-sm px-2 py-1 text-sm data-focus:bg-alveus-green-100"
+                        as="li"
+                      >
+                        {level}
+                      </ListboxOption>
```
MattIPv4: Ditto on the tooltip question

apps/website/src/components/ptz/PresetList.tsx
```diff
@@ -70,67 +90,155 @@ const PresetTools = ({
...
+            <div className="flex items-center">
+              <Listbox
+                value={zoomValue}
+                onChange={onZoomChange}
+                disabled={isPending}
+              >
+                <div className="relative">
+                  <ListboxButton
+                    className={zoomButtonClasses}
+                    aria-label="Zoom out"
```
MattIPv4: I think this should probably be a tooltip rather than an aria label, similar to the list/map view buttons?

apps/website/src/components/ptz/PresetList.tsx
```diff
@@ -70,67 +90,155 @@ const PresetTools = ({
...
+
+              <Listbox
+                value={zoomValue}
+                onChange={onZoomChange}
+                disabled={isPending}
+              >
+                <div className="relative">
+                  <ListboxButton
+                    className={zoomButtonClasses}
+                    aria-label="Zoom in"
```
MattIPv4: Ditto on making this a tooltip

apps/website/src/components/ptz/PresetList.tsx
```diff
@@ -70,67 +90,155 @@ const PresetTools = ({
...
-              args={[camera.toLowerCase(), "80"]}
-              tooltip={{ text: "Run zoom out command" }}
-              icon={IconZoomOut}
-            />
+}) => {
+  const { hasScopes, subscription } = useSubscriberAccess();
+  const canZoom = hasScopes && !!subscription.data;
+
+  const { mutate: runCommand, status } = trpc.stream.runCommand.useMutation();
+  const isPending = status === "pending";
```
MattIPv4: Any ideas on how we can surface feedback when a command fails to run (e.g., when the user is ratelimited), as we no longer have the standard run command button tooltip that'd show that?

apps/website/src/hooks/subscription.ts
MattIPv4: Can we use this on `apps/website/src/pages/about/tech/presets.tsx` where we're doing the same checks for scopes + sub?

MattIPv4 (commented):
Generally looking good, but a few UI/UX nits.

apps/website/src/components/ptz/PresetList.tsx
DanMan149: Oh, I missed it while adding back the two buttons.

apps/website/src/components/ptz/PresetList.tsx
DanMan149: Can add it back, will add a tooltip to each option showing the exact command it'll send. Only reason I left it out was it felt a bit cluttered to me.

apps/website/src/components/ptz/PresetList.tsx
DanMan149: Will add the tooltips here as well.

apps/website/src/components/ptz/PresetList.tsx
DanMan149: Will switch this to useTooltip, same as the list/map buttons.

apps/website/src/components/ptz/PresetList.tsx
MattIPv4: Yah that's fair, the other option is maybe to solve both at once by having something close to the original tooltips on the icons themselves that shows a generic version of the command that'll run?

apps/website/src/components/ptz/PresetList.tsx
DanMan149: Can bring back the same status text RunCommandButton showed ("Sending command...", "Command sent!", "Error sending command"), surfaced via tooltip on both zoom buttons since they share one mutation now.

apps/website/src/hooks/subscription.ts
DanMan149: Yep, can use it on presets.tsx as well.

apps/website/src/components/ptz/PresetList.tsx
DanMan149: image
image


I added back the tooltips, this is how it looks.
Could also move its placement to the right, think looks a bit cleaner.
image

DanMan149: Ignore the "Zoom in to 125", will replace that with "Run zoom in command".

MattIPv4: Yeh, right placement looks cleaner for the options for sure 👍

MattIPv4: Should these be suffixed with `%`? They're relative percentages to the current, right?

DanMan149: Not right now, but can do percentage values on the tooltip. So it'll look something like "Zoom in 25%" with the zoom command below it.

DanMan149: Thoughts?
image
image

MattIPv4: Why can't we update the option itself to render the %? Wouldn't it just be this?

```suggestion
                      <ListboxOption
                        key={level}
                        value={level}
                        className="cursor-pointer rounded-sm px-2 py-1 text-sm data-focus:bg-alveus-green-100"
                        as="li"
                      >
                        {level}%
                      </ListboxOption>
```

MattIPv4: I think to what @alok0 had been saying before, though, it should just be the values themselves, not computed different values? I'm supportive of `Zoom in by 50%` rather than `Zoom in to 150%`, but not sure there was agreement on that?

alok0: I'm in favor of `Zoom in 150%` or `Zoom in by 150%` as "to" implies absolute to me.

MattIPv4: Is saying `Zoom in by 150%` correct, or should it be with respect to 100%, so `Zoom in by 50%`? Is it a 50% increase, not a 150% increase?

DanMan149: Yeah, mathematically it should be the 50% increase, not 150%, since 100 is the baseline

alok0: Perhaps `Zoom in 150%` is more accurate. But I believe matching the value sent in the command is important to reducing confusion in the long run, and for teaching how do visualize zooms. shrujj

MattIPv4: Maybe this is me nitpicking the grammar, but I feel that `Zoom in 150%` still implies a relative change of 150%, reading the same as `Zoom in by 150%`?

> But I believe matching the value sent in the command is important to reducing confusion in the long run, and for teaching how do visualize zooms

I get that, but I think we're at least partially solving this by having the tooltips still show the actual command being run, and I don't think it's an overly large mental jump to understand relative percentages from a 100 baseline?

MattIPv4: I suppose, actually, we could do both? Have the dropdown option show `150%`, and then the tooltip (which is there to explain the option) show `Zoom in by 50%`?

DanMan149: so something like this?

image

MattIPv4: 👍 Yeh, while the numbers don't match, that seems actually pretty clear to me as to what the actual value is and explaining the relative change.

@alok0 thoughts?

alok0: if you want to be doubly/triply clear it could say `Zoom in by +100%`, `Zoom out by -50%`

DanMan149: Lmao, think this is good enough.

MattIPv4: 👍 I'm supportive for adding the +/- for clarity. Thanks both for the iteration on this ❤️

apps/website/src/components/ptz/PresetList.tsx
```diff
@@ -13,21 +28,28 @@ import { classes } from "@/utils/classes";
...
 import IconMapPin from "@/icons/IconMapPin";
 import IconMenu from "@/icons/IconMenu";
+import IconXCircle from "@/icons/IconXCircle";
 import IconZoomIn from "@/icons/IconZoomIn";
 import IconZoomOut from "@/icons/IconZoomOut";
 
 type PresetView = "list" | "map";
+type ZoomDirection = "in" | "out";
+
+const zoomOutLevels = ["10", "50", "70", "90"] as const;
```
MattIPv4: Should we reverse the order of this, so the smallest relative change is first, matching the zoom in levels?

apps/website/src/components/ptz/PresetList.tsx
```diff
@@ -70,67 +128,190 @@ const PresetTools = ({
...
+  }, [status]);
+
+  const zoomOutTooltip = useTooltip({
+    content: statusText ?? "Zoom out",
+    force: !!statusText,
+  });
+  const zoomInTooltip = useTooltip({
+    content: statusText ?? "Zoom in",
+    force: !!statusText,
+  });
```
MattIPv4: This is causing both icons to show the tooltip when there is a status -- I think ideally we want to only show it for the direction being used.

Might be worth moving each icon and its menu/hooks into a dedicated component to isolate them, so the tooltip is inherently only going to show for the one being used, and then have two of them in this parent component?

MattIPv4 (approved):
Thanks!

